PR #30, opened 2024-06-03
chore: correct the comment for GetManagedLiveObjs

xdevxy:
### Modifications

 correct the comment for `LiveStateCache.GetManagedLiveObjs` and fix the unit test failure


juliev0:
good catch. Looks like there was also a unit test compilation error that didn't get caught earlier?

xdevxy:
> good catch. Looks like there was also a unit test compilation error that didn't get caught earlier?
Ah, yes. maybe the CI is not triggered?

juliev0 (approved):

PR #32, opened 2024-06-04
Add unit test for controller rollout create and update

### LINKED ISSUE #22: Verify and add unit test for NumaflowControllerRollout create and update reconciliation logic

chandankumar4:
Fixes #22 

### Modifications

- Verified the Numaplane controller create feature and update feature
- Add unit test for Numaplane controller Create and update

### Verification
- Validated it in local kubernetes cluster

### DIFF
--- a/internal/controller/numaflowcontrollerrollout_controller_test.go
+++ b/internal/controller/numaflowcontrollerrollout_controller_test.go
@@ -32,7 +32,7 @@ import (
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
 )
 
-var _ = Describe("Validate Numaflow CRD", Ordered, func() {
+var _ = Describe("NumaflowControllerRollout Controller", Ordered, func() {
 	const (
 		namespace = "default"
 		duration  = 20 * time.Second
--- a/internal/controller/numaflowcontrollerrollout_controller_test.go
+++ b/internal/controller/numaflowcontrollerrollout_controller_test.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -26,37 +27,38 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
 )
 
-var _ = Describe("NumaflowControllerRollout Controller", Ordered, func() {
+var _ = Describe("Validate Numaflow CRD", Ordered, func() {
 	const (
-		namespace    = "default"
-		resourceName = "numaflow-controller"
+		namespace = "default"
+		duration  = 20 * time.Second
+		interval  = 1 * time.Second
 	)
+	Context("When creating a NumaflowControllerRollout resource", func() {
+		ctx := context.Background()
 
-	ctx := context.Background()
-
-	Context("When reconciling a resource", func() {
-		typeNamespacedName := types.NamespacedName{
-			Name:      resourceName,
-			Namespace: namespace,
+		resourceLookupKey := types.NamespacedName{
+			Name:      "numaflow-controller",
+			Namespace: "default",
 		}
 
-		numaflowControllerRollout := apiv1.NumaflowControllerRollout{
+		numaflowControllerResource := apiv1.NumaflowControllerRollout{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      resourceName,
-				Namespace: namespace,
+				Name:      "numaflow-controller",
+				Namespace: "default",
 			},
 			Spec: apiv1.NumaflowControllerRolloutSpec{
-				Controller: apiv1.Controller{Version: "1.2.1"},
+				Controller: apiv1.Controller{Version: "1.2.0"},
 			},
 		}
 
 		It("Should throw a CR validation error", func() {
 			By("Creating a NumaflowControllerRollout resource with an invalid name")
-			resource := numaflowControllerRollout
+			resource := numaflowControllerResource
 			resource.Name = "test-numaflow-controller"
 			err := k8sClient.Create(ctx, &resource)
 			Expect(err).NotTo(Succeed())
@@ -65,7 +67,7 @@ var _ = Describe("NumaflowControllerRollout Controller", Ordered, func() {
 
 		It("Should throw duplicate resource error", func() {
 			By("Creating duplicate NumaflowControllerRollout resource with the same name")
-			resource := numaflowControllerRollout
+			resource := numaflowControllerResource
 			err := k8sClient.Create(ctx, &resource)
 			Expect(err).To(Succeed())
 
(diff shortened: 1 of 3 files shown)


internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -36,48 +42,102 @@ var _ = Describe("NumaflowControllerRollout Controller", func() {
...
+			Expect(result).NotTo(BeNil())
+
+			// Fetch the resource and update the spec
+			var createdResource = &apiv1.NumaflowControllerRollout{}
+			Expect(k8sClient.Get(ctx, typeNamespacedNameForUpdate, createdResource)).To(Succeed())
+			createdResource.Spec = apiv1.NumaflowControllerRolloutSpec{
+				Controller: apiv1.Controller{Version: "1.2.1"},
+			}
+
+			// Update the resource after the spec change
```
xdevxy, line 130: We also need the delete case, right? Or you want to have a separate PR for it?

internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -36,48 +42,102 @@ var _ = Describe("NumaflowControllerRollout Controller", func() {
...
+				Spec: apiv1.NumaflowControllerRolloutSpec{
+					Controller: apiv1.Controller{Version: "1.2.0"},
+				},
+			}
+			typeNamespacedNameForUpdate := types.NamespacedName{
+				Name:      "test-resource-updated",
+				Namespace: "default",
+			}
+
+			// Create the resource to be updated
```
xdevxy, line 117: We can also verified the underlying numaflow controller spec is updated as well?

internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -36,48 +42,102 @@ var _ = Describe("NumaflowControllerRollout Controller", func() {
...
-		// 	_, err := controllerReconciler.Reconcile(ctx, sigsReconcile.Request{
-		// 		NamespacedName: typeNamespacedName,
-		// 	})
-		// 	Expect(err).NotTo(HaveOccurred())
-		// 	// TODO(user): Add more specific assertions depending on your controller's reconciliation logic.
-		// 	// Example: If you expect a certain status condition after reconciliation, verify it here.
-		// })
+
+		It("should successfully reconcile the resource", func() {
+			By("Reconciling the created resource")
```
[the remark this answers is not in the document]
chandankumar4: you mean `numaflow-controller` right?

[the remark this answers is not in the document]
chandankumar4: For checking that `numaflow-controller` is running I have to run the numaflow controller as well with all dependent configs which I think will be a kind of e2e test instead of unit test? what do you think?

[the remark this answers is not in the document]
chandankumar4: For checking that `numaflow-controller` is running I have to run the numaflow controller as well with all dependent configs which I think will be a kind of e2e test instead of unit test? what do you think?

internal/controller/numaflowcontrollerrollout_controller_test.go
chandankumar4: I'll raise a separate PR for it. Created an issue to track it: https://github.com/numaproj/numaplane/issues/46

internal/controller/numaflowcontrollerrollout_controller_test.go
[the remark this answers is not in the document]
xdevxy: even as a unit test, we'd like to verify the reconciliation logic is as expected for individual controller, with envtest that provides the mocked cluster, ideally we would want to verify the underlying numaflow controller as well. Then in the e2e test, we are going to verify in a real cluster with broader cross component coverage. Thanks!

tests/config/controller-definitions-config.yaml
```diff
@@ -1,10 +1,769 @@
 controllerDefinitions:
-  - version: "1.2.1"
+  - version: "1.2.0"
```
chandankumar4, line 2: Using this file in unit test, so that I can have at least two version of numaflow for testing

internal/controller/suite_test.go
```diff
@@ -205,22 +204,13 @@ func downloadCRD(url string, downloadDir string) {
 
 func getNumaflowControllerDefinitions() config.NumaflowControllerDefinitionConfig {
 	// Read definitions config file
-	// TODO: use this file instead "../../tests/config/controller-definitions-config.yaml"
-	data, err := os.ReadFile("../../config/manager/numaflow-controller-definitions-config.yaml")
+	configData, err := os.ReadFile("../../tests/config/controller-definitions-config.yaml")
```
chandankumar4, line 207: Using this file in unit test, so that I can have at least two version of numaflow for testing

afugazzotto: I was thinking we could use `../../tests/config/controller-definitions-config.yaml` for more test cases and add more versions to it. Some versions could be made in a way that causes errors so we can also test "unhappy" cases. Thank you for migrating from `numaflow-controller-definitions-config.yaml` to this test file.

internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -19,44 +19,46 @@ package controller
...
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
 )
 
 var _ = Describe("NumaflowControllerRollout Controller", Ordered, func() {
 	const (
-		namespace    = "default"
-		resourceName = "numaflow-controller"
+		namespace = "default"
+		duration  = 20 * time.Second
+		interval  = 1 * time.Second
```
afugazzotto, line 39: Why do you need to use different values that the default values defined in the suit_test.go file?

internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -19,44 +19,46 @@ package controller
...
+			Namespace: "default",
 		}
 
-		numaflowControllerRollout := apiv1.NumaflowControllerRollout{
+		numaflowControllerResource := apiv1.NumaflowControllerRollout{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      resourceName,
-				Namespace: namespace,
+				Name:      "numaflow-controller",
+				Namespace: "default",
```
afugazzotto, line 52: Why not reuse the namespace const defined above since it is the same?

internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -75,6 +77,42 @@ var _ = Describe("NumaflowControllerRollout Controller", Ordered, func() {
 			Expect(err.Error()).To(ContainSubstring("numaflowcontrollerrollouts.numaplane.numaproj.io \"numaflow-controller\" already exists"))
 		})
 
+		It("Should reconcile the Numaflow Controller Rollout", func() {
+			By("Verifying the phase of the NumaflowControllerRollout resource")
+			// Loop until the API call returns the desired response or a timeout occurs
+			Eventually(func() (apiv1.Phase, error) {
+				createdResource := &apiv1.NumaflowControllerRollout{}
+				_ = k8sClient.Get(ctx, resourceLookupKey, createdResource)
```
afugazzotto, line 85: Should still check for errors and expect for no error to have occurred.

internal/controller/numaflowcontrollerrollout_controller_test.go
afugazzotto: We can also change the default values and check that they work for all tests.

internal/controller/numaflowcontrollerrollout_controller_test.go
chandankumar4: yaa, sorry I missed it

internal/controller/numaflowcontrollerrollout_controller_test.go
chandankumar4: I'll update it accordingly, as it was not working earlier maybe It should work now after fixing the test cases.

xdevxy (approved):

PR #41, opened 2024-06-05
PipelineRollout Tests

### LINKED ISSUE #20: Ensure create, update and delete logic works for PipelineRollout with proper unit tests

afugazzotto:
Fixes #20 

### Modifications

Updated the `suite_test.go` file to be able to setup an environment to test PipelineRollout reconciliation logic using Numaflow CRDs. Also, added test cases to `pipelinerollout_controller_test.go` file to test create, update, and delete of a PipelineRollout resource.

### Verification

Issued `make test` command and verify all tests passed.


internal/controller/suite_test.go
```diff
@@ -27,33 +31,52 @@ import (
...
+		"https://raw.githubusercontent.com/numaproj/numaflow/main/config/base/crds/minimal/numaflow.numaproj.io_vertices.yaml",
+	}
+	externalCRDsDir = filepath.Join("..", "..", "config", "crd", "external")
+	for _, crdURL := range crdsURLs {
+		downloadCRD(crdURL, externalCRDsDir)
+	}
+
 	By("bootstrapping test environment")
+	// TODO: IDEA: could set useExistingCluster via an env var to be able to reuse some tests cases in e2e tests
+	// by setting this variable in CI for unit tests and e2e tests appropriately.
```
juliev0, line 76: what does it mean "to reuse some test cases in e2e tests"? do you have an example of what you mean?

afugazzotto: The test cases in the `pipelinerollout_controller_test.go` file test PipelineRollout lifecycle (create, update, delete operations). These could be reused in e2e tests and, when using a real cluster, also the deletion of the Numaflow Pipeline on deletion of PipelineRollout could be tested (https://github.com/numaproj/numaplane/pull/41/files#diff-31f7e2289e4cc3f50a723cc05922642d3cc20247ed71f66dc7dab1a73d3f302cR237-R245).

juliev0: Interesting. Possibly. Although, that will likely use the same test framework "github.com/stretchr/testify/suite" we were previously using....so you'd have to see if it would be easily callable from there, or otherwise may not be worth it

juliev0:
This is very elegant....but I'm not sure if I see any cases of things going wrong? should that be added to a subsequent PR? (at least the case of invalid yaml, not sure what else off the top of my head)

juliev0:
It should be noted this unit test doesn't really cover things on an individual function level, but maybe this is really the only good way to test `Reconcile()`. But it makes me wonder if we want to include individual function tests for some functions that are strictly related to data processing now or in the future. Any functions that are trivial it's probably not worth doing, but I'm not sure if any of our functions are non-trivial.

xdevxy (approved):

xdevxy:
yes, this is validation check in CRD not webhook, right? looks good then.

internal/controller/pipelinerollout_controller_test.go
```diff
@@ -27,65 +29,182 @@ import (
...
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			By("Verifying the content of the pipeline spec")
+			Expect(createdResource.Spec).Should(Equal(pipelineSpec))
+		})
+
+		It("Should have the PipelineRollout Status Phase has Running", func() {
+			Consistently(func() (apiv1.Phase, error) {
+				createdResource := &apiv1.PipelineRollout{}
```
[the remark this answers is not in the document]
afugazzotto: We need the Numaflow Controller to be able to see that the pipeline goes into healthy state. We should do this in E2E tests.

xdevxy:
@juliev0 do you want to take another look? Otherwise I will merge this. Thanks!

juliev0:
> @juliev0 do you want to take another look? Otherwise I will merge this. Thanks!

please do, thanks!

PR #47, opened 2024-06-07
Add NumaflowControllerRollout CRD name validation

### LINKED ISSUE #7: Prevent more than 1 NumaflowControllerRollout on a namespace
The easiest way to take care of this is to only allow one name for the `NumaflowControllerRollout`, e.g. "numaflow-controller" (in which case our [sample](https://github.com/numaproj-labs/numaplane/blob/rolloutControllers/config/samples/numaplane.numaproj.io_v1alpha1_numaflowcontrollerrollout.yaml#L4) is wrong and should be changed).

Addresses [this](https://github.com/numaproj-labs/numaplane/blob/96010907b30ebe8e4a0072abc8a5140bee55683c/internal/controller/numaflowcontrollerrollout_controller.go#L123) todo.

chandankumar4:
Fixes #7 

### Modifications
- Add CRD validation to support only name for `numaflowControllerRollout` to prevent having more than 1 resource in a namespace

### Verification
- Verified in local kubernetes cluster


xdevxy (approved):
In the unit test you are working one, we can add a test case for it. Thanks!

chandankumar4:
> In the unit test you are working one, we can add a test case for it. Thanks!

Sure, Have created a separate PR for it: https://github.com/numaproj/numaplane/pull/53

PR #50, opened 2024-06-07
feat: Add test for ISBServiceRollout

### LINKED ISSUE #40: Ensure create, update and delete logic works for ISBServiceRollout
Note that delete PipelineRollout and ISBServiceRollout CRs should delete the underlying pipeline and ISBservice, except the case ISBservice is shared with another pipeline.

a49ibrahim:
Fixes #40 

### Modifications

added test cases to isbservicerollout_controller_test.go file to test create, update, and delete of a isbservicerollout resource.

### Verification

ran make test command and verify all tests passed.


internal/controller/isbservicerollout_controller_test.go
```diff
@@ -31,60 +32,133 @@ import (
...
+	ctx := context.Background()
+
+	isbServiceRollout := &apiv1.ISBServiceRollout{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      isbServiceRolloutName,
+		},
+		Spec: apiv1.ISBServiceRolloutSpec{
+			InterStepBufferService: runtime.RawExtension{
+				Raw: []byte(`{
```
afugazzotto, line 51: It would be better to use a more realistic example like https://github.com/numaproj/numaplane/blob/main/config/samples/numaplane.numaproj.io_v1alpha1_isbservicerollout.yaml
Also, I'd recommend not using string but an actual Numaflow InterStepBufferService object so you can reuse it for updates and comparisons (similar to the tests in the PipelineRollout Controller).

internal/controller/isbservicerollout_controller_test.go
```diff
@@ -31,60 +32,133 @@ import (
...
+			Expect(k8sClient.Create(ctx, isbServiceRollout)).Should(Succeed())
+
+			createdResource := &apiv1.ISBServiceRollout{}
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, resourceLookupKey, createdResource)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			By("Verifying the content of the ISBServiceRollout spec field")
+			Expect(createdResource.Spec).Should(Equal(isbServiceRollout.Spec))
```
afugazzotto, line 85: Also check the created child resource to make sure it has the expected content.

internal/controller/isbservicerollout_controller_test.go
```diff
@@ -31,60 +32,133 @@ import (
...
+
+		It("Should update the ISBServiceRollout", func() {
+			By("updating the ISBServiceRollout")
+
+			currentISBServiceRollout := &apiv1.ISBServiceRollout{}
+			Expect(k8sClient.Get(ctx, resourceLookupKey, currentISBServiceRollout)).ToNot(HaveOccurred())
+
+			// Update the spec here
+			currentISBServiceRollout.Spec.InterStepBufferService = runtime.RawExtension{
+				Raw: []byte(`{
```
afugazzotto, line 96: See above comment related to this.

internal/controller/isbservicerollout_controller_test.go
afugazzotto: Similar to:
- https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout_controller_test.go#L127
- https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout_controller_test.go#L138

xdevxy:
> Fixes #40
> 
> ### Modifications
> ### Verification

can you fill in the `modification` and `verification` you have done, thanks!

xdevxy (approved):

internal/controller/isbservicerollout_controller_test.go
```diff
@@ -18,73 +18,172 @@ package controller
...
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, resourceLookupKey, createdISBR)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			By("Verifying the content of the ISBServiceRollout spec")
+			createdInterStepBufferServiceSpec := numaflowv1.InterStepBufferService{}
+			Expect(json.Unmarshal(createdISBR.Spec.InterStepBufferService.Raw, &createdInterStepBufferServiceSpec)).ToNot(HaveOccurred())
+			Expect(createdISBR.Spec).Should(Equal(isbServiceRollout.Spec))
+		})
```
xdevxy, line 99: This seems to be repeated from L76-L85. You should instead verify ISBService is created, right?

xdevxy:
The failure is actually related to the latest change and should be fixed with https://github.com/numaproj/numaplane/pull/56

xdevxy (approved):

PR #52, opened 2024-06-10
Spec hash annotation

### LINKED ISSUE #4: Check for Pipeline and ISBService changes prior to trying to update them
We can store hash of previous spec in the Annotation of the `PipelineRollout` and the `ISBServiceRollout` and then compare to hash of the new spec. (For reference, Numaflow uses [this](https://github.com/numaproj/numaflow/blob/52d839dd902f21f5cff1dd47c331642734a1f98e/pkg/shared/util/hash.go#L24).)

afugazzotto:
Fixes #4

### Modifications

Created the function `makeChildResourceFromRolloutAndUpdateSpecHash` to make a new kubernetes.GenericObject based on the given rolloutObj. The function returns the child resource object ready to be created and an operation to be performed with the returned object. The operations are defined by the RolloutChildOperation constants and are: create the child resource, update the child resource, and do nothing. The returned operation will be used after this function is called to perform create/update operations or to skip create/update.

### Verification

Used existing unit tests to make sure PipelineRollout did not change expected behavior. Added unit tests to verify new functions correctness.

### NOTE
Additional changes to improve reusability of the functions should be made. However, we should make those changes in a different PR to have smaller incremental changes.


.gitignore
```diff
@@ -30,4 +30,6 @@ bin/*
 go.work
 
 config/crd/bases/_.yaml
-.idea
\ No newline at end of file
+.idea
+
+config/crd/external/
```
xdevxy, line 35: Why do we need to add this?

internal/util/util.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+		return hex.EncodeToString(hash.Sum(nil))
+	case string:
+		return MustHash([]byte(data))
+	default:
+		return MustHash([]byte(MustJSON(v)))
+	}
+}
+
+func MustJSON(in any) string {
```
xdevxy, line 40: nit: can you add a comment for these two functions.

.gitignore
afugazzotto: While working on unit tests I had a panic that crashed the tests and cleanup did not happen. This is just in case it were to happen again so we do not risk to commit unnecessary files. It's just for convenience.

internal/controller/pipelinerollout_controller_test.go
```diff
@@ -152,6 +157,24 @@ var _ = Describe("PipelineRollout Controller", func() {
 			currentPipelineRollout := &apiv1.PipelineRollout{}
 			Expect(k8sClient.Get(ctx, resourceLookupKey, currentPipelineRollout)).ToNot(HaveOccurred())
 
+			var lastTransitionTime time.Time
```
xdevxy, line 160: Do you mind adding a test for checking the annotation is there after the first creation?

internal/controller/shared.go
```diff
@@ -0,0 +1,126 @@
...
+
+type RolloutChildOperation string
+
+const (
+	// The child resource does not exist on the cluster and needs to be created
+	RolloutChildNew RolloutChildOperation = "CREATE_CHILD_RESOURCE"
+	// The child resource exists on the cluster but needs to be update
+	RolloutChildUpdate RolloutChildOperation = "UPDATE_CHILD_RESOURCE"
+	// The child resource exists on the cluster and does not need to be update
+	RolloutChildNone RolloutChildOperation = "DO_NOTHING"
```
xdevxy, line 43: nit: `RolloutChildNoop` as discussed

xdevxy (approved):

internal/controller/pipelinerollout_controller.go
```diff
@@ -155,35 +154,22 @@ func (r *PipelineRolloutReconciler) reconcile(
...
-			if err != nil {
-				return false, err
-			}
+	if rolloutChildOp == RolloutChildNew {
+		err = kubernetes.CreateCR(ctx, r.restConfig, obj, "pipelines")
+		if err != nil {
+			return false, err
 		}
-	} else {
+	} else if rolloutChildOp == RolloutChildUpdate {
```
juliev0, line 169: @afugazzotto @xdevxy So, the logic in here only occurs if the spec of the child object has changed, right? I think there's some logic in here related to pausing that we need to do even if that's not the case, isn't there? Such as calling `isPipelinePaused()`

afugazzotto: Good catch! I also noticed that we need to call `processPipelineStatus` to update the status correctly. I'm making those change now.

juliev0: I wonder if we should only set the annotation on the Rollout spec until we truly update the object itself.

juliev0: I tend to think so, right? The annotation should represent that which has actually been deployed. 

In the pause logic below we're essentially trying to update the spec. If we don't set the annotation until we've actually deployed the object to that spec, then it is okay to use the conditional "if spec is different" to determine whether to execute that logic.

afugazzotto: So, it may be best to set the hash annotation only after a successful call to `ApplyCRSpec`? Would it make sense to set the hash annotation together with the status (ex: in `processPipelineStatus`)?

juliev0: > So, it may be best to set the hash annotation only after a successful call to ApplyCRSpec?

yes

> Would it make sense to set the hash annotation together with the status (ex: in processPipelineStatus)?

`processPipelineStatus()` should be called all the time

afugazzotto: https://github.com/numaproj/numaplane/pull/56

PR #53, opened 2024-06-12
Add unit test for crd validation

chandankumar4:
### Modifications

- Add unit test for CRD validation of numaflow controller

### Verification
- Verified in local kind cluster

### DIFF
--- a/internal/controller/numaflowcontrollerrollout_controller_test.go
+++ b/internal/controller/numaflowcontrollerrollout_controller_test.go
@@ -30,14 +30,17 @@ import (
 )
 
 var _ = Describe("NumaflowControllerRollout Controller", func() {
+	const (
+		namespace = "default"
+	)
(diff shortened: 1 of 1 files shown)


internal/controller/numaflowcontrollerrollout_controller_test.go
```diff
@@ -65,6 +68,23 @@ var _ = Describe("NumaflowControllerRollout Controller", func() {
 			By("Cleanup the specific resource instance NumaflowControllerRollout")
 			Expect(k8sClient.Delete(ctx, resource)).To(Succeed())
 		})
+
+		It("Should throw a CR validation error", func() {
```
xdevxy, line 72: its trivial, but do you mind adding another test to ensure the creation of second rollout object with numaflow-controller won't succeed. Thanks!

xdevxy (approved):

PR #56, opened 2024-06-12
fix: fixed logic to run certain operations even when the spec is unchanged

afugazzotto:
### Modifications

Operations such as pipeline pausing and status update based on pipeline and isbservice status need to be performed even if the specs are unchanged based on hash comparison.

### Verification

Unit tests and manually checking via ArgoCD and simulating a pipeline pause.

**NOTE:** we cannot test by checking that the ChildResourcesHealthy condition is True because it is based on the pipeline status which will never be healthy in the unit tests since there is no numaflow-controller running. This could be checked in e2e tests.


internal/controller/shared.go
```diff
@@ -44,13 +44,14 @@ const (
 )
 
 // makeChildResourceFromRolloutAndUpdateSpecHash makes a new kubernetes.GenericObject based on the given rolloutObj.
-// It returns the child resource object ready to be created and an operation to be performed with the returned object.
+// It returns the child resource object ready to be created, an operation to be performed with the returned object,
+// and the calculated hash of the raw spec of the rollout to be set in the rollout annotations after succesful deployment.
 // The operations are defined by the RolloutChildOperation constants.
 func makeChildResourceFromRolloutAndUpdateSpecHash(
```
xdevxy, line 50: We no longer update the Hash in this function, right? If so, the name should be updated accordingly.

internal/controller/pipelinerollout_controller.go
```diff
@@ -338,15 +338,25 @@ func applyPipelineSpec(
...
+	if rolloutChildOp != RolloutChildNoop {
+		// TODO: use UpdateSpec instead
+		err := kubernetes.ApplyCRSpec(ctx, restConfig, obj, "pipelines")
+		if err != nil {
+			numaLogger.Errorf(err, "failed to apply Pipeline: %v", err)
+			pipelineRollout.Status.MarkFailed("ApplyPipelineFailure", err.Error())
+			return err
+		}
+
+		kubernetes.SetAnnotation(pipelineRollout, apiv1.KeyHash, childResouceSpecHash)
```
xdevxy, line 355: pause/resume the pipeline spec change will cause the hash to be different right? If so, we need to calculate the new hash afterwards.

internal/controller/pipelinerollout_controller.go
```diff
@@ -338,15 +338,25 @@ func applyPipelineSpec(
...
 ) error {
 	numaLogger := logger.FromContext(ctx)
-	// TODO: use UpdateSpec instead
-	err := kubernetes.ApplyCRSpec(ctx, restConfig, obj, "pipelines")
-	if err != nil {
-		numaLogger.Errorf(err, "failed to apply Pipeline: %v", err)
-		pipelineRollout.Status.MarkFailed("ApplyPipelineFailure", err.Error())
-		return err
+
+	if rolloutChildOp != RolloutChildNoop {
```
xdevxy, line 346: We can avoid pass in `RolloutChildOperation` here by checking `RolloutChildNoop` at L171, right?

internal/controller/pipelinerollout_controller.go
afugazzotto: Only the pipeline resource will be updated and we compare the hash of the pipeline rollout `pipeline` spec. So no need to change as discussed on phone.

internal/controller/pipelinerollout_controller.go
```diff
@@ -168,7 +168,7 @@ func (r *PipelineRolloutReconciler) reconcile(
 		}
 
 		kubernetes.SetAnnotation(pipelineRollout, apiv1.KeyHash, childResouceSpecHash)
-	} else {
+	} else if rolloutChildOp != RolloutChildNoop {
```
xdevxy, line 171: nit: I'd prefer this to be `if rolloutChildOp == RolloutChildUpdate` to be more clear.

xdevxy (approved):

internal/controller/shared.go
```diff
@@ -108,19 +109,7 @@ func makeChildResourceFromRolloutAndUpdateSpecHash(
...
-func setAnnotation(obj metav1.Object, key, value string) {
-	annotations := obj.GetAnnotations()
-	if annotations == nil {
-		annotations = make(map[string]string)
-	}
-
-	annotations[key] = value
-
-	obj.SetAnnotations(annotations)
+	return &obj, rolloutChildOp, childResouceSpecHash, nil
```
juliev0, line 114: `if annotation, exists := annotations[apiv1.KeyHash]; exists && annotation != childResouceSpecHash {`

I missed this earlier. But looking at this line above, what if `exists==false` above?

internal/controller/shared.go
```diff
@@ -43,14 +43,15 @@ const (
...
 )
 
-// makeChildResourceFromRolloutAndUpdateSpecHash makes a new kubernetes.GenericObject based on the given rolloutObj.
-// It returns the child resource object ready to be created and an operation to be performed with the returned object.
+// makeChildResourceFromRolloutAndCalculatepecHash makes a new kubernetes.GenericObject based on the given rolloutObj.
+// It returns the child resource object ready to be created, an operation to be performed with the returned object,
+// and the calculated hash of the raw spec of the rollout to be set in the rollout annotations after succesful deployment.
 // The operations are defined by the RolloutChildOperation constants.
-func makeChildResourceFromRolloutAndUpdateSpecHash(
+func makeChildResourceFromRolloutAndCalculateSpecHash(
```
juliev0, line 50: I hate to be anal, but wondering if we can shorten this name? It's pretty long. I can see it's explicit about what it does, which can be nice. But if you have to name it that way, it sort of reveals a violation of the Single Responsibility Principle, that a function should do one thing. I'm not a total stickler for that rule. (I do think a function can do multiple things but still conceptually do one thing in general.)

juliev0: In any case, can we either shorten the name or break it into multiple functions?

internal/controller/pipelinerollout_controller.go
juliev0: yeah, I was thinking the same. The logic below applies to when we're trying to update.

afugazzotto: I actually just realized that we need to remove this. I had it in place in the previous PR but it needs to be removed because part of the logic inside the functions called in this if-statement branch need to execute in case the pipeline needs to be updated or not. The code in this branch needs to be refactored, but I wanted to do that in a separate PR as discussed with @xdevxy

juliev0: > I actually just realized that we need to remove this. I had it in place in the previous PR but it needs to be removed because part of the logic inside the functions called in this if-statement branch need to execute in case the pipeline needs to be updated or not.

You know, I made that comment back when you were updating the annotation immediately whenever there was a need to update. Now that you're not updating that annotation until the update has truly happened, I think it may be okay. Basically, we will revisit this "update" logic up until the update actually occurs.

internal/controller/shared.go
```diff
@@ -97,30 +98,18 @@ func makeChildResourceFromRolloutAndUpdateSpecHash(
...
 		} else {
-			return nil, RolloutChildNoop, fmt.Errorf("unable to get %s %s/%s: %v", kind, obj.Namespace, obj.Name, err)
+			return nil, RolloutChildNoop, "", fmt.Errorf("unable to get %s %s/%s: %v", kind, obj.Namespace, obj.Name, err)
 		}
 	}
 
 	if rolloutChildOp == RolloutChildNoop {
 		annotations := rolloutObj.GetAnnotations()
-		if annotation, exists := annotations[apiv1.KeyHash]; exists && annotation != childResouceSpecHash {
+		if annotation := annotations[apiv1.KeyHash]; annotation != childResouceSpecHash {
```
juliev0, line 107: ahh, that's a nice neat way to handle annotation not being present

internal/controller/pipelinerollout_controller.go
```diff
@@ -166,7 +166,9 @@ func (r *PipelineRolloutReconciler) reconcile(
 		if err != nil {
 			return false, err
 		}
-	} else if rolloutChildOp == RolloutChildUpdate {
+
+		kubernetes.SetAnnotation(pipelineRollout, apiv1.KeyHash, childResouceSpecHash)
+	} else {
```
juliev0, line 171: did you see my comment [here](https://github.com/numaproj/numaplane/pull/56#discussion_r1639024255)? I actually think it does make sense for this logic to be when we're still in the process of trying to update...unless I'm missing something?

internal/controller/pipelinerollout_controller.go
juliev0: > The code in this branch needs to be refactored

are you referring partly to my other comment regarding the long function name here?

afugazzotto:
@juliev0 @xdevxy I made some changes to the spec hashing and annotation setting code based on your feedback to try and resolve all of the concerns. Please, take a look again and let me know what you think. Thanks

internal/controller/shared.go
afugazzotto: I initially envisioned a different usage for this function. But after realizing that certain logic needs to be executed independently of the fact of having a creation or update, lead me to streamline this function into a simpler function.

juliev0:
Thank you for the adjustments @afugazzotto!

Unfortunately, and I am so sorry for this - I think I didn’t think this whole annotation thing through before. I am realizing that given that we want to do auto-healing, we want to overwrite the pipeline/ISBservice if somebody hand modifies it (like through kubectl apply). This means that we ultimately want to compare a. The spec we intend to apply (including “pause” designation) to b. the spec currently on the cluster. Therefore, there may be no value to storing the hash after all. :( 

cc @xdevxy @dpadhiar @a49ibrahim

xdevxy:
> Thank you for the adjustments @afugazzotto!
> 
> Unfortunately, and I am so sorry for this - I think I didn’t think this whole annotation thing through before. I am realizing that given that we want to do auto-healing, we want to overwrite the pipeline/ISBservice if somebody hand modifies it (like through kubectl apply). This means that we ultimately want to compare a. The spec we intend to apply (including “pause” designation) to b. the spec currently on the cluster. (We should vet that.) Therefore, there may be no value to storing the hash after all. :(
> 
> cc @xdevxy @dpadhiar @a49ibrahim

We can discuss this when we implement pipeline draining with concurrent concerns addressed. One way is to ignore the desired state in the spec when comparing hash and have dedicated logic to control the desired lifecycle of pipeline.

juliev0:
> One way is to ignore the desired state in the spec when comparing hash and have dedicated logic to control the desired lifecycle of pipeline.

Can you clarify what you mean by "ignore the desired state in the spec"? You mean use the hash only for the purpose of auto-healing? (to compare the last applied state to what's running on the cluster?) 

It seems to me that having the hash at all may be inefficient if we need to ultimately compare the desired state to what's running, but I could be wrong. Please do document the design you have in mind when you can.

xdevxy:
> > One way is to ignore the desired state in the spec when comparing hash and have dedicated logic to control the desired lifecycle of pipeline.
> 
> Can you clarify what you mean by "ignore the desired state in the spec"? You mean use the hash only for the purpose of auto-healing? (to compare the last applied state to what's running on the cluster?)
> 
> It seems to me that having the hash at all may be inefficient if we need to ultimately compare the desired state to what's running, but I could be wrong. Please do document the design you have in mind when you can.



> > One way is to ignore the desired state in the spec when comparing hash and have dedicated logic to control the desired lifecycle of pipeline.
> 
> Can you clarify what you mean by "ignore the desired state in the spec"? You mean use the hash only for the purpose of auto-healing? (to compare the last applied state to what's running on the cluster?)
> 
> It seems to me that having the hash at all may be inefficient if we need to ultimately compare the desired state to what's running, but I could be wrong. Please do document the design you have in mind when you can.



> > One way is to ignore the desired state in the spec when comparing hash and have dedicated logic to control the desired lifecycle of pipeline.
> 
> Can you clarify what you mean by "ignore the desired state in the spec"? You mean use the hash only for the purpose of auto-healing? (to compare the last applied state to what's running on the cluster?)
> 
> It seems to me that having the hash at all may be inefficient if we need to ultimately compare the desired state to what's running, but I could be wrong. Please do document the design you have in mind when you can.

It's a good point we need to also consider auto healing logic. What I mean is we probably do no even want desired lifecycle to be included in the pipelinerollout spec but only pipeline spec. We need to compare the hash for both auto healing and rollout spec change. @afugazzotto is working on that for auto healing task.

afugazzotto:
@juliev0 here is the issue I'm working on https://github.com/numaproj/numaplane/issues/64
It has more details on the idea to have both auto healing and hash of spec.

juliev0:
> #64

thank you. I have responded over there.

Issue #58, opened 2024-06-14
Script to generate Numaplane Kustomize Transformer config starting from the equivalent file on Numaplane


afugazzotto:
# Summary

If the [Kustomize Transformer config file](https://github.com/numaproj/numaflow/blob/main/docs/user-guide/reference/kustomize/numaflow-transformer-config.yaml) is updated on Numaflow, we want to automate as much as possible updating the equivalent file on the Numaplane repo.
We should create a script that downloads the file from the Numaflow repo, modifies its content (paths and kinds), and replaces the local file.
An entry to the Makefile to execute this script should be added.
The script could be executed manually and then perform a manual git commit OR we could include it in a GitHub Actions workflow to periodically update or in CI.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #59, opened 2024-06-14
Kustomize Transformer Config

afugazzotto:
### Modifications

Created the Kustomize Transformer config file starting from the [Numaflow equivalent file](https://github.com/numaproj/numaflow/blob/main/docs/user-guide/reference/kustomize/numaflow-transformer-config.yaml). Also, added some sample files for testing and documentation.

**NOTE:** if changes are made to the Numaflow Kustomize Transformer config file, we need to manually update the equivalent file on this repo. If we want to automate the task, I created this issue to do in the future: https://github.com/numaproj/numaplane/issues/58

### Verification

Manually tested by using the kustomize binary and `kubectl apply -k` command on the `config/kustomize/examples/transformer` directory


juliev0:
Interesting, I didn't know about kustomize transformer configs. 

Is this something that has been discussed with others regarding our process on the Numaportal side? Since we are supporting multiple versions of Numaflow at the same time, does it make sense to have this? (as opposed to maybe some other process where the user first creates a Pipeline spec of a given version and then encapsulates it into a Rollout CRD)

I would imagine in Numaflow that there's some codegen automation that creates this file. I don't think I want to add something like this if we don't have all of the automation that would be needed to accompany it.

cc'ing @whynowy

xdevxy (approved):

afugazzotto:
@juliev0 on the Numaflow side I did not see any automation in place to update this file. Last few changes to it seem to have been performed manually (https://github.com/numaproj/numaflow/commits/d72bd714b9f89236650c71152317f5acbb234785/docs/user-guide/reference/kustomize/numaflow-transformer-config.yaml). I created an issue https://github.com/numaproj/numaplane/issues/58 to add automation to generate this file on Numaplane side if necessary to have a more complete solution, but I'm not sure it is really needed at this time.

juliev0:
> Last few changes to it seem to have been performed manually (https://github.com/numaproj/numaflow/commits/d72bd714b9f89236650c71152317f5acbb234785/docs/user-guide/reference/kustomize/numaflow-transformer-config.yaml). 

Thank you for finding this. @whynowy What is the process currently for updating that file?

So, if a given user is using Numaflow version 1.0 and our file is updated to include all of the fields for Numaflow version 2.0 (perhaps we can assume Numaflow will only ever add new fields and not remove them? but I assume array indices could change (but maybe there aren't any in there?)) will there be any issues?

afugazzotto:
> > Last few changes to it seem to have been performed manually (https://github.com/numaproj/numaflow/commits/d72bd714b9f89236650c71152317f5acbb234785/docs/user-guide/reference/kustomize/numaflow-transformer-config.yaml).
> 
> Thank you for finding this. @whynowy What is the process currently for updating that file?
> 
> So, if a given user is using Numaflow version 1.0 and our file is updated to include all of the fields for Numaflow version 2.0 (perhaps we can assume Numaflow will only ever add new fields and not remove them? but I assume array indices could change (but maybe there aren't any in there?)) will there be any issues?

If versioning is a concern, we could make versioned releases of this file generated based on the specified "version" (or commit hash) of the equivalent file on the Numaflow repo. Maybe we could create versions at certain points in time.

PR #62, opened 2024-06-17
fix: avoid error on status update

### LINKED ISSUE #61: PipelineRollout and ISBServiceRollout Status Update Error
**Describe the bug**
When updating the status of PipelineRollout and ISBServiceRollout, we get the following error:
```
Operation cannot be fulfilled on isbservicerollouts.numaplane.numaproj.io \"isbservicerollout-test\": the object has been modified; please apply your changes to the latest version and try again
```
The error is not always shown, but most of the times.

**To Reproduce**
Running unit tests will show the error most of the times.

**Expected behavior**
The error message should not be shown and status should be updated as expected.

---

afugazzotto:
Fixes #61 

### Modifications

Implemented `RetryOnConflict` to get the latest version of the resource and update it's status with the status of the in-memory object modified during reconciliation.

### Verification

Running unit tests multiple times will no longer show the error mentioned in the issue.

### DIFF
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -23,6 +23,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


afugazzotto:
I tested a version of our code without performing any updates to the rollout objects nor updating the in-memory objects and the error was still showing up. The code was simply reading the rollout CR, create the child resource, and issue rollout status update (with no changes to it). There may be a different solution to solve this issue given the above observations.

juliev0 (approved):

xdevxy (approved):

PR #66, opened 2024-06-18
Auto healing

### LINKED ISSUE #64: Spec Hash Annotation and AutoHealing
# Summary

Pipeline/ISBService need to auto heal if changed manually even if rollouts are unchanged by comparing annotation hash of spec.

# Details

To avoid performing multiple expensive `update` calls on reconciliation, we will keep track of the following 3 hash strings:
- `hashA`: previous PipelineRollout pipeline spec hash (the one stored in the PipelineRollout annotations)
- `hashB`: current PipelineRollout pipeline spec that we calculate at reconciliation
- `hashC`: hash of Pipeline spec on the cluster without the Lifecycle field (we need to calculate this everytime)

for the following scenarios:
- If `hashA` == `hashB` && `hashC` == `hashB`: there are no changes made to the PipelineRollout nor to the Pipeline; therefore, we do not need to perform any updates
- If `hashA` == `hashB` && `hashC` != `hashB`: there are no changes made to the PipelineRollout; however, the Pipeline on the cluster was changed. Therefore, we need to autoheal the Pipeline by applying the PipelineRollout pipeline spec to overwrite the Pipeline CR on the cluster.
- If `hashA` != `hashB` the PipelineRollout pipeline spec has changed and we need to apply it. Note: no need to check `hashC` in this case.

**NOTE:** instead of computing `hashC` by reading the Pipeline everytime and calculating the hash, we considered gettting the `hashC` from the Pipeline since Numaflow already stores it. However, this includes the Lifecycle field which does not work for us since we want to exclude pausing state from the hash.
Therefore, we can either:
- A. read the Pipeline CR and calculate `hashC` OR
- B. read the Pipeline CR and compare the spec directly against the PipelineRollout pipeline spec.

Case B may not be any more performant than case A since for both cases we need to read the pipeline and, because we are using RawExtension type, we will probably have to perform marshalling/unmarshalling operations to convert to map and then perform deep comparisons. Hashing and comparing strings may be more performant. Appropriate performance testing should be performed if this is a great concern.

---

afugazzotto:
Fixes #64 

### Modifications

Issue #64 details the modifications.

### Verification

Added new unit tests to cover the case in which a child resource changes compared to the owner resource.

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -407,6 +407,8 @@ func applyPipelineSpec(
 ) error {
 	numaLogger := logger.FromContext(ctx)
 
+	// TODO: compare the in-cluster Pipeline lifecycle state when implementing dedicated logic managing desired pipeline lifecycle.
+	// Ideally, create a function to perform various checks and return if the CR should be updated or not.
 	shouldUpdateCR := true
 	if currentRolloutSpecHash == pipelineRollout.Annotations[apiv1.KeyHash] && currentRolloutSpecHash == pipelineSpecHash {
 		shouldUpdateCR = false
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -189,7 +189,7 @@ func (r *PipelineRolloutReconciler) reconcile(
 	}
 
 	// Get the object to see if it exists
-	pipeline, err := kubernetes.GetResource(ctx, r.restConfig, &obj, "pipelines")
+	pipeline, err := kubernetes.GetCR(ctx, r.restConfig, &obj, "pipelines")
 	if err != nil {
 		// create object as it doesn't exist
 		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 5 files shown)


internal/util/kubernetes/kubernetes.go
```diff
@@ -237,7 +237,7 @@ func GetResource(
 // ApplyCRSpec either creates or updates an object identified by the RawExtension, using the new definition,
 // first checking to see if there's a difference in Spec before applying
 // TODO: use CreateCR and UpdateCR instead
-func ApplyCRSpec(ctx context.Context, restConfig *rest.Config, object *GenericObject, pluralName string, specHash string, rolloutHashAnnotation string) error {
+func ApplyCRSpec(ctx context.Context, restConfig *rest.Config, object *GenericObject, pluralName string, currentRolloutSpecHash string, previousRolloutSpecHash string, childSpecHash string) error {
```
juliev0, line 240: We don't have to do it in this PR (I'd be happy to refactor later), but I'd like to see this function go back to being more of a pure function that's not aware of hashes/rollouts and simply "applies CRs"

internal/controller/pipelinerollout_controller.go
```diff
@@ -196,9 +198,19 @@ func (r *PipelineRolloutReconciler) reconcile(
...
+		// Calculate child resource spec hash only if no changes are necessary by the rollout spec comparison
+		pipelineSpecHash := ""
+		if currentRolloutSpecHash == pipelineRollout.Annotations[apiv1.KeyHash] {
+			pipelineSpecHash, err = calculateChildSpecHash(pipeline)
+			if err != nil {
+				numaLogger.Errorf(err, "failed to calculate spec hash from Pipeline CR: %v", err)
+				return false, err
+			}
+		}
+
```
juliev0, line 213: looks like all of the lifecycle management below is being done regardless of whether we need to be doing an update

afugazzotto: That is correct. We still want to perform certain operations independently from having to also perform an update (ex: pipeline pausing logic, ApplyCRSpec in case of creation, processPipelineStatus).

internal/util/kubernetes/kubernetes.go
afugazzotto: refactored

internal/controller/pipelinerollout_controller.go
```diff
@@ -370,19 +402,25 @@ func applyPipelineSpec(
...
 	pipelineRollout *apiv1.PipelineRollout,
-	specHash string,
+	currentRolloutSpecHash string,
+	pipelineSpecHash string,
 ) error {
 	numaLogger := logger.FromContext(ctx)
 
+	shouldUpdateCR := true
+	if currentRolloutSpecHash == pipelineRollout.Annotations[apiv1.KeyHash] && currentRolloutSpecHash == pipelineSpecHash {
+		shouldUpdateCR = false
```
xdevxy, line 412: Can you add a TODO: compare in cluster lifeCycle state when having dedicated logic managing desired pipeline lifeCycle.

xdevxy (approved):

internal/controller/pipelinerollout_controller.go
juliev0: Sorry, I specifically was referring to the pipeline pausing logic - that should only pertain if we are trying to perform an update of the spec, shouldn't it?

internal/util/kubernetes/kubernetes.go
```diff
@@ -281,9 +281,7 @@ func ApplyCRSpec(ctx context.Context, restConfig *rest.Config, object *GenericOb
 	} else {
 		numaLogger.Debugf("found existing Resource definition for %s/%s: %+v", object.Namespace, object.Name, resource)
 
-		// If the existing annotation matches the new hash, then nothing to do: log and return
-		if rolloutHashAnnotation == specHash {
-			numaLogger.Debug("Rollout spec is unchanged (by spec hash comparison). No updates will be performed")
+		if !shouldUpdateCR {
```
juliev0, line 284: thanks for doing this!

internal/controller/pipelinerollout_controller.go
afugazzotto: Yes, in the next PRs (specifically when we will work on the updating feature) we should create functions to define when updates and pausing should be made based on the result of a function checking for specific constraints. In general, the logic below this line, and specifically that related to pausing, will be revisited.

PR #71, opened 2024-06-21
Install Numaflow CRDs as a depenency

xdevxy:
### Modifications

Since Numaplane is managing Numaflow, it is a dependency for sure. This PR includes the installation process with the Numaflow minimal CRDs.  If the destination environment already have Numaflow installed then this is just a noop. But it will be helpful for clusters that don't. 

### Verification
Verified in a local cluster

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -5,3 +5,5 @@ resources:
 - bases/numaplane.numaproj.io_pipelinerollouts.yaml
 - bases/numaplane.numaproj.io_numaflowcontrollerrollouts.yaml
 - bases/numaplane.numaproj.io_isbservicerollouts.yaml
+# install numaflow minimal CRDs as a dependency
+- https://github.com/numaproj/numaflow/config/advanced-install/minimal-crds?ref=stable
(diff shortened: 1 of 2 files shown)


afugazzotto (approved):

juliev0:
@xdevxy  Do we need this in our install.yaml? I just noticed that because this is here, and because we brought this file over to `numa-manifest-generator`, it is trying to install the numaflow CRDs there, which it should not, because that's part of the "numaflow" manifest installation

xdevxy:
yeah, after we merged the addon, I think this is no longer needed internally, but this may still needed for open source users. Will this bring any issues if numaplane manifest try to install numaflow CRD again?

juliev0:
> yeah, after we merged the addon, I think this is no longer needed internally, but this may still needed for open source users. Will this bring any issues if numaplane manifest try to install numaflow CRD again?

I definitely think it will create opportunity for error to have this installed by 2 things.

juliev0:
I guess I do see why you wanted to include it here - the idea was that somebody could basically deploy numaplane and have what they need. But if we do keep it over here, then maybe we should have a separate manifest that gets created which doesn't include them? cc @whynowy

xdevxy:
yeah, make sense to separate it out :)  do you mind creating an issue for it? Thanks!

PR #83, opened 2024-06-25
Removed hash of spec from annotations of rollouts

afugazzotto:
### Modifications

Removed hash of spec from annotations of rollouts.
Performing RawExtension comparison instead.

### Verification

Unit tests and manual test.


internal/controller/pipelinerollout_controller.go
```diff
@@ -208,16 +198,6 @@ func (r *PipelineRolloutReconciler) reconcile(
...
-		pipelineSpecHash, err = calculateChildSpecHash(pipeline)
-		if err != nil {
-			numaLogger.Errorf(err, "failed to calculate spec hash from Pipeline CR: %v", err)
-			return false, err
-		}
-	}
-
 	// If the pipeline already exists, first check if the pipeline status
 	// is pausing. If so, re-enqueue immediately.
 	pausing := isPipelinePausing(ctx, pipeline)
```
juliev0, line 203: This is similar to a comment I made before...all of this pausing logic - isn't it only needed in the case that our desired spec is different from our running spec?

That's why I mentioned I thought the logic would be like this:
```
If Pipeline doesn't exist, create it
else:
   if specs are different  <-- this is the only place we need a comparison and not later on
     perform lifecycle management (i.e. either pause pipeline, continue waiting for it to pause, or unpause and update spec)
processPipelineStatus()
```

But maybe there is something I'm missing for why we would need the pausing logic even in the case that there's no spec difference?

internal/controller/isbservicerollout_controller.go
```diff
@@ -215,13 +196,17 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 	return nil
 }
 
-func (r *ISBServiceRolloutReconciler) calculateChildSpecHash(ctx context.Context, obj *kubernetes.GenericObject) (string, error) {
+func (r *ISBServiceRolloutReconciler) shouldUpdateCR(ctx context.Context, obj *kubernetes.GenericObject) (bool, error) {
```
xdevxy, line 199: why not pass in the object directly and only check this when a CR exist?

internal/controller/pipelinerollout_controller.go
afugazzotto: Shouldn't pipeline pausing be an independent operation to having tp update the spec?

internal/controller/isbservicerollout_controller.go
afugazzotto: Good point. I'll actually move the entire comparison logic inside ApplyCRSpec since we are only comparing specs directly now.

internal/controller/pipelinerollout_controller.go
juliev0: Hmm, let me think about this. So, at the moment, we're only dealing with pausing when we need to update our spec (as opposed to pausing when controller or isbservice needs us to). But you're right that we need to be forward thinking and think about the other cases too.

In my mind it's like "either we need to update our spec, or controller or isbservice is requesting that we pause". So, the question is whether that conditional statement should be up here vs inside of `ApplyCRSpec()`

xdevxy (approved):

internal/controller/pipelinerollout_controller.go
juliev0: Perhaps we could leave it as is, and assume that all of this will get addressed when you make the changes for "multiple pausers"

juliev0: I think I'm good with it as is.

internal/util/kubernetes/kubernetes.go
```diff
@@ -281,7 +282,13 @@ func ApplyCRSpec(ctx context.Context, restConfig *rest.Config, object *GenericOb
 	} else {
 		numaLogger.Debugf("found existing Resource definition for %s/%s: %+v", object.Namespace, object.Name, resource)
 
-		if !shouldUpdateCR {
+		currentObj, err := UnstructuredToObject(resource)
+		if err != nil {
+			return fmt.Errorf("error attempting to convert unstructured resource to generic object: %v", err)
+		}
+
+		if reflect.DeepEqual(currentObj.Spec, object.Spec) {
```
juliev0, line 290: nit: can we rename "object" to something more informative?

juliev0: deleting my comment. I think it's fine actually

internal/controller/pipelinerollout_controller.go
afugazzotto: I'll leave it as is for now then. I'm sure adjustments will need to be made once working on the next set of changes we have discussed.

internal/controller/pipelinerollout_controller.go
```diff
@@ -434,8 +385,3 @@ func applyPipelineSpec(
 	processPipelineStatus(ctx, pipeline, pipelineRollout)
```
juliev0, line 385: this is a separate issue outside of this PR, but I think we should move this outside of this function. Processing the pipeline status is something we should do independent of applying a Pipeline spec.

juliev0: any objections if I move that?

afugazzotto: none from me, it make sense

PR #86, opened 2024-06-25
fix: relocate call to processPipelineStatus()

juliev0:
### Modifications

`processPipelineStatus()` was previously being called from the function `applyCRSpec()`, but really incorporating the child's status into ours is something we want to do independent of applying a spec.

### Verification

1. applied PipelineRollout with no ISBServiceRollout available
2. observed that Pipeline was unhealthy due to no ISBService, and therefore PipelineRollout was set to ChildResourcesHealthy=false
3. applied ISBServiceRollout
4. observed that Pipeline became healthy, and then PipelineRollout got set to ChildResourcesHealthy=true

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -166,7 +166,7 @@ func (r *PipelineRolloutReconciler) reconcile(
 		controllerutil.AddFinalizer(pipelineRollout, finalizerName)
 	}
 
-	obj := kubernetes.GenericObject{
+	newPipelineDef := kubernetes.GenericObject{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Pipeline",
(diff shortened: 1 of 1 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -197,28 +197,31 @@ func (r *PipelineRolloutReconciler) reconcile(
 		return false, fmt.Errorf("error getting Pipeline: %v", err)
 	}
 
+	// propagate the pipeline's status into PipelineRollout's status
+	processPipelineStatus(ctx, existingPipelineDef, pipelineRollout)
```
juliev0, line 201: I think it should actually be okay to do this early on in the function. Since we watch the underlying child resource, if a child's status changes based on any changes we make to the child's spec, that will make another call to `Reconcile()` and we will process it then.

afugazzotto (approved):

Issue #85, opened 2024-06-25
Numaflow Controller Rollout should set its Status to reflect child resources


juliev0:
# Summary

PipelineRollout and ISBServiceRollout both set a Condition to indicate if the child resources are healthy, but this doesn't exist for NumaflowControllerRollout.

(Note, this may need revisiting if child status should be done a different way, per [this](https://github.com/numaproj/numaplane/issues/72#issuecomment-2190040449) comment)

# Use Cases

When would you use this?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
Not sure why this wasn't closed automatically...

PR #92, opened 2024-07-01
Observed Generation

### LINKED ISSUE #88: Incorporate ObservedGeneration into Rollouts' Statuses
# Summary

When we return our health status to ArgoCD, we want to make sure that what we return isn't outdated. [This](https://alenkacz.medium.com/kubernetes-operator-best-practices-implementing-observedgeneration-250728868792) article describes the risk. If a spec has changed but hasn't yet been reconciled, the Status may look like everything is in a good "Running" state even though it hasn't been reconciled.

For each of our Rollout CRDs, we should include an `ObservedGeneration` field in our Status.

We should also make sure that if we're still in the process of updating our CR, our Phase should be "Pending" and not "Running" yet. (Maybe also change the name "Running" to "Deployed".) 

This way when we write our ArgoCD health check script, we can making it conditional that `Rollout.metadata.Generation==Rollout.status.observedGeneration && Rollout.status.phase=="Deployed" && Rollout.status.conditions[ChildResourcesHealthy]==true` (the first 2 conditions verify that we've incorporated the latest spec and also it's done being incorporated)

Consider that in order to truly decide we can be in "Deployed" state, ultimately we need to have confidence that our children have also been reconciled. Numaflow now has an [issue](https://github.com/numaproj/numaflow/issues/1773) that its Pipeline and ISBService need to include ObservedGeneration as well. So, if our child resource's generation matches its observedGeneration, we can set our own phase to "Deployed". 

Note that for Pipelines, since we first pause (drain) the pipeline prior to updating it and running, we will be in "Pending" state that whole time. 

Proposal linked [here](https://docs.google.com/document/d/1llAM4RL6t8FCpAWVUwJCeKd3WNZ_vr0OqJ29OmUxm_c/edit?usp=sharing)

---

afugazzotto:
Fixes #88 #85 

### Modifications

Added ObservedGeneration to the rollouts Status and setting Phase and ConditionConfigured based on child resource status.
NOTE: Numaflow needs to be updated to include ObservedGeneration status.

### Verification

Unable to test until Numaflow is updated to include ObservedGeneration status. Unit tests will fail until then.

### DIFF
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -95,6 +95,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	isbServiceRollout = isbServiceRolloutOrig.DeepCopy()
 
 	isbServiceRollout.Status.Init(isbServiceRollout.Generation)
+	// TODO: update status of CR throughout the reconciliation process (write a function to encapsulate the retry logic below)
 
 	err := r.reconcile(ctx, isbServiceRollout)
 	if err != nil {
@@ -187,6 +188,8 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 
 	processISBServiceStatus(ctx, isbsvc, isbServiceRollout)
 
+	isbServiceRollout.Status.MarkDeployed()
+
 	return nil
 }
 
@@ -237,11 +240,6 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
 		// this will have been set to Unknown in the call to InitConditions()
 	default:
 		rollout.Status.MarkChildResourcesHealthy()
-	}
-
-	// Set ISBServiceRollout CR status to Deployed only if the ISBService has completely reconciled: Generation == ObservedGeneration
-	if isbsvc.Generation == isbsvcStatus.ObservedGeneration {
-		rollout.Status.MarkDeployed()
 	}
 }
 
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -94,7 +94,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	isbServiceRolloutOrig := isbServiceRollout
 	isbServiceRollout = isbServiceRolloutOrig.DeepCopy()
 
-	isbServiceRollout.Status.Init()
+	isbServiceRollout.Status.Init(isbServiceRollout.Generation)
 
 	err := r.reconcile(ctx, isbServiceRollout)
 	if err != nil {
@@ -242,7 +242,6 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
 	// Set ISBServiceRollout CR status Phase and ObservedGeneration only if the ISBService has completely reconciled: Generation == ObservedGeneration
 	if isbsvc.Generation == isbsvcStatus.ObservedGeneration {
 		rollout.Status.MarkDeployed()
-		rollout.Status.SetObservedGeneration(rollout.Generation)
 	}
 }
 
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -240,10 +240,10 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
 	case numaflowv1.ISBSvcPhaseUnknown:
 		rollout.Status.MarkChildResourcesHealthUnknown("ISBSvcUnknown", "ISBService Phase Unknown", rollout.Generation)
 	default:
-		if rollout.Generation == isbsvcStatus.ObservedGeneration {
+		if isbsvc.Generation == isbsvcStatus.ObservedGeneration {
 			rollout.Status.MarkChildResourcesHealthy(rollout.Generation)
 		} else {
-			rollout.Status.MarkChildResourcesUnhealthy("GenerationMismatch", "ISBService Generation mismatch the ObservedGeneration", rollout.Generation)
+			rollout.Status.MarkChildResourcesUnhealthy("Progressing", "Mismatch between ISBService Generation and ObservedGeneration", rollout.Generation)
 		}
 	}
 }
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -174,7 +174,7 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 	}
 
 	// after the Apply, Get the ISBService so that we can propagate its health into our Status
-	existingISBService, err := kubernetes.GetCR(ctx, r.restConfig, &obj, "interstepbufferservices")
+	isbsvc, err := kubernetes.GetCR(ctx, r.restConfig, &obj, "interstepbufferservices")
 	if err != nil {
 		numaLogger.Errorf(err, "failed to get ISBServices: %v", err)
 		return err
@@ -185,7 +185,7 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		return fmt.Errorf("failed to apply PodDisruptionBudget for ISBServiceRollout %s, err: %v", isbServiceRollout.Name, err)
 	}
 
-	processISBServiceStatus(ctx, existingISBService, isbServiceRollout)
+	processISBServiceStatus(ctx, isbsvc, isbServiceRollout)
 
 	return nil
 }
@@ -239,7 +239,7 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
 		rollout.Status.MarkChildResourcesHealthy()
 	}
 
-	// Set ISBServiceRollout CR status Phase only if the ISBService has completely reconciled: Generation == ObservedGeneration
+	// Set ISBServiceRollout CR status to Deployed only if the ISBService has completely reconciled: Generation == ObservedGeneration
 	if isbsvc.Generation == isbsvcStatus.ObservedGeneration {
 		rollout.Status.MarkDeployed()
 	}
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -240,7 +240,10 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
 	case numaflowv1.ISBSvcPhaseUnknown:
 		rollout.Status.MarkChildResourcesHealthUnknown("ISBSvcUnknown", "ISBService Phase Unknown", rollout.Generation)
 	default:
-		if isbsvc.Generation == isbsvcStatus.ObservedGeneration {
+		// NOTE: this assumes that Numaflow default ObservedGeneration is -1
+		// `isbsvcStatus.ObservedGeneration == 0` is used to avoid backward compatibility
+		// issues for Numaflow versions that do not have ObservedGeneration
+		if isbsvcStatus.ObservedGeneration == 0 || isbsvc.Generation <= isbsvcStatus.ObservedGeneration {
 			rollout.Status.MarkChildResourcesHealthy(rollout.Generation)
 		} else {
 			rollout.Status.MarkChildResourcesUnhealthy("Progressing", "Mismatch between ISBService Generation and ObservedGeneration", rollout.Generation)
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -174,7 +174,7 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 	}
 
 	// after the Apply, Get the ISBService so that we can propagate its health into our Status
-	isbsvc, err := kubernetes.GetCR(ctx, r.restConfig, &obj, "interstepbufferservices")
+	existingISBService, err := kubernetes.GetCR(ctx, r.restConfig, &obj, "interstepbufferservices")
 	if err != nil {
 		numaLogger.Errorf(err, "failed to get ISBServices: %v", err)
 		return err
@@ -185,9 +185,17 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		return fmt.Errorf("failed to apply PodDisruptionBudget for ISBServiceRollout %s, err: %v", isbServiceRollout.Name, err)
 	}
 
-	processISBServiceStatus(ctx, isbsvc, isbServiceRollout)
+	processISBServiceStatus(ctx, existingISBService, isbServiceRollout)
 
-	isbServiceRollout.Status.MarkRunning()
+	// Set ISBServiceRollout CR status Phase and ObservedGeneration only if the ISBService has completely reconciled: Generation == ObservedGeneration
+	existingISBServiceStatus, err := kubernetes.ParseStatus(existingISBService)
+	if err != nil {
+		return fmt.Errorf("unable to parse status for existing ISBService %s/%s: %v", existingISBService.Namespace, existingISBService.Name, err)
+	}
+	if existingISBService.Generation == existingISBServiceStatus.ObservedGeneration {
+		isbServiceRollout.Status.MarkDeployed()
+		isbServiceRollout.Status.SetObservedGeneration(isbServiceRollout.Generation)
+	}
 
 	return nil
 }
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -94,7 +94,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	isbServiceRolloutOrig := isbServiceRollout
 	isbServiceRollout = isbServiceRolloutOrig.DeepCopy()
 
-	isbServiceRollout.Status.InitConditions()
+	isbServiceRollout.Status.Init()
 
 	err := r.reconcile(ctx, isbServiceRollout)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/controller/isbservicerollout_controller.go
```diff
@@ -96,7 +96,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	isbServiceRolloutOrig := isbServiceRollout
 	isbServiceRollout = isbServiceRolloutOrig.DeepCopy()
 
-	isbServiceRollout.Status.InitConditions()
+	isbServiceRollout.Status.Init(isbServiceRollout.Generation)
```
juliev0, line 99: I like this

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -367,6 +374,46 @@ func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.
...
+	// Parse the status of the existing Numaflow Controller Deployment
+	existingDeploymentStatus, err := kubernetes.ParseStatus(existingDeployment)
+	if err != nil {
+		return fmt.Errorf("unable to parse status for existing Numaflow Controller Deployment %s/%s: %v", existingDeployment.Namespace, existingDeployment.Name, err)
+	}
+
+	// Set NumaflowControllerRollout CR status Phase and ObservedGeneration only if the Numaflow Controller Deployment has completely reconciled (Generation == ObservedGeneration)
+	if existingDeployment.Generation == existingDeploymentStatus.ObservedGeneration {
+		controllerRollout.Status.MarkDeployed()
+		controllerRollout.Status.SetObservedGeneration(controllerRollout.Generation)
```
juliev0, line 411: Is this line needed here? Did you mean to remove it like you did from the other reconcilers?

afugazzotto: Forgot to remove that. Thanks

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -364,6 +374,53 @@ func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.
...
+	var existingDeploymentStatus appsv1.DeploymentStatus
+	if len(existingDeployment.Status.Raw) > 0 {
+		err = json.Unmarshal(existingDeployment.Status.Raw, &existingDeploymentStatus)
+		if err != nil {
+			return fmt.Errorf("unable to parse status for existing Numaflow Controller Deployment %s/%s: %v", existingDeployment.Namespace, existingDeployment.Name, err)
+		}
+	}
+
+	if existingDeployment.Generation == existingDeploymentStatus.ObservedGeneration {
+		if existingDeploymentStatus.Replicas == existingDeploymentStatus.ReadyReplicas {
```
xdevxy, line 412: I think check the complete condition is the right way to tell the deployment is completed successful. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -27,19 +27,26 @@ import (
...
-	ConditionChildResourcesHealthy ConditionType = "ChildResourcesHealthy"
+	// PhaseFailed indicates that one or more errors have occurred during reconciliation.
+	PhaseFailed Phase = "Failed"
+
+	// ConditionChildResourceHealthy indicates if the child resource is in a healthy phase (ex: not pending, not paused, not deleting, etc.).
+	// Child health is set to True only if the child resource generation is equal to its own observedGeneration.
+	ConditionChildResourceHealthy ConditionType = "ChildResourcesHealthy"
+
+	// ConditionChildResourceDeployed indicates that the child resource was deployed.
+	ConditionChildResourceDeployed ConditionType = "ChildResourceDeployed"
```
xdevxy, line 49: We are going to add pausing/paused condition to pipelinerollout, right?

afugazzotto: At this time, if the pipeline is pausing or paused I set the child resource condition as unhealthy [here](https://github.com/numaproj/numaplane/blob/observedGeneration/internal/controller/pipelinerollout_controller.go#L283). If we want to show in ArgoCD the health as `suspended` similar to [jobs](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#job) using custom LUA script, then I can create a condition for the pipeline status to say a pipeline is paused or pausing and that can be used in the LUA script.

internal/controller/numaflowcontrollerrollout_controller.go
afugazzotto: From the document:
> The condition holds even when availability of replicas changes (which does instead affect the Available condition).
Should we check that both the `Available` and `Progressing` conditions are set to `True`?

pkg/apis/numaplane/v1alpha1/status_types.go
afugazzotto: Introduced pause or pausing condition.

internal/controller/numaflowcontrollerrollout_controller.go
afugazzotto: Used similar code as that used by ArgoProj gitops-engine as discussed with @xdevxy.

internal/controller/isbservicerollout_controller.go
```diff
@@ -234,11 +234,17 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
...
+		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcPending", "ISBService Pending", rollout.Generation)
 	case numaflowv1.ISBSvcPhaseUnknown:
-		// this will have been set to Unknown in the call to InitConditions()
+		rollout.Status.MarkChildResourcesHealthUnknown("ISBSvcUnknown", "ISBService Phase Unknown", rollout.Generation)
 	default:
-		rollout.Status.MarkChildResourcesHealthy()
+		if rollout.Generation == isbsvcStatus.ObservedGeneration {
+			rollout.Status.MarkChildResourcesHealthy(rollout.Generation)
+		} else {
+			rollout.Status.MarkChildResourcesUnhealthy("GenerationMismatch", "ISBService Generation mismatch the ObservedGeneration", rollout.Generation)
```
juliev0, line 246: it looks like in the NumaflowControllerRollout reconciler we call this "Progressing" but call it "GenerationMismatch" here?

juliev0 (commented):
made one comment, but otherwise this is looking good to me, thanks

internal/controller/isbservicerollout_controller.go
afugazzotto: Good catch. I think we should use `Progressing` for all instead of `GenerationMismatch` unless we think that there could be a case in which the generation and observedGeneration will never match. What do you think?

juliev0: I think we should use Progressing as well. Keep it standard with what `Deployments` do.

internal/controller/pipelinerollout_controller.go
```diff
@@ -270,14 +272,32 @@ func processPipelineStatus(ctx context.Context, pipeline *kubernetes.GenericObje
...
 	pipelinePhase := numaflowv1.PipelinePhase(pipelineStatus.Phase)
 	switch pipelinePhase {
 	case numaflowv1.PipelinePhaseFailed:
-		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed")
+		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed", pipelineRollout.Generation)
+	case numaflowv1.PipelinePhasePaused, numaflowv1.PipelinePhasePausing:
+		// TODO: update string(pipelinePhase) when working on strng comparison changes
+		pipelineRollout.Status.MarkPipelinePausingOrPausedWithReason(string(pipelinePhase), pipelineRollout.Generation)
+		if pipeline.Generation == pipelineStatus.ObservedGeneration {
+			pipelineRollout.Status.MarkChildResourcesHealthy(pipelineRollout.Generation)
```
juliev0, line 287: So, it looks like if it's paused or pausing, but the generation of the child matches its observedGeneration, we mark it healthy. Personally, I prefer that. But want to verify @xdevxy is okay with that, since it seems like her document proposed it would be unhealthy, right?

xdevxy: I don't feel strong about this as long as we have a condition to indicate paused/pausing specifically. However if we go with this logic, this needs to be updated https://github.com/numaproj/numaplane/pull/92/files/7b887a13f05abde419720205788e4d325d65f39a#diff-e8e263da81dda7f4a11f5070b4e4c2571f119bbd43ac576a8d32f3af098d7ae7R44-R45 accordingly. My doc was just referring to this.

afugazzotto: I will update that comment

internal/controller/isbservicerollout_controller.go
```diff
@@ -234,11 +234,17 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
...
-		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcFailed", "ISBService Failed")
+		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcFailed", "ISBService Failed", rollout.Generation)
+	case numaflowv1.ISBSvcPhasePending:
+		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcPending", "ISBService Pending", rollout.Generation)
 	case numaflowv1.ISBSvcPhaseUnknown:
-		// this will have been set to Unknown in the call to InitConditions()
+		rollout.Status.MarkChildResourcesHealthUnknown("ISBSvcUnknown", "ISBService Phase Unknown", rollout.Generation)
 	default:
-		rollout.Status.MarkChildResourcesHealthy()
+		if isbsvc.Generation == isbsvcStatus.ObservedGeneration {
```
xdevxy, line 243: why not checking ObservedGeneration is not empty first? Since it can be absent for existing numaflow versions.

juliev0: I'm a little confused here - I thought we didn't need to worry about backward compatibility for older versions of Numaflow after all?

xdevxy: sorry for the confusion, no worry about backward compatibility for older version to me means: 1) we don't need to ensure we can correctly detect if the changes has been applied for existing numaflow versions that does not have `observedGeneration` in status yet 2) we should not introduce any bugs if someone is running existing numaflow versions. If we have this current checking, with existing numaflow version it will never be healthy, right?

internal/controller/pipelinerollout_controller.go
```diff
@@ -270,14 +272,32 @@ func processPipelineStatus(ctx context.Context, pipeline *kubernetes.GenericObje
...
+
 	pipelinePhase := numaflowv1.PipelinePhase(pipelineStatus.Phase)
 	switch pipelinePhase {
 	case numaflowv1.PipelinePhaseFailed:
-		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed")
+		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed", pipelineRollout.Generation)
+	case numaflowv1.PipelinePhasePaused, numaflowv1.PipelinePhasePausing:
+		// TODO: update string(pipelinePhase) when working on strng comparison changes
+		pipelineRollout.Status.MarkPipelinePausingOrPausedWithReason(string(pipelinePhase), pipelineRollout.Generation)
+		if pipeline.Generation == pipelineStatus.ObservedGeneration {
```
xdevxy, line 286: we should use `<=` whenever we compare `observedGeneration` to follow the practice from k8s

xdevxy (approved):

internal/controller/isbservicerollout_controller.go
juliev0: In that case, I guess the code won't truly be able to detect the difference between "this version of numaflow doesn't support observedGeneration" and "Numaflow Controller slow/down" for a newly created resource

internal/controller/pipelinerollout_controller.go
juliev0: Do we know why there could ever be a case where `generation < observedGeneration`?

internal/controller/isbservicerollout_controller.go
afugazzotto: @xdevxy and I discussed and we think that [this](https://github.com/numaproj/numaplane/pull/92/commits/2728f241b2c600322e76d4924fae524e2fb12573) should address that concern. The comment explains that we assume that Numaflow will set default ObservedGeneration to -1. So, when the generation goes from 0 to 1, the ObservedGeneration will be set to 1 skipping the zero value.

internal/controller/pipelinerollout_controller.go
afugazzotto: This is related to [this comment](https://github.com/numaproj/numaplane/pull/92#discussion_r1667089759)

internal/controller/isbservicerollout_controller.go
xdevxy: which code you are referring to? this one? https://github.com/numaproj/numaplane/pull/92/files#diff-9a10dece0454f1f44e2e6076e2b3ada6a08a66de674f546276972944ec61be9aR265-R268. Only "version of numaflow doesn't support observedGeneration" can ever be 0 in this case, right?

internal/controller/pipelinerollout_controller.go
xdevxy: I think it is an extra protection that `ObservedGeneration` set to a number higher than `generation` after reconciliation.

internal/controller/isbservicerollout_controller.go
```diff
@@ -234,11 +234,20 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
...
-		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcFailed", "ISBService Failed")
+		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcFailed", "ISBService Failed", rollout.Generation)
+	case numaflowv1.ISBSvcPhasePending:
+		rollout.Status.MarkChildResourcesUnhealthy("ISBSvcPending", "ISBService Pending", rollout.Generation)
 	case numaflowv1.ISBSvcPhaseUnknown:
-		// this will have been set to Unknown in the call to InitConditions()
+		rollout.Status.MarkChildResourcesHealthUnknown("ISBSvcUnknown", "ISBService Phase Unknown", rollout.Generation)
 	default:
-		rollout.Status.MarkChildResourcesHealthy()
+		// NOTE: this assumes that Numaflow default ObservedGeneration is -1
```
juliev0, line 243: sorry, where is the -1 being set?

afugazzotto: This will be done in Numaflow. @xdevxy is making those changes.

internal/controller/isbservicerollout_controller.go
juliev0: > @xdevxy and I discussed and we think that [this](https://github.com/numaproj/numaplane/pull/92/commits/2728f241b2c600322e76d4924fae524e2fb12573) should address that concern. The comment explains that we assume that Numaflow will set default ObservedGeneration to -1. So, when the generation goes from 0 to 1, the ObservedGeneration will be set to 1 skipping the zero value.

That is a creative idea. So, this would be the default version set in the CRD itself I guess?

xdevxy: yes

juliev0 (approved):

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -364,6 +374,90 @@ func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.
...
+
+	// Try to get the Numaflow Controller Deployment
+	deplObj := kubernetes.GenericObject{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Deployment",
+			APIVersion: "apps/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "numaflow-controller",
+			Namespace: controllerRollout.Namespace,
```
qhuai, line 397: Is there a way to avoid hardcoding the strings? Are they available in the installed CRDs?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -364,6 +374,90 @@ func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.
...
+		cond := getDeploymentCondition(deploymentStatus, appsv1.DeploymentProgressing)
+		if cond != nil && cond.Reason == "ProgressDeadlineExceeded" {
+			msg := fmt.Sprintf("Deployment %q exceeded its progress deadline", deployment.Name)
+			controllerRollout.Status.MarkChildResourcesUnhealthy("Degraded", msg, controllerRollout.Generation)
+			return nil
+		} else if deploymentSpec.Replicas != nil && deploymentStatus.UpdatedReplicas < *deploymentSpec.Replicas {
+			msg := fmt.Sprintf("Waiting for Deployment rollout to finish: %d out of %d new replicas have been updated...", deploymentStatus.UpdatedReplicas, *deploymentSpec.Replicas)
+			controllerRollout.Status.MarkChildResourcesUnhealthy("Progressing", msg, controllerRollout.Generation)
+			return nil
+		} else if deploymentStatus.Replicas > deploymentStatus.UpdatedReplicas {
```
qhuai, line 441: We may never encounter this issue but if deploymentStatus.Replicas could ever be "nil" - there is a check in line #437, will this comparison trigger an exception?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -364,6 +374,90 @@ func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.
...
+	var deploymentStatus appsv1.DeploymentStatus
+	if len(deployment.Status.Raw) > 0 {
+		err = json.Unmarshal(deployment.Status.Raw, &deploymentStatus)
+		if err != nil {
+			return fmt.Errorf("unable to parse status for existing Numaflow Controller Deployment %s/%s: %v", deployment.Namespace, deployment.Name, err)
+		}
+	}
+
+	// Health Check borrowed from argoproj/gitops-engine/pkg/health/health_deployment.go https://github.com/argoproj/gitops-engine/blob/master/pkg/health/health_deployment.go#L27
+	if deployment.Generation <= deploymentStatus.ObservedGeneration {
```
qhuai, line 431: This `<=` comparison is probably safe to do. However, in what scenario would generation < observed generation? Due to potential k8s bugs?

afugazzotto: [This](https://github.com/numaproj/numaplane/pull/92#discussion_r1667084952) and [this](https://github.com/numaproj/numaplane/pull/92#discussion_r1667089759) explain why we decided to use `<=`.

internal/controller/numaflowcontrollerrollout_controller.go
afugazzotto: There is already a [constant](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontrollerrollout_controller.go#L63) for it, I forgot to use it. Will fix with my next PR. Thanks for catching this.

Issue #10, opened 2024-05-29
Fix - log level change should propagate to syncer


juliev0:

juliev0:
@shubhamdixit863 do you want to try this? I think we may need to call [getCluster()](https://github.com/numaproj/numaplane/blob/v0.1.0/internal/sync/cache.go#L137) if either of these values change.

If `includedResources` changes, [this call to loadCacheSettings()](https://github.com/numaproj/numaplane/blob/v0.1.0/internal/sync/cache.go#L139) would have a different result. 
If `logLevel` changes, we need to pass [this](https://github.com/numaproj/numaplane/blob/v0.1.0/internal/sync/cache.go#L184) in again. 

As part of this PR, make sure that if you modify those values, everything is still working correctly.

PR #93, opened 2024-07-01
Add callback functionality for the numaplane-controller-config ConfigMap

juliev0:
### Modifications

Currently, we can ask the ConfigManager for the latest version of the ConfigMap. This works when we have a goroutine that's running that needs to get the latest and greatest values.

For cases in which we have some persistent state that's created based on value(s) in the ConfigMap, it could be nice to have a callback called to update that state.

This PR takes care of updating the Base Logger used by Numaplane through a callback.

Other remaining cases include:
- updating the `includedResources` passed to gitops-engine
- updating the Log Level that gitops-engine is using

### Verification

While running, updated LogLevel=4 (debug) and observed that take effect in the log

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -146,6 +146,7 @@ func (cm *ConfigManager) loadGlobalConfig(
 		}
 		cm.config = &newConfig
 
+		// call any registered callbacks
 		for _, f := range cm.callbacks {
 			f(*cm.config)
 		}
@@ -167,6 +168,7 @@ func CloneWithSerialization[T NumaflowControllerDefinitionConfig | GlobalConfig]
 	return &clone, nil
 }
 
+// RegisterCallback adds a callback to be called when the config changes
 func (cm *ConfigManager) RegisterCallback(f func(config GlobalConfig)) {
 	cm.lock.Lock()
 	defer cm.lock.Unlock()
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -12,7 +12,15 @@ import (
 )
 
 type ConfigManager struct {
-	config        *GlobalConfig
+	config *GlobalConfig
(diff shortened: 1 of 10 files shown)


internal/controller/config/config.go
```diff
@@ -12,7 +12,15 @@ import (
 )
 
 type ConfigManager struct {
-	config        *GlobalConfig
+	config *GlobalConfig
+	lock   *sync.RWMutex
+	// if the configmap changes, these callbacks will be called
+	callbacks []func(config GlobalConfig)
+
+	controllerDefMgr ControllerDefinitionsManager
```
juliev0, line 20: moved this into its own struct so it can have its own lock

internal/util/kubernetes/kubernetes.go
```diff
@@ -274,7 +274,7 @@ func ApplyCRSpec(ctx context.Context, restConfig *rest.Config, object *GenericOb
 			if err != nil {
 				return fmt.Errorf("failed to create Resource %s/%s, err=%v", object.Namespace, object.Name, err)
 			}
-			numaLogger.Debugf("successfully created resource %s/%s", object.Namespace, object.Name)
+			numaLogger.Infof("successfully created resource %s/%s", object.Namespace, object.Name)
```
juliev0, line 277: wanted to add at least one line that would be called at Info level when we update something (here for new resource and below for updated resource)

afugazzotto (approved):

xdevxy:
@juliev0 do you mind creating issues to track the remaining tasks? Thanks!

juliev0:
> @juliev0 do you mind creating issues to track the remaining tasks? Thanks!

So, I modified the original issue [here](https://github.com/numaproj/numaplane/issues/10) to encompass what's left to be done.

PR #101, opened 2024-07-06
add modularity to kubernetes functions

juliev0:
### Modifications

All of the Kubernetes-related functions were in the same file, which was getting long. I wanted to create some separation to put into separate files:
- generic code related to using the dynamic client to get/update/create resources 
- generic utilities for interacting with unstructured objects
- generic utilities for interacting with common Kubernetes types (structured)

Also, I don't think we should put code related to specific Rollout types in this directory - rather those can be collocated with the Rollout itself.


internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -232,6 +234,68 @@ func applyOwnershipToManifests(manifests []string, controllerRollout *apiv1.Numa
 	return manifestsWithOwnership, nil
 }
 
+func applyOwnership(manifest string, controllerRollout *apiv1.NumaflowControllerRollout) ([]byte, error) {
```
juliev0, line 237: this is specific to NumaflowControllerRollout - I'd rather keep it collocated

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -513,7 +577,7 @@ func toUnstructuredAndApplyLabel(manifests []string, name string) ([]*unstructur
 			return nil, err
 		}
 		target := &unstructured.Unstructured{Object: obj}
-		err = kubernetes.SetNumaplaneInstanceLabel(target, common.LabelKeyNumaplaneInstance, name)
+		err = kubernetes.SetLabel(target, common.LabelKeyNumaplaneInstance, name)
```
juliev0, line 580: there's nothing about this function that's specific to Numaplane Instance label

afugazzotto (approved):
Thank you for reorganizing. I like these ideas.

PR #102, opened 2024-07-07
Incorporate RateLimitingInterface to be invoked by PipelineRollout Reconcile()

juliev0:
### Modifications

This is a first step to doing pause concurrency through [this](https://docs.google.com/document/d/1XXibcR5B4qEPUplEyQ7RRXl0gQp1xDQnF40BGuyWU1Q/edit) model.

We essentially need a queue to process PipelineRollout CRs so that it can be invoked by `PipelineRollout.Reconcile()`, and by `NumaflowControllerRollout.Reconcile()` and `ISBServiceRollout.Reconcile()` whenever they need to make or remove a pause request.

### Verification

Caused `PipelineRollout.Reconcile()` to be called different ways on a PipelineRollout and observed they did the right thing.

Also, verified proper shutdown by watching the Pod log as I did a `kubectl delete pod` on the Numaplane Controller Pod.


internal/util/kubernetes/dynamic_client.go
```diff
@@ -187,7 +187,7 @@ func CreateCR(
 	if err != nil {
 		return fmt.Errorf("failed to create Resource %s/%s, err=%v", object.Namespace, object.Name, err)
 	}
-	numaLogger.Debugf("successfully created resource %s/%s", object.Namespace, object.Name)
+	numaLogger.Infof("successfully created resource %s/%s", object.Namespace, object.Name)
```
juliev0, line 190: want to have at least one "Info" level line for whenever we create the child resource. Previously did this for the `ApplyCRSpec()` function - didn't realize I needed to do it here too

cmd/main.go
```diff
@@ -164,6 +164,7 @@ func main() {
 	if err = pipelineRolloutReconciler.SetupWithManager(mgr); err != nil {
 		numaLogger.Fatal(err, "Unable to set up PipelineRollout controller")
 	}
+	defer pipelineRolloutReconciler.Shutdown()
```
juliev0, line 167: graceful shutdown of the workers using the queue

internal/controller/pipelinerollout_controller.go
```diff
@@ -145,6 +183,80 @@ func (r *PipelineRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	return ctrl.Result{}, nil
 }
 
+func (r *PipelineRolloutReconciler) Shutdown() {
+	numaLogger := logger.GetBaseLogger().WithName(loggerName)
```
afugazzotto, line 187: Why not getting the logger from context?

internal/controller/suite_test.go
```diff
@@ -127,18 +127,16 @@ var _ = BeforeSuite(func() {
...
 
-	err = (&PipelineRolloutReconciler{
-		client:     k8sManager.GetClient(),
-		scheme:     k8sManager.GetScheme(),
-		restConfig: cfg,
-	}).SetupWithManager(k8sManager)
+	err = NewPipelineRolloutReconciler(
+		k8sManager.GetClient(),
+		k8sManager.GetScheme(),
+		cfg).SetupWithManager(k8sManager)
```
afugazzotto, line 133: I preferred not to call `NewPipelineRolloutReconciler` to be able to create a more custom reconciler for testing (ex: with less workers, etc.). However, I am not opposed to this since it reduces the amount of testing code we have to write. It's a bit of a tradeoff. What are you reasons behind this choice?

internal/controller/pipelinerollout_controller.go
```diff
@@ -44,25 +49,49 @@ import (
...
+
+	r := &PipelineRolloutReconciler{
 		client,
 		s,
 		restConfig,
+		pipelineRolloutQueue,
+		&sync.WaitGroup{},
 	}
+
+	go r.runWorkers(ctx)
```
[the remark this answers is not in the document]
juliev0: good call. Originally, I had this waiting on a context to be shutdown but I changed that logic so it doesn't need to be a goroutine anymore.

internal/controller/suite_test.go
juliev0: `NewPipelineRolloutReconciler()` starts up the workers to do the reconciliations

internal/controller/pipelinerollout_controller.go
juliev0: Yes, that's probably good if I pass in `context`. We can always get it from `context` if we're not the highest function on the stack that we can write to.

internal/controller/suite_test.go
juliev0: Plus, it allows testing of this function as well.

juliev0:
@afugazzotto I addressed your comments. Please re-review when you have time (tomorrow is fine). Thanks!

afugazzotto (approved):

PR #120, opened 2024-07-16
Explicitly define `Pipeline` and `ISBSvc` in the Rollout CRDs

### LINKED ISSUE #119: Explicitly define Pipeline and ISBSvc in `PipelineRollout` and `ISBServiceRollout`  CRD
# Summary

Currently, PipelineRollout CR embed Pipeline Spec directly as `runtime.RawExtension` with the naming of [Pipeline](https://github.com/numaproj/numaplane/blob/main/pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go#L31). It can confuse user this is only includes the Pipeline Spec but not the whole Pipeline definition. Similarly for ISBService. We should update the CRD to make it straightforward. 

---

xdevxy:
Fixes #119 

### Modifications

This PR explicitly define `Pipeline` and `ISBSvc` in the Rollout CRDs to make it clear what are the fields are used in Rollout CRDs. Currently only add `Spec` field.

### Verification
`make test`


pkg/apis/numaplane/v1alpha1/isbservicerollout_types.go
```diff
@@ -26,7 +26,12 @@ import (
 
 // ISBServiceRolloutSpec defines the desired state of ISBServiceRollout
 type ISBServiceRolloutSpec struct {
-	InterStepBufferService runtime.RawExtension `json:"interStepBufferService"`
+	InterStepBufferService InterStepBufferService `json:"InterStepBufferService"`
```
juliev0, line 29:
```suggestion
	InterStepBufferService InterStepBufferService `json:"interStepBufferService"`
```

juliev0:
I was going to ask if you verified that pausing still works, but I guess our unit test effectively does that since it verifies that the pipeline [got updated](https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout_controller_test.go#L178) (which would only happen if the pausing worked).

KeranYang (commented):
Thanks, Hao for making the quick change. Can we also update the [samples](https://github.com/numaproj/numaplane/tree/main/config/samples) to reflect the change?

pkg/apis/numaplane/v1alpha1/isbservicerollout_types.go
juliev0: shouldn't this be lower case?

KeranYang: Yes, it should match the numaflow version of it.

KeranYang: https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/pipeline_types.go#L453

KeranYang: taking back my response, this one doesn't need to match. The service name one needs to.

afugazzotto (approved):
LGTM

juliev0 (approved):

KeranYang (approved):

PR #125, opened 2024-07-19
Avoid data loss when pipeline, ISBService, or Numaflow Controller is updated

### LINKED ISSUE #9: Pause all Pipelines on the namespace prior to updating a Controller/ISBService
Document [here](https://docs.google.com/document/d/1xHzzaH8DgE9rxo_xCnbeWDmW49THgXQ8yDiOReb5edU/edit). Note that we may need to check the [status](https://github.com/numaproj/numaflow/issues/1773) of Numaflow-Controller deployment to see if latest change has being reconciled before resuming the pipelines.

### LINKED ISSUE #137: Pause feature: Rollout Status fields
# Summary

Make sure we provide all of the Status information included in Slide 14 [here](https://docs.google.com/presentation/d/1XtP-k5ny1RFEWUFgkROFgp4WcPdc_9CWgTDGL0A2NE0/edit#slide=id.g2784171e776_1_0) (both Rollout Status fields and ArgoCD)

# Use Cases

When would you use this?

---

juliev0:
Fixes #9 
Fixes #137 

### Modifications

The change is as described in [these slides](https://docs.google.com/presentation/d/1XtP-k5ny1RFEWUFgkROFgp4WcPdc_9CWgTDGL0A2NE0/edit#slide=id.p).

Basically, pausing of pipelines occurs in these cases:

- Pipeline needs to update
- ISBService needs to update
- Numaflow Controller needs to update

In order for the ISBService to be able to update, the following occurs:

1. ISBServiceRO reconciler indicates in memory its need for pausing and then enqueues all Pipelines using it to be processed by PipelineRO reconciliation queue
2. PipelineRO reconciliation worker sees ISBService's request for pausing and pauses pipeline
3. ISBServiceRO reconciler waits for all pipelines to be paused
4. Once they are paused, it proceeds to update
5. Once it has verified that the ISBService plus its underlying StatefulSet is completely updated, it sets its own need for pausing to true and enqueues the Pipelines
6. PipelineRO reconciliation worker sees that nobody has a need for pausing anymore and resumes pipeline

The Numaflow Controller RO reconciler follows a similar process.

### Verification

Only added a few unit tests, where the functionality was complicated. Wasn't terribly useful to add to our Gingko unit tests as explained [here](https://github.com/numaproj/numaplane/pull/125/files#r1686737460), as this feature introduces more interdependencies between the different rollouts and I'm thinking it may make more sense to move some of this unit test logic into the e2e test.

I did do a lot of manual testing, however. I tested both with the feature flag on and off.

**Testing with feature flag off:**
Performed basic CRUD (creation, update, deletion) and verified that the Status phase got set and Conditions.

**Testing with feature flag on:**
Procedure:

1. Create NumaflowControllerRollout, ISBServiceRollout, PipelineRollout
8. Edit each one and observe that the Pipeline gets paused prior to update occurring, and then resumes
9. Verified Status looks correct, including that `ChildResourcesDeployed.ObservedGeneration` matches the generation in which the change occurred

### Remaining Items

- add more testing (I think it mostly should occur in e2e but we can discuss the unit testing thing)
- do a manual test in which I can confirm there is no data loss when doing all of these updates
- address all of the TODOs I added (one of them is that I'd like to if possible have a single struct for "existing Object" and for "new Object" which includes everything in it - the GenericObject form, the Unstructured form, the Pipeline/ISBService spec form and pass that in everywhere, so that we can avoid doing too many conversions back and forth between these)


config/install.yaml
```diff
@@ -851,6 +851,14 @@ rules:
...
   verbs:
   - '*'
+- apiGroups:
+  - apps
+  resources:
+  - statefulsets
+  verbs:
+  - get
+  - list
+  - watch
```
juliev0, line 861: For now, we need to watch the StatefulSet of the ISBService to determine if it's done reconciling. If/when we incorporate this into the ISBService Status directly, then this can be removed.

config/samples/numaplane.numaproj.io_v1alpha1_isbservicerollout.yaml
```diff
@@ -8,6 +8,6 @@ spec:
     spec:
       # Example from https://github.com/numaproj/numaflow/blob/main/examples/0-isbsvc-jetstream.yaml
       jetstream:
-        version: latest
+        version: 2.9.8
```
juliev0, line 11: helps with testing to lock this down to 2.9.8

internal/controller/isbservicerollout_controller.go
```diff
@@ -100,8 +106,9 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	isbServiceRollout.Status.Init(isbServiceRollout.Generation)
 
-	err := r.reconcile(ctx, isbServiceRollout)
+	result, err := r.reconcile(ctx, isbServiceRollout)
```
juliev0, line 109: with this, we can return a `RequeueAfter`

internal/controller/isbservicerollout_controller.go
```diff
@@ -140,31 +147,38 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	// add Finalizer so we can ensure that we take appropriate action when CRD is deleted
 	if !controllerutil.ContainsFinalizer(isbServiceRollout, finalizerName) {
 		controllerutil.AddFinalizer(isbServiceRollout, finalizerName)
 	}
 
-	obj := kubernetes.GenericObject{
+	_, pauseRequestExists := GetPauseModule().getISBSvcPauseRequest(isbServiceRollout.Namespace, isbServiceRollout.Name)
+	if !pauseRequestExists {
+		// this is just creating an entry in the map if it doesn't already exist
+		GetPauseModule().newISBServicePauseRequest(isbServiceRollout.Namespace, isbServiceRollout.Name)
```
juliev0, line 178: map entry for this ISBService gets added when the ISBService is added, and gets deleted when ISBService is deleted. We always persist memory to indicate the pause state that ISBService requires; otherwise, the PipelineRollout reconciler has incomplete information.

internal/controller/isbservicerollout_controller.go
```diff
@@ -177,28 +191,186 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
...
+		// set the Status appropriately to "Pending" or "Deployed"
+		// if isbServiceNeedsUpdating - this means there's a mismatch between the desired ISBService spec and actual ISBService spec
+		// if there's a generation mismatch - this means we haven't even observed the current generation
+		// we may match the first case and not the second when we've observed the generation change but we're pausing pipelines
+		// we may match the second case and not the first if we need to update something other than ISBService spec
+		if isbServiceNeedsUpdating || isbServiceRollout.Status.ObservedGeneration < isbServiceRollout.Generation {
+			isbServiceRollout.Status.MarkPending()
+		} else {
+			isbServiceRollout.Status.MarkDeployed(isbServiceRollout.Generation)
+		}
```
juliev0, line 243: `Pending` now means we haven't updated the spec (or anything else) as we should: either we haven't even observed the latest generation (e.g. maybe we errored out while trying to process it) or we have observed it but simply are waiting for pipelines to pause

internal/controller/pipelinerollout_controller.go
```diff
@@ -55,6 +57,8 @@ const (
 	numWorkers                = 16 // can consider making configurable
 )
 
+var pipelineROReconciler *PipelineRolloutReconciler
```
juliev0, line 60: this enables the other rollouts to call `pipelineROReconciler.enqueue()`

internal/controller/pipelinerollout_controller_test.go
```diff
@@ -183,8 +189,19 @@ var _ = Describe("PipelineRollout Controller", Ordered, func() {
...
+				if err != nil {
+					return "", err
+				}
+				return string(updatedChildResource.Spec.Lifecycle.DesiredPhase), nil
+			}, timeout, interval).Should(Equal("Paused"))
+
 			By("Verifying that the PipelineRollout Status Phase is Deployed and ObservedGeneration matches Generation")
-			verifyStatusPhase(ctx, apiv1.PipelineRolloutGroupVersionKind, namespace, pipelineRolloutName, apiv1.PhaseDeployed)
+			verifyStatusPhase(ctx, apiv1.PipelineRolloutGroupVersionKind, namespace, pipelineRolloutName, apiv1.PhaseDeployed)*/
+
```
juliev0, line 204: The unit test is currently testing with the `DataLossPrevention` feature flag turned off. 

Once we try to test with this feature on, I am starting to think much of our unit testing should actually become e2e tests instead. The idea of each reconciler being a unit unto itself without cross dependencies with the other reconcilers is becoming less the case. If I were to try to verify that the Pipeline gets fully deployed from being paused to becoming running and applying the new spec, I really need to have a Numaflow Controller running as well.

internal/util/kubernetes/dynamic_client.go
```diff
@@ -49,7 +46,7 @@ func ParseStatus(obj *GenericObject) (GenericStatus, error) {
 	return status, nil
 }
 
-func GetResource(
+func GetUnstructuredCR(
```
juliev0, line 49: Each method (`Get`, `List`, `Create`, and `Update`) now has a version which uses `Unstructured` type and a version which uses `GenericObject` type

internal/util/kubernetes/unstructured_util.go
```diff
@@ -1,13 +1,309 @@
...
+
+	"github.com/numaproj/numaplane/internal/util/logger"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 )
 
 // this file contains utility functions for working with Unstructured types
```
juliev0, line 20: Not sure what you think. I had trouble figuring out what should go in `dynamic_client.go` vs `unstructured_util.go` so I just put everything in here and removed `dynamic_client.go`

internal/util/kubernetes/unstructured_util.go
```diff
@@ -1,13 +1,309 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Rationale for commenting this out:
+	// If we get a conflict, it means that the Status we were starting from in this reconciliation was old.
+	// If the new Status is derived from the old Status, then that would've been an invalid state.
+	// todo: since the primary motivation for this retry logic was to avoid errors showing up in the log, we may be
+	// able to instead conditionally log a warning vs an error for that
+	//err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
juliev0, line 270: @afugazzotto I wanted to bring it to your attention that I commented this out. Any time our `Reconcile()` function returns an error, there will be an automatic retry. 

I think in general with "retry on conflict", the solution is to Get the current object and then do all of the logic that you need to do to correctly update that object from its current state. My concern is this type of scenario:
1. In the previous reconciliation, one of our `Status.Conditions` was updated.
2. Our informer for the `Status` is slow, so for our next reconciliation we have a somewhat outdated version of the `Status` which doesn't include that previous update.
3. Our `object.Status` below doesn't include that update, but then it gets set in the object

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -108,7 +108,10 @@ func (s *Status) GetCondition(t ConditionType) *metav1.Condition {
 // Init sets certain Status parameters to a default initial state
 func (status *Status) Init(generation int64) {
 	status.SetObservedGeneration(generation)
-	status.MarkPending()
+	// rationale for commenting this out:
+	// "Pending" is now something we indicate when a rollout has been updated and we are trying to deploy it,
+	// as opposed to meaning that we're "pending reconciliation"
+	//status.MarkPending()
```
juliev0, line 114: @afugazzotto wanted to bring this one to your attention as well. I needed to change that.

internal/util/kubernetes/unstructured_util.go
afugazzotto: I introduced the `RetryOnConflict` with [this PR](https://github.com/numaproj/numaplane/pull/62) to fix the error we were seeing when updating state: `Operation cannot be fulfilled on ...: the object has been modified; please apply your changes to the latest version and try again`. If with your queueing implementation the error is no longer present even without the `RetryOnConflict` logic, it can be removed.

juliev0: yeah, I recall it was because we were seeing that error in the log, and even though it would retry, it would be alarming to users to see an error. Let me look into addressing that "false alarm" error separately...

config/samples/numaplane.numaproj.io_v1alpha1_isbservicerollout.yaml
juliev0, line 11: Note to self: describe behavior seen with incompatibility of versions and consider upgrade strategy in those cases

juliev0: So, the behavior I was seeing was that the new StatefulSet would not come up on its own when going between 2.10 and 2.9. I confirmed from Derek that the backward incompatibility is expected to occur even when the pipelines are fully paused. He sent [this](https://docs.nats.io/release-notes/whats_new/whats_new_210#downgrade-warnings) link.

xdevxy: in this case, we should probably always do delete and create/replace for isb upgrade? I'm going to create issue for it. and I think this needs to be addressed ASAP.

xdevxy: https://github.com/numaproj/numaplane/issues/147

juliev0: I spoke again to Derek, and he said "users should just always use version 2.10+". So, I suppose maybe I shouldn't set this 2.9.8 after all in that case. It looks like if I use Numaflow 1.2.1, I have access to 2 different Jetstream versions in 2.10 that I could test with. Maybe I'll set it to 2.10.11?

internal/util/util.go
```diff
@@ -10,8 +15,52 @@ func StructToStruct(src any, dest any) error {
...
 	err = json.Unmarshal(jsonBytes, dest)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to convert json %s: err=%s", string(jsonBytes), err)
 	}
 
 	return nil
 }
+
+func CompareMapsIgnoringNulls(a map[string]interface{}, b map[string]interface{}) bool {
```
xdevxy, line 24: nit: can you add a comment here?

internal/util/kubernetes/unstructured_util.go
xdevxy: This is just a pure remove? anything new added?

internal/controller/suite_test.go
```diff
@@ -109,6 +110,11 @@ var _ = BeforeSuite(func() {
 		UseExistingCluster: &useExistingCluster,
 	}
 
+	numaLogger := logger.New().WithName("controller-manager")
```
xdevxy, line 113: why is this change?

juliev0: It didn't look like we were ever setting the log level in here previously, and if a test were to fail, it would be convenient to set the log level to 4 below and re-run.

Below, though, I see we are actually testing `configManager.LoadAllConfigs()` and could theoretically be actually creating a ConfigMap to read in down there with the log level in it. 

This file is really seeming more and more like not a unit test, so I imagine we will replace a lot of this stuff anyway. If you prefer for me to take this out, I can.

internal/util/kubernetes/unstructured_util.go
juliev0: I added List functions: `ListCR()` and `ListUnstructuredCR()`, and now there's an "unstructured" version of each of the functions: Get, List, Create, Update

juliev0: I also added `CompareSpecs()` and `DeepCopy()` functions

internal/controller/config/config.go
```diff
@@ -53,6 +53,8 @@ func (*ConfigManager) GetControllerDefinitionsMgr() *ControllerDefinitionsManage
 type GlobalConfig struct {
 	LogLevel          int    `json:"logLevel" mapstructure:"logLevel"`
 	IncludedResources string `json:"includedResources" mapstructure:"includedResources"`
+	// Feature flag - if enabled causes pipelines to be paused when numaflow controller or ISB Service get updated
```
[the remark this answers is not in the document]
juliev0: good call, yes, I'll change the comment

internal/common/common.go
```diff
@@ -1,15 +1,44 @@
...
+var (
+	NumaflowGroupVersion = fmt.Sprintf("%s/%s", NumaflowAPIGroup, NumaflowAPIVersion)
+
+	PipelineGVR = schema.GroupVersionResource{
+		Group:    NumaflowAPIGroup,
+		Version:  NumaflowAPIVersion,
+		Resource: "pipelines",
+	}
+
+	DataLossPrevention bool
```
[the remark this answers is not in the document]
juliev0: So, currently I'm only ever using the value of this once at the beginning. The users all just create a local variable equal to this value on start up. I originally had it so that it could theoretically get changed in the middle of running. But then I was thinking that that would mean having to potentially write code to address it being changed from true to false or false to true, which could create more complexities. Do you think maybe it's worth a thought exercise to figure out just how complicated or not complicated it would be?

One thing to note is that if we do modify it and don't modify the Numaplane Controller Deployment itself, we would have to restart the Pods on every cluster. This can be done by writing a fairly simple script, however, assuming that we have the right to write to resources in numaplane-system namespace. (We could pursue getting myself and others "ClusterAppAdmin" privilege to our namespace)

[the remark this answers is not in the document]
juliev0: Update: per [this](https://intuit-teams.slack.com/archives/C05K82P3N9J/p1722295803818349) conversation, it seems we should have this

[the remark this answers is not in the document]
xdevxy: I think we would want this to be configurable dynamically. We can track this as a separate issue and do this as a follow up though.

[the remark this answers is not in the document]
juliev0: okay, added Issue for it: https://github.com/numaproj/numaplane/issues/155

internal/common/common.go
```diff
@@ -1,15 +1,44 @@
...
+	LabelKeyNumaplaneInstance = "numaplane.numaproj.io/tracking-id"
+
 	// LabelKeyNumaplaneControllerConfig is the label key used to identify the configmap for the controller definitions
 	LabelKeyNumaplaneControllerConfig = "numaplane.numaproj.io/config"
 	// LabelValueNumaplaneControllerConfig is the label value used to identify the configmap for the controller definitions
 	LabelValueNumaplaneControllerConfig = "numaflow-controller-definitions"
+
+	LabelKeyISBServiceNameForStatefulSet = "numaflow.numaproj.io/isbsvc-name"
+
+	LabelKeyISBServiceNameForPipeline = "numaplane.numaproj.io/isbsvc-name"
```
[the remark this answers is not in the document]
juliev0: Added a comment to the second, and decided to remove the first and instead reference the Numaflow constant directly

tests/manifests/controller_definitions.yaml
```diff
@@ -744,3 +744,773 @@ controllerDefinitions:
           - configMap:
               name: numaflow-controller-config
             name: controller-config-volume
+
+- version: "0.01" # "numaio-rc-0.01"
```
juliev0, line 748: I wonder if we may want to name these by their image names plus tags? I didn't expect Derek to create an image named "numaflow-rc:v0.01" (different semantic versioning scheme than "numaflow" image)

internal/controller/config/config.go
```diff
@@ -53,6 +53,10 @@ func (*ConfigManager) GetControllerDefinitionsMgr() *ControllerDefinitionsManage
 type GlobalConfig struct {
 	LogLevel          int    `json:"logLevel" mapstructure:"logLevel"`
 	IncludedResources string `json:"includedResources" mapstructure:"includedResources"`
+	// Feature flag - if enabled causes pipeline(s) to be paused when pipeline, numaflow controller, or ISB Service gets updated
+	DataLossPrevention bool `json:"dataLossPrevention" mapstructure:"dataLossPrevention"`
+	// List of Numaflow Controller image names to look for
+	NumaflowControllerImageNames []string `json:"numaflowControllerImageNames" mapstructure:"numaflowControllerImageNames"`
```
xdevxy, line 59: We need to add unit test for loading these two new fields right?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -64,6 +65,11 @@ const (
 const (
 	ControllerNumaflowControllerRollout = "numaflow-controller-rollout-controller"
 	NumaflowControllerDeploymentName    = "numaflow-controller"
+	DefaultNumaflowControllerImageName  = "numaflow"
+)
+
+var (
+	delayedRequeue = ctrl.Result{RequeueAfter: 20 * time.Second}
```
xdevxy, line 72: why we only have a `delayedRequeue` specifically to numaflow controller but not isbsvc or pipeline?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -192,42 +215,186 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
+		// this is just creating an entry in the map if it doesn't already exist
+		GetPauseModule().newControllerPauseRequest(namespace)
+	}
+
+	deployment, deploymentExists, err := r.getNumaflowControllerDeployment(ctx, controllerRollout)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if deploymentExists {
```
xdevxy, line 248: Can they be a race condition, that `deploymentExists` returns false during checking, but it become true when `sync` happens and we have a version upgrade?

internal/controller/config/config.go
juliev0: do you mean config_test.go? Do you see any real chance of this not loading? In general, I think unit tests should be for complicated areas of code, right?

internal/controller/numaflowcontrollerrollout_controller.go
juliev0: sorry, I'm also using it in isbsvc - let me move this to common.go

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -141,6 +151,19 @@ func (r *NumaflowControllerRolloutReconciler) Reconcile(ctx context.Context, req
...
 	}
 
+	deployment, deploymentExists, err := r.getNumaflowControllerDeployment(ctx, numaflowControllerRollout)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if deploymentExists {
+		// update our Status with the Deployment's Status
+		err = r.processNumaflowControllerStatus(ctx, numaflowControllerRollout, deployment)
```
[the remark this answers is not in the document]
juliev0: I see. Are you thinking of the case in which somebody manually deleted the Deployment? When I was writing this, the case I had in mind was the more typical one in which the NumaflowControllerRollout was just created for the first time.

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -192,42 +215,186 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
 	controllerRollout *apiv1.NumaflowControllerRollout,
 	namespace string,
-) error {
+) (ctrl.Result, error) {
 	numaLogger := logger.FromContext(ctx)
 
 	if !controllerRollout.DeletionTimestamp.IsZero() {
 		numaLogger.Info("Deleting NumaflowControllerRollout")
 		if controllerutil.ContainsFinalizer(controllerRollout, finalizerName) {
+			GetPauseModule().deleteControllerPauseRequest(namespace)
```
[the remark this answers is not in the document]
juliev0: It is just taking up a little extra memory. But what do you mean by "forgot"?

internal/controller/numaflowcontrollerrollout_controller.go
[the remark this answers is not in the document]
juliev0: sorry, deleting my comment - I didn't see where this code block was before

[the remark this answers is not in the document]
juliev0: I guess you're right - if we just created the Deployment for the first time, we can still process its health, which is probably in some sort of "Progressing" state

internal/controller/numaflowcontrollerrollout_controller.go
juliev0: oh, are you referring to the case in which somebody manually deleted the Numaflow Controller Deployment?

internal/controller/config/config.go
juliev0: (I did start working in a separate branch on adding unit tests for the reconcile() functions, though)

juliev0:
e2e is failing due to [this](https://github.com/numaproj/numaflow/issues/1881)

internal/controller/pipelinerollout_controller.go
```diff
@@ -472,35 +610,66 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-}
+	if found {
+		unstrucNew := unstruc.DeepCopy()
+		specMapAsInterface, found := unstrucNew.Object["spec"]
+		if found {
+			specMap, ok := specMapAsInterface.(map[string]interface{})
+			if ok {
+				lifecycleMapAsInterface, found := specMap["lifecycle"]
+				if found {
+					if ok {
```
afugazzotto, line 653: This can be removed.

internal/controller/isbservicerollout_controller.go
```diff
@@ -177,28 +190,192 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
...
+	if err = r.applyPodDisruptionBudget(ctx, isbServiceRollout); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to apply PodDisruptionBudget for ISBServiceRollout %s, err: %v", isbServiceRollout.Name, err)
 	}
 
-	// after the Apply, Get the ISBService so that we can propagate its health into our Status
-	isbsvc, err := kubernetes.GetCR(ctx, r.restConfig, &obj, "interstepbufferservices")
+	return ctrl.Result{}, nil
+}
+
+// take the existing pipeline and merge anything needed from the new pipeline definition
```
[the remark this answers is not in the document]
juliev0: thanks, will fix

internal/controller/isbservicerollout_controller.go
```diff
@@ -249,10 +488,17 @@ func processISBServiceStatus(ctx context.Context, isbsvc *kubernetes.GenericObje
...
+		if err != nil {
+			numaLogger.Errorf(err, "failed while determining if ISBService is fully reconciled: %+v, %v", isbsvc, err)
+			return
+		}
+
+		if reconciled {
 			rollout.Status.MarkChildResourcesHealthy(rollout.Generation)
 		} else {
-			rollout.Status.MarkChildResourcesUnhealthy("Progressing", "Mismatch between ISBService Generation and ObservedGeneration", rollout.Generation)
+			rollout.Status.MarkChildResourcesUnhealthy("Progressing", nonreconciledReason, rollout.Generation)
```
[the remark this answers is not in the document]
juliev0: got it

internal/controller/numaflowcontrollerrollout_controller.go
juliev0: We're waiting for the Pipelines to get paused - we explicitly want to delay before we check again

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -192,42 +215,186 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
+				numaLogger.Info("Numaflow Controller either needs to or is in the process of updating")
+
+				// request pause if we haven't already
+				pauseRequestUpdated, err := r.requestPipelinesPause(ctx, controllerRollout, true)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				if !pauseRequestUpdated { // don't do this yet if we just made a request - it's too soon for anything to have happened
+					// check if the pipelines are all paused and we're trying to update the spec
+					if controllerDeploymentNeedsUpdating {
```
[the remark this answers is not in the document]
juliev0: I think one nice thing about having these separate is that I can have a separate comment to explain why I have the condition `!pauseRequestUpdated`. Is there a reason you prefer to combine them?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -192,42 +215,186 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
+			} else {
+				// remove any pause requirement if necessary
+				_, err := r.requestPipelinesPause(ctx, controllerRollout, false)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+			}
+		}
+
+	}
```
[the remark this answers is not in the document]
juliev0: Okay, let me look into that.

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -192,42 +215,186 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
+		status, err := kubernetes.ParseStatus(pipeline)
+		if err != nil {
+			return false, err
+		}
+		if status.Phase != "Paused" {
+			return false, nil
+		}
+	}
+	return true, nil
+}
```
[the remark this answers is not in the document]
juliev0: Yes, maybe if both Reconcilers implement a common interface which includes a `getPipelines()` function, then this could a single function that uses it.

internal/controller/numaflowcontrollerrollout_controller.go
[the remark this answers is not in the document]
afugazzotto: I prefer them to be combined to avoid to much nesting and indentation (for readability). You could always specify in the comment that the comment only applies to one of the 2 variables.

internal/controller/pipelinerollout_controller.go
```diff
@@ -55,6 +57,18 @@ const (
...
+const (
+	PipelinePhaseUnknown   = ""
+	PipelinePhaseRunning   = "Running"
+	PipelinePhaseSucceeded = "Succeeded"
+	PipelinePhaseFailed    = "Failed"
+	PipelinePhasePausing   = "Pausing"
+	PipelinePhasePaused    = "Paused"
+	PipelinePhaseDeleting  = "Deleting"
+)
+
```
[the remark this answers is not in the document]
juliev0: yeah, good question. I guess we need to decide where in our code we still want to reference numaflow directly and where we don't. Theoretically, if we expect something to not change in Numaflow, then we may be okay to reference it directly, right? And eliminate copy/pasting into our code.

internal/controller/pipelinerollout_controller.go
juliev0: good find

internal/controller/pipelinerollout_controller.go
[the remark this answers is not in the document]
afugazzotto: I'm also not sure. We should have a separate PR to address that by looking at the entire code base and see where the dependency is used and if it could either be removed or acceptable to stay.

internal/controller/suite_test.go
xdevxy: got it, I'm good with keep it here, but why not set to 4 initially to show debug log?

internal/controller/config/config.go
xdevxy: this is just to make sure the parsing/loading is correct, e.g adding these fields for verification in https://github.com/numaproj/numaplane/blob/340cb90285797be8e49a056298ed8e5c664eaad3/internal/controller/config/config_test.go#L27.

internal/controller/numaflowcontrollerrollout_controller.go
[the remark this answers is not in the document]
xdevxy: like in case we can refactor later and this line of code get refactored out :) Looks like not a fatal issue even in that case.

internal/controller/numaflowcontrollerrollout_controller.go
xdevxy: that could be one case.

internal/controller/numaflowcontrollerrollout_controller.go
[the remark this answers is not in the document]
juliev0: okay, I'm fine to modify it

internal/controller/suite_test.go
juliev0: I'm good with that

xdevxy (approved):

juliev0:
@xdevxy @afugazzotto I have created a separate Issue to address the 4 comments that I haven't addressed yet in this one: https://github.com/numaproj/numaplane/issues/163

internal/controller/numaflowcontrollerrollout_controller.go
juliev0: @xdevxy  Thinking about this one more. I think `deploymentExists` can happen in 3 cases:
- the `NumaflowControllerRollout` has just been created: we won't try to pause and that seems okay (theoretically, there could be Pipelines running and `NumaflowControllerRollout` has been removed and recreated, but I don't think we can guarantee anything in that case, unless we tried to figure out which version of numa container was running on all of the Pipelines)
- the `NumaflowControllerRollout` was created in the previous reconciliation and the Informer cache is old: if by some rare chance, the `NumaflowControllerRollout` also just updated its version at the very same time, we won't try to pause and that seems okay as the NumaflowController only just got rolled out anyway
- somebody manually deleted the Numaflow Controller `Deployment`: if by some rare chance, the  `NumaflowControllerRollout` also just updated its version at the very same time, we won't try to pause - this case wouldn't be great but it's so rare, it seems kind of low priority. If we did try to handle it, I would just imagine that we could store the previous version in the NumaflowControllerRollout's Status and compare to that. Want me to maybe just add an issue for this, or how do you want me to move forward?

xdevxy: yeah I think these are rare cases so let's consider it low priority and having issue to track this for now.

PR #128, opened 2024-07-19
Status Update Error Requeue and Bug Fix

afugazzotto:
### Modifications

- On status update error, requeue the reconcile key to retry.
- BugFix on kustomization file for following error:
```
> kubectl kustomize tests/manifests | sed 's@quay.io/numaproj/@quay.io/numaproj/@' | sed 's/numaplane-controller:latest/numaplane-controller:latest/' | kubectl apply -f -

error: merging from generator &{0xc0013029c0 <nil>}: id resid.ResId{Gvk:resid.Gvk{Group:"", Version:"v1", Kind:"ConfigMap", isClusterScoped:false}, Name:"numaflow-controller-definitions-config", Namespace:"numaplane-system"} does not exist; cannot merge or replace
```

### DIFF
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -104,7 +104,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	if err != nil {
 		statusUpdateErr := r.updateISBServiceRolloutStatusToFailed(ctx, isbServiceRollout, err)
 		if statusUpdateErr != nil {
-			return ctrl.Result{}, statusUpdateErr
+			return ctrl.Result{RequeueAfter: statusUpdateErrorRequeueAfterDuration}, statusUpdateErr
 		}
 
 		return ctrl.Result{}, err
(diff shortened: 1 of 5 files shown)


xdevxy (approved):

internal/controller/isbservicerollout_controller.go
```diff
@@ -104,7 +104,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	if err != nil {
 		statusUpdateErr := r.updateISBServiceRolloutStatusToFailed(ctx, isbServiceRollout, err)
 		if statusUpdateErr != nil {
-			return ctrl.Result{}, statusUpdateErr
+			return ctrl.Result{RequeueAfter: statusUpdateErrorRequeueAfterDuration}, statusUpdateErr
```
juliev0, line 107: Do we need to include `RequeueAfter`? Whenever you return an error to `Reconcile()` it automatically requeues:

Screenshot 2024-07-16 at 9 37 01 PM

afugazzotto: I was thinking to provide extra time in case of status update error, that is why the delay. If you believe that is not necessary, I can remove it or I can change the delay length.

juliev0: Is there any reason why the default delay time wouldn't be sufficient? It automatically uses an exponential backoff algorithm to compute when that would be.

afugazzotto: I looked a bit more about the backoff and retry logic and I think that can be enough. I'll remove the `RequeueAfter` timeout. Thanks

juliev0: thanks so much!

PR #130, opened 2024-07-22
Remove RequeueAfter and fix Numaflow Controller config definitions bug

afugazzotto:
### Modifications

Removed RequeueAfter from controllers on status update error and fixed bug with deploying Numaflow Controller config definitions.

### DIFF
--- a/tests/manifests/kustomization.yaml
+++ b/tests/manifests/kustomization.yaml
@@ -25,3 +25,6 @@ configMapGenerator:
     - controller_definitions.yaml
   options:
     immutable: true
+    disableNameSuffixHash: true
+    labels:
+      numaplane.numaproj.io/config: numaflow-controller-definitions
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -104,7 +104,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	if err != nil {
 		statusUpdateErr := r.updateISBServiceRolloutStatusToFailed(ctx, isbServiceRollout, err)
 		if statusUpdateErr != nil {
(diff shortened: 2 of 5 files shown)


tests/manifests/kustomization.yaml
```diff
@@ -25,3 +25,6 @@ configMapGenerator:
     - controller_definitions.yaml
   options:
     immutable: true
+    disableNameSuffixHash: true
```
juliev0, line 28: I'm looking up what this is now. So, it basically means we'd just have a single ConfigMap in our cluster rather than one for each time the version of the ConfigMap changed, right?

juliev0 (approved):

juliev0:
@xdevxy Do we need that `codecov/patch` CI requirement?

tests/manifests/kustomization.yaml
afugazzotto: I understood that was the wanted behavior. @xdevxy do we want one configmap or multiple since the controller searches by label anyways?

juliev0: yeah, seems like the hash thing is really just related to the ConfigMap being volume mounted into the container, which we're not doing

afugazzotto: Since the controller searches by label, `disableNameSuffixHash: true` is not really necessary (I personally do not like hashes if not necessary). Also, we do not need `behavior: merge` at this time since we only have a single definitions file and since it was causing error. I'm thinking the error was due to also having set `immutable: true` but I'm not sure. I'm trying to see if separating the definitions in multiple files and enabling merge would make more sense.

xdevxy: > I understood that was the wanted behavior. @xdevxy do we want one configmap or multiple since the controller searches by label anyways?

I'm actually fine with either.

xdevxy (approved):

xdevxy:
> @xdevxy Do we need that `codecov/patch` CI requirement?

I think for this particular one we can override.

PR #135, opened 2024-07-24
Update Kustomize Transformer Config

xdevxy:
### Modifications

Modify the Kustomize Transformer config file based on the latest rollout spec changes.

### Verification

Manually tested by using the kustomize binary and `kubectl apply -k` command on the `config/kustomize/examples/transformer` directory

### DIFF
--- a/config/kustomize/examples/transformer/kustomization.yaml
+++ b/config/kustomize/examples/transformer/kustomization.yaml
@@ -2,7 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 configurations:
-- ./numaplane-transformer-config.yaml
+- https://raw.githubusercontent.com/numaproj/numaplane/main/config/kustomize/numaplane-transformer-config.yaml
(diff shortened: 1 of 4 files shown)


config/kustomize/examples/transformer/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 configurations:
-- https://raw.githubusercontent.com/numaproj/numaplane/main/config/kustomize/numaplane-transformer-config.yaml
```
afugazzotto, line 5: I chose to use the URL in case a user chooses to only download the examples directory but I'm ok if you want to change this.

afugazzotto (approved):

config/kustomize/examples/transformer/kustomization.yaml
xdevxy: Actually let me switch it back.

PR #144, opened 2024-07-26
Added Callback to the StateCache For Post Global Config Change

shubhamdixit863:
Yes

Fixes (https://github.com/numaproj/numaplane/issues/10)

### Modifications

While initializing the LiveState Cache in the `Init` method, a callback is registered that calls the `getCluster` method to update the cache. This ensures that changes in configuration, such as `includedResources` and `logLevel`, are dynamically applied, keeping the cache current.

### Verification

Todo

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -181,7 +181,6 @@ func main() {
 		kubectl,
 		customMetrics,
 	)
-
 	if err != nil {
 		numaLogger.Fatal(err, "Unable to create NumaflowControllerRollout controller")
 	}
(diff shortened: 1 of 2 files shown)


juliev0:
Hey @shubhamdixit863 - the code looks right, but did you do any testing? This issue is all about the testing - since this code involves the Numaflow Controller Rollout, can you check this procedure?:

- Create `NumaflowControllerRollout` and see that the new child resources got created
- Update "includedResources" in the file to add a new resource
- Modify child Deployment "numaflow-controller" in some way  and see if auto-healing works (might be okay if it doesn't but need to know)
- Modify `NumaflowControllerRollout` to switch from version 1.2.1 to 1.1.7 (if you're merged with latest master you should be able to use 1.1.7 now due to [this](https://github.com/numaproj/numaplane/blob/main/tests/manifests/controller_definitions.yaml#L386))

- Update "includedResources" in the configmap to remove the new resource you added 
- Modify child Deployment "numaflow-controller" in some way and see if auto-healing works (might be okay if it doesn't but need to know)
- Modify `NumaflowControllerRollout` to switch back to version 1.2.1

PR #148, opened 2024-07-26
Fix generation of pkg/client auto-generation

juliev0:
This auto-generates the pkg/client library, which among other things means we can use a Kubernetes rest config to generate a kubernetes client for Numaplane CRDs.

This can be used for our e2e tests and unit tests. 

### Modifications

Copied over the update-codegen.sh shell script from numaflow. Also, needed to add the "clientgen" directive to our CR structs.

Also, needed to change the order in the Makefile such that the controller-gen call is made after the call to `update-codegen.sh`; otherwise the DeepCopy file would get wiped.


Makefile
```diff
@@ -90,15 +90,14 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
...
-
 .PHONY: codegen
-codegen: generate manifests
+codegen: manifests controller-gen
+## Generate pkg/client directory
 	./hack/update-codegen.sh
 	rm -rf ./vendor
 	go mod tidy
+## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
juliev0, line 100: needed to move this after the call to `update-codegen.sh` or otherwise it would wipe one of the autogenerated files

Makefile
```diff
@@ -121,7 +120,7 @@ golangci-lint:
 	}
 
 .PHONY: lint
-lint: generate golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter & yamllint
```
juliev0, line 123: not sure we need auto-generated code to be linted, do we?

pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go
```diff
@@ -41,11 +41,11 @@ type PipelineRolloutStatus struct {
...
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
-//+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
-//+kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase",description="The current phase"
-
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase",description="The current phase"
```
juliev0, line 48: just added some extra spaces...probably doesn't have any effect, though

Mainly just needed to add "genclient" at the top

afugazzotto (approved):

juliev0:
I believe this falls into the "trivial-enough" category to merge after one approval without waiting for another, so merging

PR #149, opened 2024-07-29
fix: remove numarollout from rollout crd

darshansimha:
We need to do this to make sure the Numarollout extension only shows up in pipeline, isb and controller rollout crd's and not in rollout crd.

https://intuit-teams.slack.com/archives/C05K82P3N9J/p1722274324823689


xdevxy (commented):
Thanks @darshansimha, but why do we need to do this? do you mind adding a description in the summary?

darshansimha:
> Thanks @darshansimha, but why do we need to do this? do you mind adding a description in the summary?

We need to do this to make sure the Numarollout extension only shows up in pipeline, isb and controller rollout crd's and not in rollout crd.

mshakira (approved):

whynowy (approved):

PR #166, opened 2024-08-05
e2e test includes testing for both dataLossPrevention true and false

juliev0:
### Modifications

This PR introduces the testing of the e2e test in the CI for both `dataLossPrevention==true` and `dataLossPrevention==false`. (The e2e test is doing both if you look at the steps.)

(Perhaps later once we remove the feature flag and instead have the ability selectively turn this feature on or off for a given namespace, we can test a different way, but this is useful for now.)

This PR also fixes a few errors in the test that enable it to now succeed. 

### Verification

e2e test now runs for both cases as part of CI. I actually visually verified by also having a couple of other terminal windows up at the same time as I was running the test, one doing `watch -n 1 kubectl get pods` and the other doing `watch -n 1 kubectl get pipeline` to make sure that the pausing was happening when it was supposed to.


config/samples/numaplane.numaproj.io_v1alpha1_isbservicerollout.yaml
```diff
@@ -9,3 +9,5 @@ spec:
       # Example from https://github.com/numaproj/numaflow/blob/main/examples/0-isbsvc-jetstream.yaml
       jetstream:
         version: 2.10.3
+        persistence:
+          volumeSize: 1Gi
```
juliev0, line 13: I accidentally removed this from here when my real intention was to remove it from the test manifests - restoring

Makefile
```diff
@@ -179,8 +181,8 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 
 .PHONY: start
 start: image
-	$(KUBECTL) apply -f tests/manifests/numaplane-ns.yaml
-	$(KUBECTL) kustomize tests/manifests | sed 's@quay.io/numaproj/@$(IMAGE_NAMESPACE)/@' | sed 's/$(IMG):$(BASE_VERSION)/$(IMG):$(VERSION)/' | $(KUBECTL) apply -f -
+	$(KUBECTL) apply -f tests/manifests/default/numaplane-ns.yaml
+	$(KUBECTL) kustomize $(TEST_MANIFEST_DIR) | sed 's@quay.io/numaproj/@$(IMAGE_NAMESPACE)/@' | sed 's/$(IMG):$(BASE_VERSION)/$(IMG):$(VERSION)/' | $(KUBECTL) apply -f -
```
juliev0, line 185: be able to test either our standard manifest or an alternative one

juliev0: (useful in this case for testing the feature flag)

internal/controller/pipelinerollout_controller.go
```diff
@@ -361,14 +360,12 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 		return fmt.Errorf("failed to convert existing Pipeline spec %q into PipelineSpec type, err=%v", string(existingPipelineDef.Spec.Raw), err)
 	}
 
-	// Does pipeline spec need to be updated? is it already being updated?
-	pipelineNeedsToUpdate, pipelineIsUpdating, err := isPipelineUpdating(ctx, newPipelineDef, existingPipelineDef)
+	// Does pipeline spec need to be updated?
+	pipelineNeedsToUpdate, err := pipelineNeedsUpdating(ctx, newPipelineDef, existingPipelineDef)
```
juliev0, line 364: a little refactoring - the previous function was kind of doing 2 things, now it just does one

internal/controller/pipelinerollout_controller.go
```diff
@@ -395,19 +392,34 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
...
+
+	existingPipelinePauseDesired := existingPipelineSpec.Lifecycle.DesiredPhase == string(numaflowv1.PipelinePhasePaused)
+
+	numaLogger.Debugf("pipelineNeedsToUpdate: %t, pipelineUpdating: %t, existingPipelinePauseDesired=%t", pipelineNeedsToUpdate, pipelineUpdating, existingPipelinePauseDesired)
+
 	// If there is a need to update, does it require a pause?
+	// Here we look for whether we are trying to update Pipeline, or whether we are already processing an update which required us to pause
 	var pipelineUpdateRequiresPause bool
-	if pipelineNeedsToUpdate || pipelineIsUpdating {
+	if pipelineNeedsToUpdate || (pipelineUpdating && existingPipelinePauseDesired) {
```
juliev0, line 422: this is all part of the same refactoring.

tests/e2e/suite_test.go
```diff
@@ -48,7 +48,7 @@ var (
 	ctx           context.Context
 	cancel        context.CancelFunc
 	suiteTimeout  = 5 * time.Minute
-	testTimeout   = 1 * time.Minute
+	testTimeout   = 2 * time.Minute
```
juliev0, line 51: may need a little longer since we now need to wait for longer things like ISBService to be fully created, etc

.github/workflows/ci.yaml
```diff
@@ -152,6 +152,17 @@ jobs:
...
+          GOPATH: /home/runner/go
+        run: KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true make test-e2e
+
+      - name: Install Numaplane with dataLossPrevention=true
+        env:
+          GOPATH: /home/runner/go
+        run:
+          KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true TEST_MANIFEST_DIR=tests/manifests/special-cases/pause make start
+
+      - name: Run tests with dataLossPrevention=true
```
juliev0, line 165: first run the test according to `make start` which has the feature flag off, then with feature flag on

tests/e2e/functional_test.go
```diff
@@ -218,33 +215,29 @@ var _ = Describe("PipelineRollout e2e", func() {
 
 		// new NumaflowController spec
 		updatedNumaflowControllerSpec := apiv1.NumaflowControllerRolloutSpec{
-			Controller: apiv1.Controller{Version: "1.1.7"},
+			Controller: apiv1.Controller{Version: "0.0.6"},
```
juliev0, line 218: this is one of the latest RC manifests for numaflow, to bring in all of the changes

.github/workflows/ci.yaml
```diff
@@ -152,6 +152,17 @@ jobs:
           KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true make start
 
       - name: Run tests
```
xdevxy, line 154: nit: run e2e tests

.github/workflows/ci.yaml
xdevxy, line 165: nit: Run e2e tests with dataLossPrevention=true

xdevxy (approved):

internal/controller/pipelinerollout_controller.go
```diff
@@ -395,19 +392,34 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
...
+		return fmt.Errorf("failed to convert existing Pipeline spec %q into PipelineSpec type, err=%v", string(existingPipelineDef.Spec.Raw), err)
 	}
 
+	// is the Pipeline currently being reconciled?
+	pipelineUpdating, err := pipelineIsUpdating(newPipelineDef, existingPipelineDef)
+	if err != nil {
+		return err
+	}
+
+	existingPipelinePauseDesired := existingPipelineSpec.Lifecycle.DesiredPhase == string(numaflowv1.PipelinePhasePaused)
```
juliev0, line 415: I needed to add this line which was not in here before, so that we only consider cases of pipeline being reconciled while it's being paused. 

Background: the reason we are checking to see if the pipeline is being reconciled still is to make sure that we allow the pipeline update to go completely all the way through before we run it again. We don't want to resume it until this has completed.

But this makes it so we only concern ourselves of this while we have actually paused the pipeline.

Otherwise, we incorrectly will pause the pipeline simply because we are in the middle of reconciling the change to set `desiredPhase=Running`

PR #172, opened 2024-08-06
feat: adapt rollouts to use child resource's new conditions for status

### LINKED ISSUE #109: Adapt PipelineRollout and ISBServiceRollout status to account for Pipeline and ISBService ChildResourcesHealthy Conditions
# Summary

Once [the work](https://github.com/numaproj/numaflow/issues/1774) described [here](https://docs.google.com/document/d/1VCHA-tHjUQsC_9G-gNLRnR4_uQ0q7p07ccGdbbfnLPA/edit) is done, we need to update our `PipelineRollout` and `ISBServiceRollout` status to reflect the conditions.

Theoretically, we should be able to keep the same Status fields for the Rollouts themselves, but adapt the prior code which just looked for `Pipeline.phase==failed` or ISBSvc.phase==failed` to also look for the new Conditions.

# Use Cases

When would you use this?

---

dpadhiar:
Fixes #109 

### Modifications

There were new conditions for Numaflow resources Pipeline and ISBService introduced in this [PR](https://github.com/numaproj/numaflow/pull/1860) which include:
- DaemonServiceHealthy
- SideInputManagerHealthy
- VerticesHealthy
- ChildrenResourcesHealthy

We incorporate checking these new conditions into our current logic for determining their respective Rollout's status. Currently this logic only checks the child resource's Phase which we can be insufficient.

### Verification

Running `make test` locally and creating PipelineRollout and ISBServiceRollout that created child resources with the new conditions.


internal/controller/isbservicerollout_controller.go
```diff
@@ -601,3 +603,12 @@ func (r *ISBServiceRolloutReconciler) updateISBServiceRolloutStatusToFailed(ctx
 
 	return statusUpdateErr
 }
+
+func getChildrenResourcesHealthyCondition(conditions []metav1.Condition) *metav1.Condition {
```
xdevxy, line 607: Why not simplify this to return true and false similar to pipeline?

dpadhiar: No technical reason, just had written this before implementing PipelineRollout approach. Will change to simplify.

juliev0:
If Pipeline or ISBService has ChildResourcesUnhealthy.Reason=="Progressing", than the Rollout's ChildResourcesUnhealthy.Reason also needs to be set to "Progressing"

xdevxy (approved):

juliev0 (approved):

PR #173, opened 2024-08-08
Add event recorder for ISBService, Pipeline and NumaflowController Rollout

### LINKED ISSUE #153: Emit K8s events for major events
# Summary

If we are logging major events, we should also emit K8s events so we can see those in `kubectl get events` command

# Use Cases

Debugging and monitoring

---

chandankumar4:
Fixes #153 

### Modifications

- Record event for ISBServiceRollout, PipelineRollout and NumaflowControllerRollout

### DIFF
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -61,7 +61,8 @@ type ISBServiceRolloutReconciler struct {
 	scheme        *runtime.Scheme
 	restConfig    *rest.Config
 	customMetrics *metrics.CustomMetrics
-	recorder      record.EventRecorder
+	// the recorder is used to record events
+	recorder record.EventRecorder
 }
 
 func NewISBServiceRolloutReconciler(
@@ -102,9 +103,7 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		if apierrors.IsNotFound(err) {
 			return ctrl.Result{}, nil
 		} else {
-			r.customMetrics.ISBServicesSyncFailed.WithLabelValues().Inc()
-			numaLogger.Error(err, "Unable to get ISBServiceRollout", "request", req)
-			r.recorder.Eventf(isbServiceRollout, corev1.EventTypeWarning, "GetISBServiceFailed", "Failed to get isb service rollout: %v", err.Error())
+			r.ErrorHandler(isbServiceRollout, err, "GetISBServiceFailed", "Failed to get isb service rollout")
 			return ctrl.Result{}, err
 		}
 	}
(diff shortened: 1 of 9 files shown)


internal/controller/isbservicerollout_controller.go
```diff
@@ -143,12 +153,14 @@ func (r *ISBServiceRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		statusUpdateErr := r.updateISBServiceRolloutStatus(ctx, isbServiceRollout)
 		if statusUpdateErr != nil {
 			r.customMetrics.ISBServicesSyncFailed.WithLabelValues().Inc()
+			r.recorder.Eventf(isbServiceRollout, corev1.EventTypeWarning, "StatusUpdateFailed", "Failed to update isb service rollout status: %v", err.Error())
```
juliev0, line 156: maybe the `ISBServicesSyncFailed` metric, the Event, and possibly the logging of the error should all go into one function?

chandankumar4: I don't think it's valid for cases, like [here](https://github.com/numaproj/numaplane/blob/4ea4cbf46347fd47154151d34613f1a7f2abcfe8/internal/controller/isbservicerollout_controller.go#L141) logger is not available

juliev0: How do you mean that `logger` isn't available there? It's always available. If it's a function which is sort of at the top of the stack in our code we call `logger.GetBaseLogger()` like we do in that function [here](https://github.com/numaproj/numaplane/blob/4ea4cbf46347fd47154151d34613f1a7f2abcfe8/internal/controller/isbservicerollout_controller.go#L95). Otherwise, we call `logger.FromContext(ctx)` to get the logger.

chandankumar4: I mean, Currently we are not logging the error here and if we have a separate fn which will log error as well then we might get duplicate log as it's returning the error from here so parent fn will also print that error. if that's okay then I'll make the change.

juliev0: Are we always returning an error from `Reconcile()` - I'm curious if doing so results in an error being logged by the controller-runtime framework. It seems to me like it does. In that case, we could remove where we are explicitly logging because the framework does it anyway. Then this function could just write the Event and write the metric.

internal/controller/isbservicerollout_controller.go
```diff
@@ -201,13 +205,12 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
...
 	if err != nil {
-		// create object as it doesn't exist
+		// create an object as it doesn't exist
 		if apierrors.IsNotFound(err) {
 			numaLogger.Debugf("ISBService %s/%s doesn't exist so creating", isbServiceRollout.Namespace, isbServiceRollout.Name)
 			isbServiceRollout.Status.MarkPending()
 
-			err = kubernetes.CreateCR(ctx, r.restConfig, newISBServiceDef, "interstepbufferservices")
-			if err != nil {
+			if err := kubernetes.CreateCR(ctx, r.restConfig, newISBServiceDef, "interstepbufferservices"); err != nil {
```
juliev0, line 213: should we be calling `r.ErrorHandler()` here?

chandankumar4: No, as it's already called via parent

juliev0 (approved):

internal/controller/isbservicerollout_controller.go
chandankumar4: Make the changes accordingly

PR #185, opened 2024-08-08
chore: remove secrets from watched resources

xdevxy:
### Modifications

We should not need to cache secrets.

### Verification

make manifests

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -1049,7 +1049,7 @@ data:
   config.yaml: |-
     logLevel: 3
     includedResources: "group=apps,kind=Deployment;\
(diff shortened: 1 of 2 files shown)


afugazzotto (approved):

juliev0:
In that case, should we also be removing it from the [numaplane-role](https://github.com/numaproj/numaplane/blob/main/config/rbac/role.yaml#L17)?

xdevxy:
> In that case, should we also be removing it from the [numaplane-role](https://github.com/numaproj/numaplane/blob/main/config/rbac/role.yaml#L17)?

will do a separate PR for it, need to investigate why there's e2e test failure

PR #187, opened 2024-08-09
USDE Config

### LINKED ISSUE #175: Watch USDE configuration via a new ConfigMap to load paths
# Summary

Watch the USDE ConfigMap to read paths to exclude from the USDE comparison.

---

afugazzotto:
Fixes #175 

### Modifications

Added ConfigMap for USDE Milestone 1. The ConfigMap is watched using the config watcher and the `numaplane.numaproj.io/config` label.

### Verification

Added unit tests to ensure correct config loading.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -24,10 +24,14 @@ const (
 	// LabelKeyNumaplaneInstance Resource metadata labels (keys and values) used for tracking
 	LabelKeyNumaplaneInstance = "numaplane.numaproj.io/tracking-id"
 
-	// LabelKeyNumaplaneControllerConfig is the label key used to identify the configmap for the controller definitions
+	// LabelKeyNumaplaneControllerConfig is the label key used to identify additional Numaplane ConfigMaps (ex: Numaflow Controller definitions, USDE, etc.)
 	LabelKeyNumaplaneControllerConfig = "numaplane.numaproj.io/config"
-	// LabelValueNumaplaneControllerConfig is the label value used to identify the configmap for the controller definitions
-	LabelValueNumaplaneControllerConfig = "numaflow-controller-definitions"
+
+	// LabelValueNumaflowControllerDefinitions is the label value used to identify the Numaplane ConfigMap for the Numaflow Controller definitions
+	LabelValueNumaflowControllerDefinitions = "numaflow-controller-definitions"
+
+	// LabelValueNumaflowControllerDefinitions is the label value used to identify the Numaplane ConfigMap for the Numaflow Controller definitions
+	LabelValueUSDEConfig = "usde-config"
 
 	// LabelKeyISBServiceNameForPipeline is the label key used to identify the ISBService being used by a Pipeline
 	// This is useful as a Label to quickly locate all Pipelines of a given ISBService
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -30,7 +30,7 @@ const (
 	// LabelValueNumaflowControllerDefinitions is the label value used to identify the Numaplane ConfigMap for the Numaflow Controller definitions
 	LabelValueNumaflowControllerDefinitions = "numaflow-controller-definitions"
 
-	// LabelValueNumaflowControllerDefinitions is the label value used to identify the Numaplane ConfigMap for the Numaflow Controller definitions
+	// LabelValueUSDEConfig is the label value used to identify the USDE ConfigMap
 	LabelValueUSDEConfig = "usde-config"
 
 	// LabelKeyISBServiceNameForPipeline is the label key used to identify the ISBService being used by a Pipeline
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -205,9 +205,14 @@ func (cm *ConfigManager) UnsetUSDEConfig() {
 	cm.usdeConfig = &USDEConfig{}
 }
 
-func (cm *ConfigManager) GetUSDEConfig() *USDEConfig {
+func (cm *ConfigManager) GetUSDEConfig() (USDEConfig, error) {
 	cm.usdeConfigLock.Lock()
 	defer cm.usdeConfigLock.Unlock()
 
-	return cm.usdeConfig
+	usdeConfig, err := CloneWithSerialization(cm.usdeConfig)
+	if err != nil {
+		return USDEConfig{}, err
+	}
+
+	return *usdeConfig, nil
 }
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -20,7 +20,7 @@ type ConfigManager struct {
 	numaflowControllerDefMgr NumaflowControllerDefinitionsManager
 
 	// USDE Config
-	usdeConfig     *USDEConfig
+	usdeConfig     USDEConfig
 	usdeConfigLock *sync.RWMutex
 }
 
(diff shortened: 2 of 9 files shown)


internal/common/common.go
```diff
@@ -24,10 +24,14 @@ const (
...
-	// LabelKeyNumaplaneControllerConfig is the label key used to identify the configmap for the controller definitions
+	// LabelKeyNumaplaneControllerConfig is the label key used to identify additional Numaplane ConfigMaps (ex: Numaflow Controller definitions, USDE, etc.)
 	LabelKeyNumaplaneControllerConfig = "numaplane.numaproj.io/config"
-	// LabelValueNumaplaneControllerConfig is the label value used to identify the configmap for the controller definitions
-	LabelValueNumaplaneControllerConfig = "numaflow-controller-definitions"
+
+	// LabelValueNumaflowControllerDefinitions is the label value used to identify the Numaplane ConfigMap for the Numaflow Controller definitions
+	LabelValueNumaflowControllerDefinitions = "numaflow-controller-definitions"
+
+	// LabelValueNumaflowControllerDefinitions is the label value used to identify the Numaplane ConfigMap for the Numaflow Controller definitions
```
juliev0, line 33: accidental copy/paste

internal/util/kubernetes/config_watcher.go
```diff
@@ -34,44 +34,86 @@ func StartConfigMapWatcher(ctx context.Context, config *rest.Config) error {
...
+
+	// Add or update the controller definition config based on a version if the configmap has the correct label
+	for _, v := range configMap.Data {
+		var controllerConfig config.NumaflowControllerDefinitionConfig
+		if err := yaml.Unmarshal([]byte(v), &controllerConfig); err != nil {
+			numaLogger.Error(err, "failed to unmarshal Numaflow Controller Definitions config")
+			continue
+		}
+
+		// controller config definition is immutable, so no need to update the existing config
```
xdevxy, line 89: USDE config will not be immutable, right?

internal/common/common.go
afugazzotto: yep, fixed it, thanks

internal/util/kubernetes/config_watcher.go
afugazzotto: not at this time

internal/controller/config/config.go
```diff
@@ -179,3 +190,24 @@ func (cm *ConfigManager) RegisterCallback(f func(config GlobalConfig)) {
...
+	defer cm.usdeConfigLock.Unlock()
+
+	cm.usdeConfig = &USDEConfig{}
+}
+
+func (cm *ConfigManager) GetUSDEConfig() *USDEConfig {
+	cm.usdeConfigLock.Lock()
+	defer cm.usdeConfigLock.Unlock()
+
+	return cm.usdeConfig
```
juliev0, line 212: should we return a copy of this since otherwise the caller can refer to the reference while it's being modified, right?

xdevxy: is the intention is to see the latest modification anyway?

juliev0: I'm just referring to the fact that we have a lock, but is it effective if the caller has a reference? those values in the reference can change after it calls `GetUSDEConfig()`

internal/util/kubernetes/config_watcher.go
```diff
@@ -34,44 +34,86 @@ func StartConfigMapWatcher(ctx context.Context, config *rest.Config) error {
...
+		} else if event.Type == watch.Deleted {
+			config.GetConfigManagerInstance().GetControllerDefinitionsMgr().RemoveNumaflowControllerDefinitionConfig(controllerConfig)
 		}
 	}
 }
+
+func handleUSDEConfigMapEvent(configMap *corev1.ConfigMap, event watch.Event) error {
+	usdeConfig := config.USDEConfig{}
+
+	err := yaml.Unmarshal([]byte(configMap.Data["pipelineSpecExcludedPaths"]), &usdeConfig.PipelineSpecExcludedPaths)
```
juliev0, line 101: does Delete case work okay? Does `configMap==nil`?

afugazzotto: When deleting the ConfigMap, the configMap argument on line 98 is still the previous ConfigMap. In any case, I added some nil checks (just in case) and I also moved the logic inside the event if-statement because I think it makes more sense in this case.

juliev0 (approved):

internal/util/kubernetes/config_watcher.go
juliev0: perfect, thanks

internal/controller/config/config.go
xdevxy: @afugazzotto I don't understand why we return a copy here. if the copy is used how to detect the runtime changes that happened? unless this is only used for unit test. marking this as change request until the question get clarified.

xdevxy (changes requested):
https://github.com/numaproj/numaplane/pull/187#discussion_r1712346474

internal/controller/config/config.go
afugazzotto: @xdevxy if the config is being used and there is a change, then it would make more sense to return the reference instead of making a copy. However, I see a similar logic [here](https://github.com/numaproj/numaplane/blob/main/internal/controller/config/config.go#L66)? Would we also need to change that? How is the GlobalConfig handling changes when returning a copy?

juliev0: If you're returning a reference, then I don't see any point in having a lock.

The GlobalConfig was purposely written to return a copy - the idea is that the caller may want to get multiple values from the ConfigMap and ensure atomic correctness between them.

juliev0: The call to `GetConfig()` is used periodically in goroutines to get an atomically correct version of the Config.

In addition, the GlobalConfig has a callback capability - this is used to update any other copies of values that are stored. That's what we're doing for the LogLevel.

juliev0: Hey Hao. Is there any need to hold up this PR? When I spoke to you on Friday, it sounded like you just wanted to understand how Tony was going to use the code?

xdevxy: yes, sorry was out yesterday. I think it is fine to return a reference, however the read and write to the reference needs to be protected by lock (e.g. rwlock). If we return a copy then like what Julie said we need to use a callback or etc. I'm fine with either model, and I don't want to hold this PR for this. We can do as a follow up.

xdevxy (approved):

internal/controller/config/config.go
afugazzotto: Made changes after last conversation with @xdevxy

internal/controller/config/config.go
```diff
@@ -179,3 +190,24 @@ func (cm *ConfigManager) RegisterCallback(f func(config GlobalConfig)) {
 
 	cm.callbacks = append(cm.callbacks, f)
 }
+
+func (cm *ConfigManager) UpdateUSDEConfig(config USDEConfig) {
+	cm.usdeConfigLock.Lock()
+	defer cm.usdeConfigLock.Unlock()
+
+	cm.usdeConfig = config
```
juliev0, line 198: Does this perform a Deep copy or shallow copy?

internal/controller/config/config.go
```diff
@@ -179,3 +190,24 @@ func (cm *ConfigManager) RegisterCallback(f func(config GlobalConfig)) {
...
+	defer cm.usdeConfigLock.Unlock()
+
+	cm.usdeConfig = USDEConfig{}
+}
+
+func (cm *ConfigManager) GetUSDEConfig() USDEConfig {
+	cm.usdeConfigLock.Lock()
+	defer cm.usdeConfigLock.Unlock()
+
+	return cm.usdeConfig
```
juliev0, line 212: Does this perform a deep copy or shallow copy?

juliev0: discussed offline that the array will be copied wholesale, so I'm good

PR #190, opened 2024-08-13
fix: separate out logic for each Condition

### LINKED ISSUE #189: Pipeline can't be marked both "Failed" and "Unpaused" at the same time
**Describe the bug**
[This](https://github.com/numaproj/numaplane/blob/v0.3.0/internal/controller/pipelinerollout_controller.go#L434) line to set `Condition.PipelinePausedOrPausing=False` can't be called if Pipeline is in a Failed state.

Note that through testing, I realized that a Pipeline that's Failed can then have its desiredPhase set to Paused, in which case its phase becomes "Paused". Therefore, a Pipeline that actually has `phase=Failed` is not in a paused state.

**To Reproduce**

1. Set `dataLossPrevention=true`
2. Update PipelineRollout so spec has a validation error. 
3. Observe that Pipeline will be paused so it can be updated and then will go into a failed state
4. PipelineRollout Conditions show ChildResourcesHealthy=false, but PausingOrPaused is still set to true when it shouldn't be

---

juliev0:
Fixes #189 

### Modifications

Separate out the logic for deriving the "ChildResourcesHealthy" Condition value from the "PipelinePaused" Condition. This way we better ensure that the value for each of them is set one way or another.

### Verification

Tested with both `dataLossPrevention` true and false. Modified PipelineRollout a few times and each time did `kubectl get pipeline -o yaml` to check the Conditions' Status was correct as far as message, reason, and observedGeneration.

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -606,15 +606,6 @@ func (r *PipelineRolloutReconciler) checkForPauseRequest(ctx context.Context, pi
 	return controllerRequestsPause || isbsvcRequestsPause, true, nil
 }
 
-func setPipelineHealthStatus(pipeline *kubernetes.GenericObject, pipelineRollout *apiv1.PipelineRollout, pipelineObservedGeneration int64) {
-
-	if pipelineObservedGenerationCurrent(pipeline.Generation, pipelineObservedGeneration) {
-		pipelineRollout.Status.MarkChildResourcesHealthy(pipelineRollout.Generation)
-	} else {
-		pipelineRollout.Status.MarkChildResourcesUnhealthy("Progressing", fmt.Sprintf("Mismatch between Pipeline Generation %d and ObservedGeneration %d", pipeline.Generation, pipelineObservedGeneration), pipelineRollout.Generation)
-	}
-}
-
 func pipelineObservedGenerationCurrent(generation int64, observedGeneration int64) bool {
 	return generation <= observedGeneration
 }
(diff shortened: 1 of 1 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -646,28 +637,45 @@ func (r *PipelineRolloutReconciler) processPipelineStatus(ctx context.Context, p
...
+	} else if pipelineChildResourceStatus == "False" {
+		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed, Pipeline Child Resource(s) Unhealthy", pipelineRollout.Generation)
 	} else if pipelinePhase == numaflowv1.PipelinePhasePaused || pipelinePhase == numaflowv1.PipelinePhasePausing {
-		reason := fmt.Sprintf("Pipeline%s", string(pipelinePhase))
-		msg := fmt.Sprintf("Pipeline %s", strings.ToLower(string(pipelinePhase)))
-		pipelineRollout.Status.MarkPipelinePausingOrPaused(reason, msg, pipelineRollout.Generation)
-		setPipelineHealthStatus(existingPipelineDef, pipelineRollout, pipelineStatus.ObservedGeneration)
-	} else if pipelinePhase == numaflowv1.PipelinePhaseUnknown || pipelineChildResourceStatus == "Unknown" {
-		pipelineRollout.Status.MarkChildResourcesHealthUnknown("PipelineUnknown", "Pipeline Phase Unknown", pipelineRollout.Generation)
+		pipelineRollout.Status.MarkChildResourcesHealthUnknown("PipelineUnknown", "Pipeline Pausing - health unknown", pipelineRollout.Generation)
```
juliev0, line 658: I've noticed that in Numaflow, if a Pipeline was in a Failed state and is then set to `desiredPhase=Paused`, it takes the Pods down to 0 and then sets its "phase" to "Paused". Therefore, I thought we may want to just call this "Unknown health"? (I'm open)

internal/controller/pipelinerollout_controller.go
```diff
@@ -646,28 +637,45 @@ func (r *PipelineRolloutReconciler) processPipelineStatus(ctx context.Context, p
...
 	} else if pipelinePhase == numaflowv1.PipelinePhaseDeleting {
 		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineDeleting", "Pipeline Deleting", pipelineRollout.Generation)
+	} else if pipelinePhase == numaflowv1.PipelinePhaseUnknown || pipelineChildResourceStatus == "Unknown" {
+		pipelineRollout.Status.MarkChildResourcesHealthUnknown("PipelineUnknown", "Pipeline Phase Unknown", pipelineRollout.Generation)
 	} else {
-		setPipelineHealthStatus(existingPipelineDef, pipelineRollout, pipelineStatus.ObservedGeneration)
-		pipelineRollout.Status.MarkPipelineUnpaused(pipelineDef.Generation)
```
juliev0, line 667: this `pipelineDef.Generation` actually wasn't set to anything, so this was effectively not setting `observedGeneration` at all

internal/controller/pipelinerollout_controller.go
```diff
@@ -646,28 +637,45 @@ func (r *PipelineRolloutReconciler) processPipelineStatus(ctx context.Context, p
...
 
-	if pipelineChildResourceReason == "Progressing" {
+	if pipelineChildResourceReason == "Progressing" || !pipelineObservedGenerationCurrent(pipeline.Generation, pipelineStatus.ObservedGeneration) {
 		pipelineRollout.Status.MarkChildResourcesUnhealthy("Progressing", "Pipeline Progressing", pipelineRollout.Generation)
-	} else if pipelinePhase == numaflowv1.PipelinePhaseFailed || pipelineChildResourceStatus == "False" {
-		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed", pipelineRollout.Generation)
+	} else if pipelinePhase == numaflowv1.PipelinePhaseFailed {
+		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Phase=Failed", pipelineRollout.Generation)
+	} else if pipelineChildResourceStatus == "False" {
+		pipelineRollout.Status.MarkChildResourcesUnhealthy("PipelineFailed", "Pipeline Failed, Pipeline Child Resource(s) Unhealthy", pipelineRollout.Generation)
```
xdevxy, line 656: As far as I can tell this PR didn't introduce any changes for the reason of the condition, right? @dpadhiar can you make sure argocd changes is up to date with latest condition definition? Thanks!

dpadhiar: This shouldn't make any change for reason of condition as [the lua scripts also check that when the generations are mismatched](https://github.com/argoproj/argo-cd/blob/d2231577c7f667d86bd0aa9505f871ecf1fde2bb/resource_customizations/numaplane.numaproj.io/PipelineRollout/health.lua#L17), we set the Rollout as progressing. This should be good!

xdevxy (approved):

PR #208, opened 2024-08-20
feat: add numaflow controller definition for v1.3.0

dpadhiar:
### Modifications

Adding the controller definition for `v1.3.0` which is the most recent Numaflow version that includes monovertices.

### Verification

Ran `make start` to update local cm and `make test` to verify tests still work using this definition.

Created sample rollouts to see that the resources were created and functioned properly using this new definition.

### DIFF
--- a/tests/manifests/default/controller_definitions.yaml
+++ b/tests/manifests/default/controller_definitions.yaml
@@ -1,5 +1,5 @@
 controllerDefinitions:
-- version: "1.3.0-rc1"
+- version: "1.3.0"
   fullSpec: |
     apiVersion: v1
     kind: ServiceAccount
@@ -318,7 +318,7 @@ controllerDefinitions:
             - controller
             env:
             - name: NUMAFLOW_IMAGE
-              value: quay.io/numaproj/numaflow:1.3.0-rc1
+              value: quay.io/numaproj/numaflow:v1.3.0
             - name: NAMESPACE
               valueFrom:
                 fieldRef:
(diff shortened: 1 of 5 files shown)


tests/manifests/default/controller_definitions.yaml
```diff
@@ -248,141 +286,115 @@ controllerDefinitions:
...
+            app.kubernetes.io/component: controller-manager
+            app.kubernetes.io/name: controller-manager
+            app.kubernetes.io/part-of: numaflow
+        spec:
+          containers:
+          - args:
+            - controller
+            env:
+            - name: NUMAFLOW_IMAGE
+              value: quay.io/numaproj/numaflow:1.3.0-rc1
```
vigith, line 321:
```suggestion
              value: quay.io/numaproj/numaflow:1.3.0
```

we have released 1.3.0

juliev0: Do you want to replace that here @dpadhiar ?

juliev0:
I think the way to verify this is to create your sample NumaflowControllerRollout and then the other Rollouts, and make sure everything gets created successfully. (and no weird errors in Numaflow log)

tests/manifests/default/controller_definitions.yaml
dpadhiar: Updated now to use image `v1.3.0`

xdevxy (approved):

juliev0 (approved):

PR #209, opened 2024-08-20
feat: User Strategy in Namespace-level ConfigMap

### LINKED ISSUE #177: User Strategy Selection Field
# Summary

Add user strategy selection field to namespace-level ConfigMap.
Add various scenarios to the [design doc](https://docs.google.com/document/d/1BzmSMhqr_Qs4X42TufRwfB8f2RJXaypSXDVxh-VKy3s) (ex: consider different strategies selected for different rollout resources).

---

afugazzotto:
Fixes #177 

### Modifications

Added a namespace-level ConfigMap users can use to add user-related customization parameters. Modified the config watcher to look for ConfigMaps in all namespaces and handle only those of interest based on labels and the namespace they are in.

### Verification

Manually tested.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -33,7 +33,7 @@ const (
 	// LabelValueUSDEConfig is the label value used to identify the USDE ConfigMap
 	LabelValueUSDEConfig = "usde-config"
 
-	// LabelValueNamespaceConfig is the label value used to identify the namespace-level ConfigMap
+	// LabelValueNamespaceConfig is the label value used to identify the user's namespace-level ConfigMap
 	LabelValueNamespaceConfig = "namespace-level-config"
 
 	// LabelKeyISBServiceNameForPipeline is the label key used to identify the ISBService being used by a Pipeline
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -33,6 +33,9 @@ const (
 	// LabelValueUSDEConfig is the label value used to identify the USDE ConfigMap
 	LabelValueUSDEConfig = "usde-config"
 
+	// LabelValueNamespaceConfig is the label value used to identify the namespace-level ConfigMap
+	LabelValueNamespaceConfig = "namespace-level-config"
+
 	// LabelKeyISBServiceNameForPipeline is the label key used to identify the ISBService being used by a Pipeline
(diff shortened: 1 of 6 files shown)


internal/common/common.go
```diff
@@ -33,6 +33,9 @@ const (
 	// LabelValueUSDEConfig is the label value used to identify the USDE ConfigMap
 	LabelValueUSDEConfig = "usde-config"
 
+	// LabelValueNamespaceConfig is the label value used to identify the namespace-level ConfigMap
```
juliev0, line 36:
```suggestion
	// LabelValueNamespaceConfig is the label value used to identify the user's namespace-level ConfigMap
```

juliev0: feel free to re-word, but wanted to contrast from "numaplane-system" namespace for clarity

internal/controller/config/config.go
```diff
@@ -22,6 +22,10 @@ type ConfigManager struct {
 	// USDE Config
 	usdeConfig     USDEConfig
 	usdeConfigLock *sync.RWMutex
+
+	// Namespace-level Config
```
juliev0, line 26:
```suggestion
	// User Namespace-level Config
```

internal/util/kubernetes/config_watcher.go
```diff
@@ -38,7 +40,7 @@ func StartConfigMapWatcher(ctx context.Context, config *rest.Config) error {
 func watchConfigMaps(ctx context.Context, client kubernetes.Interface, namespace string) {
```
juliev0, line 40: nit: the contract of this function currently makes it seem like we're watching ConfigMaps specifically in the `namespace` that's passed in as a parameter, which we're not. We could either just rename the parameter as `numaplaneNamespace` or something like that, or don't pass in `namespace` at all (the caller is calling `os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")` which could theoretically happen here)

internal/usde/usde.go
```diff
@@ -0,0 +1,39 @@
+package usde
```
afugazzotto, line 1: This file will later also contain the USDE logic and more USDE related code.

juliev0 (approved):

xdevxy (approved):

PR #210, opened 2024-08-21
test: add e2e tests for monovertexrollouts

### LINKED ISSUE #196: E2E test for MonoVertex Rollout

dpadhiar:
Fixes #196 

### Modifications

Adds e2e test cases for MonoVertexRollouts which are creating, updating and deleting a MonoVertexRollout.

### Verification

Running `make test-e2e` locally and checking that CI is passing


tests/e2e/monovertex.go
```diff
@@ -0,0 +1,108 @@
...
+			return false
+		}
+		return f(retrievedMonoVertexSpec)
+	}).WithTimeout(testTimeout).Should(BeTrue())
+
+}
+
+func verifyMonoVertexReady(namespace, monoVertexName string) {
+
+	document("Verifying that the MonoVertex exists")
```
juliev0, line 54:
```suggestion
	document("Verifying that the MonoVertex is running")
```

tests/e2e/functional_test.go
```diff
@@ -216,6 +267,8 @@ var _ = Describe("Functional e2e", Serial, func() {
 			return rollout, nil
 		})
 
+		// TODO: update this controller image when Numaflow v1.3.1 is released
+		//       versions prior to v1.3.0 do not contain the MonoVertex CRD and could lead to issues in future tests
 		verifyNumaflowControllerDeployment(Namespace, func(d appsv1.Deployment) bool {
 			return d.Spec.Template.Spec.Containers[0].Image == "quay.io/numaio/numaflow-rc:v0.0.6"
```
juliev0, line 273: So, what's happening now? It looks like we are updating to v0.0.6, so how is the code below working? Was MonoVertex CRD in 0.0.6?

tests/e2e/monovertex.go
```diff
@@ -0,0 +1,108 @@
...
+		*rollout, err = f(*rollout)
+		if err != nil {
+			return err
+		}
+
+		_, err = monoVertexRolloutClient.Update(ctx, rollout, metav1.UpdateOptions{})
+		return err
+	})
+	Expect(err).ShouldNot(HaveOccurred())
+}
```
juliev0, line 108: Glad to see all of these functions in here!

tests/e2e/functional_test.go
dpadhiar: The tests following this rely on just the Numaplane Controller to update/delete the CRDs. This way, the Numaflow Controller version doesn't impact us. In the future it could if we do any tests checking the MonoVertex pods after updating the controller version.

juliev0: Ahh yes, since the CRDs aren't actually contained within the Numaflow Controller version, that part doesn't matter. We just don't have the MonoVertex being reconciled.

juliev0: In that case, we should update the comment which says "versions prior to v1.3.0 do not contain the MonoVertex CRD and could lead to issues in future" to "versions prior to v1.3.0 do not reconcile MonoVertex"

juliev0 (approved):

Issue #176, opened 2024-08-08
Implement USDE Comparison


afugazzotto:
# Summary

Use the USDE configuration paths to perform a comparison of the current spec vs the new spec excluding those paths. At this time, use option 1 from the [design doc](https://docs.google.com/document/d/1BzmSMhqr_Qs4X42TufRwfB8f2RJXaypSXDVxh-VKy3s). If there is extra time, try to implement option 2.
Also, figure out if `reflect.DeepEqual` is enough to use or if we need to evaluate alternatives such as: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/equality/semantic.go

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #215, opened 2024-08-28
USDE main logic

### LINKED ISSUE #178: Upgrade Strategy Decision Engine (USDE) Implementation (Milestone 1)
Supported strategies for Milestone 1 are manual upgrades (PPND feature flag off) or PPND (feature flag on). Where PPND = Pipeline Pause and Drain.

Milestone 1 Tasks:
- [x] #175 
- [x] #176 
- [x] #177 
- [x] #228 
- [x] #248

---

afugazzotto:
Fixes #178 #176 

### Modifications

Implemented USDE main logic that uses USDE config and Namespace-level related parameters to decide the upgrade strategy to use. Integrated this logic into the PipelineRollout reconciliation process.

### Verification

Unit tests and manual tests.

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -407,9 +407,9 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 		return fmt.Errorf("failed to convert existing Pipeline spec %q into PipelineSpec type, err=%v", string(existingPipelineDef.Spec.Raw), err)
 	}
 
-	// TTODO:
-	// usde.GetUpgradeStrategy(newPipelineDef, existingPipelineDef) => do-nothing, direct-apply, ppnd?
-	// based on function returned value, perform appropriate operations below and/or pass the value around
+	// TODO: based on function returned value, perform appropriate operations below and/or pass the value around
+	// strategy, err := usde.GetUpgradeStrategy(newPipelineDef, existingPipelineDef)
+	// (remember to apply this also to ISBSvc)
 	// INSTEAD OF:
 	// Does pipeline spec need to be updated?
 	pipelineNeedsToUpdate, err := pipelineNeedsUpdating(ctx, newPipelineDef, existingPipelineDef)
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -407,6 +407,10 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 		return fmt.Errorf("failed to convert existing Pipeline spec %q into PipelineSpec type, err=%v", string(existingPipelineDef.Spec.Raw), err)
 	}
 
+	// TTODO:
+	// usde.GetUpgradeStrategy(newPipelineDef, existingPipelineDef) => do-nothing, direct-apply, ppnd?
+	// based on function returned value, perform appropriate operations below and/or pass the value around
+	// INSTEAD OF:
 	// Does pipeline spec need to be updated?
 	pipelineNeedsToUpdate, err := pipelineNeedsUpdating(ctx, newPipelineDef, existingPipelineDef)
 	if err != nil {
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -45,6 +45,7 @@ import (
 
 	numaflowv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaplane/internal/common"
+	"github.com/numaproj/numaplane/internal/usde"
 
 	"github.com/numaproj/numaplane/internal/util"
 	"github.com/numaproj/numaplane/internal/util/kubernetes"
@@ -398,6 +399,7 @@ func mergePipeline(existingPipeline *kubernetes.GenericObject, newPipeline *kube
 }
 func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context, pipelineRollout *apiv1.PipelineRollout,
 	existingPipelineDef *kubernetes.GenericObject, newPipelineDef *kubernetes.GenericObject, syncStartTime time.Time) error {
+	numaLogger := logger.FromContext(ctx)
 
 	// Get the fields we need from both the Pipeline spec we have and the one we want
 	// TODO: consider having one struct which include our GenericObject plus our PipelineSpec so we can avoid multiple repeat conversions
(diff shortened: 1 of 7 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -451,11 +456,11 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 //
 // - as long as there's no other requirement to pause, set desiredPhase=Running
 func (r *PipelineRolloutReconciler) processExistingPipelineWithoutDataLoss(ctx context.Context, pipelineRollout *apiv1.PipelineRollout,
-	existingPipelineDef, newPipelineDef *kubernetes.GenericObject, pipelineNeedsToUpdate bool) error {
+	existingPipelineDef, newPipelineDef *kubernetes.GenericObject, pipelineNeedsToUpdate bool, upgradeStrategy usde.UpgradeStrategy) error {
 
 	numaLogger := logger.FromContext(ctx)
 
-	shouldBePaused, err := r.shouldBePaused(ctx, pipelineRollout, existingPipelineDef, newPipelineDef, pipelineNeedsToUpdate)
+	shouldBePaused, err := r.shouldBePaused(ctx, pipelineRollout, existingPipelineDef, newPipelineDef, pipelineNeedsToUpdate, upgradeStrategy)
```
juliev0, line 463: I'm thinking we don't need to pause the pipeline if we're using Progressive. So, we could conditionally call this only if using PPND

juliev0: Even if it's determined that pausing is an aspect of Progressive, it won't happen here I don't think.

afugazzotto: At this time, we do not have Progressive implementation, so PPND and Progressive should both pause the pipeline for now. This logic will need to be changed once Progressive is available.

internal/util/util.go
```diff
@@ -80,3 +82,165 @@ func removeNullValuesFromJSONMap(m map[string]interface{}) bool {
...
+			if len(typedVal) == 0 {
+				delete(obj, key)
+			} else {
+				obj[key] = typedVal
+			}
+
+		case nil:
+			delete(obj, key)
+
+		}
```
juliev0, line 177: will you need to incorporate my recent [fix](https://github.com/numaproj/numaplane/pull/213) to look for keys with zero values?

afugazzotto: Yes, I will add those cases since they are needed. Alternatively, I could use your `removeNullValuesFromJSONMap` instead of the `cleanup` function. What do you prefer?

internal/controller/pipelinerollout_controller.go
juliev0: I think it's a little confusing as is, though, since I don't think we need to change `shouldBePaused()` at all. There are a bunch of TODOs that you added in there that should go over here instead. Right now nobody should be using Progressive strategy at all, so we can do whatever we want if they call that, such as issue an error, or just use apply.

internal/util/util.go
juliev0: whatever you prefer. I guess the name `cleanup()` should probably at least be modified to be more clear about what it means

internal/controller/isbservicerollout_controller.go
```diff
@@ -284,7 +293,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 		isbServiceRollout.Status.MarkDeployed(isbServiceRollout.Generation)
 	}
 
-	if common.DataLossPrevention {
+	if common.DataLossPrevention { // TODO: should this check be here? `&& upgradeStrategy == usde.UpgradeStrategyPPND`
```
juliev0, line 296: regarding the `TODO`, if we keep the function below named `processChildObjectWithoutDataLoss()` I think it can include Progressive. But we could also just change the name to something like `processChildObjectPPND()`

internal/controller/isbservicerollout_controller.go
```diff
@@ -336,30 +345,15 @@ func (r *ISBServiceRolloutReconciler) getRolloutKey(rolloutNamespace string, rol
...
-	err = util.StructToStruct(&isbServiceRollout.Spec.InterStepBufferService.Spec, &newSpecAsMap)
-	if err != nil {
-		return false, false, err
+		return false, err
 	}
-	// TODO: see if we can just use DeepEqual or if there may be null values we need to ignore
-	isbServiceNeedsToUpdate := !util.CompareMapsIgnoringNulls(existingSpecAsMap, newSpecAsMap)
 
-	return isbServiceNeedsToUpdate, !isbServiceReconciled, nil
+	return !isbServiceReconciled, nil
```
juliev0, line 356: this is good having this function truly just being `isISBServiceUpdating()` - it wasn't great that it represented both things before

internal/controller/pipelinerollout_controller.go
```diff
@@ -399,20 +400,26 @@ func mergePipeline(existingPipeline *kubernetes.GenericObject, newPipeline *kube
...
+	numaLogger.WithValues("upgradeStrategy", upgradeStrategy.String()).Info("derived upgrade strategy")
+
+	var newPipelineSpec PipelineSpec
+	if err := json.Unmarshal(newPipelineDef.Spec.Raw, &newPipelineSpec); err != nil {
+		return fmt.Errorf("failed to convert new Pipeline spec %q into PipelineSpec type, err=%v", string(newPipelineDef.Spec.Raw), err)
+	}
+	specBasedPause := (newPipelineSpec.Lifecycle.DesiredPhase == string(numaflowv1.PipelinePhasePaused) || newPipelineSpec.Lifecycle.DesiredPhase == string(numaflowv1.PipelinePhasePausing))
+
+	pipelineNeedsToUpdate := (upgradeStrategy != usde.UpgradeStrategyNoOp && upgradeStrategy != usde.UpgradeStrategyError) || specBasedPause
+
```
juliev0, line 422: Perfect. Yeah, I see you want `specBasedPause` here because you'll also need it for Progressive strategy?

internal/controller/isbservicerollout_controller.go
afugazzotto: That may be best so we can have a top level switch to choose between APPLY, PPND, or PROGRESSIVE.

internal/controller/pipelinerollout_controller.go
```diff
@@ -519,13 +525,7 @@ func (r *PipelineRolloutReconciler) shouldBePaused(ctx context.Context, pipeline
...
 	// If there is a need to update, does it require a pause?
 	// Here we look for whether we are trying to update Pipeline, or whether we are already processing an update which required us to pause
-	var pipelineUpdateRequiresPause bool
-	if pipelineNeedsToUpdate || pipelineUpdating {
-		pipelineUpdateRequiresPause, err = needsPausing(existingPipelineDef, newPipelineDef)
-		if err != nil {
-			return nil, err
-		}
-	}
+	pipelineUpdateRequiresPause := pipelineNeedsToUpdate || pipelineUpdating
```
juliev0, line 528: nit: the name `pipelineUpdateRequiresPause` is a bit misleading now (since I gather all updates in this function require pausing) - perhaps we no longer need this variable set and can just include`pipelineNeedsToUpdate || pipelineUpdating` directly in the setting of `shouldBePaused` below?

juliev0:
Let me know when it's no longer a Draft and I'll go through it more closely

internal/controller/pipelinerollout_controller.go
afugazzotto: Correct, unless Progressive ends up not needing it and we can incorporate it back into the PPND logic only.

internal/controller/pipelinerollout_controller.go
afugazzotto: Good call, will make that change.

internal/controller/pipelinerollout_controller.go
juliev0: I think most likely Progressive strategy will still need to honor user's wish to have Pipeline paused if they specify it

PR #220, opened 2024-08-31
fix: fix ISBServiceRollout/NumaflowControllerRollout Pausing condition, and add unit tests for ISBServiceRollout pause feature

juliev0:
### Modifications

1. Was accidentally writing to a local copy of the `Status` for the `ISBServiceRollout` and `NumaflowControllerRollout` rather than to the actual `Status` that gets passed back to Kubernetes in the `Update()` call in `Reconcile()` function. Now passing in the `ISBServiceRollout` and `NumaflowControllerRollout` directly 
2. Decided to add unit tests for ISBServiceRollout, which is part of testing this feature
3. Removed an if statement condition which was in all 3 rollouts and is no longer needed (wasn't doing anything) 

### Verification

Reconciling ISBServiceRollout as well as NumaflowControllerRollout with `dataLossPrevention=true` causes the PausingPipelines condition to show (either True or False):
```
    - lastTransitionTime: "2024-08-31T05:22:54Z"
      message: no need for pause
      observedGeneration: 7
      reason: NoPause
      status: "False"
      type: PausingPipelines
```

I tested modification of each one (`ISBServiceRollout` and `NumaflowControllerRollout`) and observed the condition go to True and then to False.


internal/controller/isbservicerollout_controller.go
```diff
@@ -316,11 +315,9 @@ func (r *ISBServiceRolloutReconciler) updateISBService(ctx context.Context, isbS
 	return nil
 }
 
-func (r *ISBServiceRolloutReconciler) markRolloutPaused(ctx context.Context, rolloutNamespace string, rolloutName string, paused bool) error {
-	isbServiceRollout := &apiv1.ISBServiceRollout{}
-	if err := r.client.Get(ctx, k8stypes.NamespacedName{Namespace: rolloutNamespace, Name: rolloutName}, isbServiceRollout); err != nil {
-		return err
-	}
+func (r *ISBServiceRolloutReconciler) markRolloutPaused(ctx context.Context, rollout client.Object, paused bool) error {
```
juliev0, line 318: pass the rollout itself in now so that we can write directly to the status

Makefile
```diff
@@ -111,7 +111,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: codegen fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -race -short -v $$(go list ./... | grep -v /tests/e2e) 
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -p 1 -race -short -v $$(go list ./... | grep -v /tests/e2e)
```
juliev0, line 114: we can't run test suites in parallel - if we do, they stomp on each other

config/install.yaml
```diff
@@ -842,6 +851,9 @@ spec:
     - jsonPath: .status.replicas
       name: Current
       type: string
+    - jsonPath: .status.readyReplicas
+      name: Ready
+      type: string
```
juliev0, line 856: seems this is getting updated since Numaflow CRDs have been updated in `numaflow:stable`

internal/controller/isbservicerollout_controller.go
```diff
@@ -275,17 +274,15 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 
 	// set the Status appropriately to "Pending" or "Deployed"
 	// if isbServiceNeedsUpdating - this means there's a mismatch between the desired ISBService spec and actual ISBService spec
-	// if there's a generation mismatch - this means we haven't even observed the current generation
-	// we may match the first case and not the second when we've observed the generation change but we're pausing pipelines
-	// we may match the second case and not the first if we need to update something other than ISBService spec
-	if isbServiceNeedsUpdating || isbServiceRollout.Status.ObservedGeneration < isbServiceRollout.Generation {
+	// Note that this will be reset to "Deployed" later on if a deployment occurs
+	if isbServiceNeedsUpdating {
```
juliev0, line 278: the setting of `ObservedGeneration` got changed since this code was written, such that it always gets set before we get here for each reconciliation, so no need for the second half of the if statement anymore

internal/controller/pipelinerollout_controller.go
```diff
@@ -415,10 +415,7 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 
 	// set the Status appropriately to "Pending" or "Deployed"
 	// if pipelineNeedsToUpdate - this means there's a mismatch between the desired Pipeline spec and actual Pipeline spec
-	// if there's a generation mismatch - this means we haven't even observed the current generation
-	// we may match the first case and not the second when we've observed the generation change but we're pausing
-	// we may match the second case and not the first if we need to update something other than Pipeline spec
-	if pipelineNeedsToUpdate || pipelineRollout.Status.ObservedGeneration < pipelineRollout.Generation {
+	if pipelineNeedsToUpdate {
```
juliev0, line 418: same comment as for [ISBServiceRollout](https://github.com/numaproj/numaplane/pull/220/files#r1742375855)

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -269,16 +269,14 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
 
 		// set the Status appropriately to "Pending" or "Deployed"
 		// if controllerDeploymentNeedsUpdating - this means there's a mismatch between the desired NumaflowControllerRollout version and actual
-		// if there's a generation mismatch - this means we haven't even observed the current generation
-		// we may match the first case and not the second when we've observed the generation change but we're pausing pipelines
-		// we may match the second case and not the first if we need to update something other than Numaflow Controller version
-		if controllerDeploymentNeedsUpdating || controllerRollout.Status.ObservedGeneration < controllerRollout.Generation {
+		// Note that this will be reset to "Deployed" later on if a deployment occurs
+		if controllerDeploymentNeedsUpdating {
```
juliev0, line 273: same comment as for [ISBServiceRollout](https://github.com/numaproj/numaplane/pull/220/files#r1742375855)

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -697,16 +695,13 @@ func (r *NumaflowControllerRolloutReconciler) processNumaflowControllerStatus(ct
...
-	}
+func (r *NumaflowControllerRolloutReconciler) markRolloutPaused(ctx context.Context, rollout client.Object, paused bool) error {
+	controllerRollout := rollout.(*apiv1.NumaflowControllerRollout)
 
 	if paused {
-		rollout.Status.MarkPausingPipelines(rollout.Generation)
+		controllerRollout.Status.MarkPausingPipelines(controllerRollout.Generation)
 	} else {
-		rollout.Status.MarkUnpausingPipelines(rollout.Generation)
+		controllerRollout.Status.MarkUnpausingPipelines(controllerRollout.Generation)
```
juliev0, line 704: same comment as for [ISBServiceRollout](https://github.com/numaproj/numaplane/pull/220/files#r1739970803)

internal/controller/test_common.go
```diff
@@ -0,0 +1,148 @@
...
+	numaflowv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+	numaflowversioned "github.com/numaproj/numaflow/pkg/client/clientset/versioned"
+	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+var (
+	defaultNamespace = "default"
+)
```
juliev0, line 25: some of this code is somewhat duplicative of suite_test.go (`BeforeSuite()`) but my plan is that that will be removed once we replace the Ginkgo unit testing (for which "unit" is a struct) with unit testing that makes it so the "unit" is a function

internal/controller/suite_test.go
```diff
@@ -78,6 +79,9 @@ func TestControllers(t *testing.T) {
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+	//cfg, _, _, _, err := prepareK8SEnvironment()
+	//Expect(err).NotTo(HaveOccurred())
```
juliev0, line 83: I was trying to use the same common function here for both the new unit test and the previous unit tests, but for some reason it wasn't working here. I decided not to dwell on that since we plan on replacing the ginkgo struct-level unit tests with function-level unit tests.

internal/controller/suite_test.go
```diff
@@ -297,5 +306,5 @@ func verifyStatusPhase(ctx context.Context, gvk schema.GroupVersionKind, namespa
 		}
 
 		return apiv1.Phase(phase) == desiredPhase && observedGeneration == generation, nil
-	}, duration, interval).Should(BeTrue())
+	}, timeout, interval).Should(BeTrue())
```
juliev0, line 309: discussed with Tony. I believe this makes more sense as an `Eventually` test.

internal/controller/suite_test.go
xdevxy: do you mind creating an issue for tracking this? Thanks!

xdevxy (approved):

Makefile
juliev0: I'm seeing ~10 minutes instead of ~8 minutes

internal/controller/suite_test.go
juliev0: okay, I'm partly thinking it will be de-prioritized in favor of removing the test at some point, but in case we don't end up removing it, the issue can be there

Makefile
juliev0: another option in the future would be to create parallelization through a matrix of tests that each test a subset of the tests.

We can explore how other repositories are dealing with this problem.

juliev0: Actually, what I read is that if you don't use `-p 1`, then the default behavior is that "multiple test suites" run in parallel. I think other repos aren't using test suites at all, so I believe that their unit tests run sequentially. Just ours is using the Ginkgo test suite.

PR #227, opened 2024-09-05
feat: USDE Comparison Logic

### LINKED ISSUE #176: Implement USDE Comparison

afugazzotto:
Fixes #176 

### Modifications

Implemented USDE main logic that uses USDE config and Namespace-level related parameters to decide the upgrade strategy to use. The changes also include adding the `inProgressUpgrade` field to the PipelineRollout status.

### Verification

Unit tests and manual tests.


config/samples/namespace-level-config.yaml
```diff
@@ -6,5 +6,6 @@ metadata:
   labels:
     numaplane.numaproj.io/config: namespace-level-config
 data:
+  # TODO-PROGRESSIVE: before the PROGRESSIVE strategy is implemented, users will only be able to choose "pause-and-drain". Afterwards, "progressive" should also be an option. Remove this comment line after implementing PROGRESSIVE strategy.
```
afugazzotto, line 9: You will find `TODO-PROGRESSIVE` comments in other files as well. These comments will need to be addressed when the PROGRESSIVE strategy is implemented.

internal/controller/config/usde_config.go
```diff
@@ -0,0 +1,71 @@
...
+		// TODO-PROGRESSIVE: replace line above for line below
+		// return fmt.Errorf("empty strategy (allowed values are: %s or %s)", ProgressiveStrategyID, PPNDStrategyID)
+	}
+
+	// Remove the double quotes around the string
+	strategyStr := string(data[1 : len(data)-1])
+
+	// Make sure the string is one of the possible strategy values
+	if strategyStr != PPNDStrategyID {
+		return fmt.Errorf("invalid strategy %s (allowed value is: %s)", string(data[:]), PPNDStrategyID)
```
juliev0, line 56: this seems unusual to try to preprocess the `[]byte` before doing the Unmarshal. Can the same not be achieved by unmarshalling and then validating?

internal/usde/usde.go
```diff
@@ -1,39 +1,98 @@
...
+		return UpgradeStrategyError, nil, err
+	}
+
+	numaLogger.WithValues("existingSpecOnlyApplyPaths", existingSpecOnlyApplyPaths, "existingSpecWithoutApplyPaths", existingSpecWithoutApplyPaths).Debug("split existing spec")
+
+	specsDiffer := true
+
+	// Compare specs without the apply fields and check user's strategy to either return PPND or Progressive
+	if !reflect.DeepEqual(newSpecWithoutApplyPaths, existingSpecWithoutApplyPaths) {
+		userUpgradeStrategy := config.GetConfigManagerInstance().GetNamespaceConfig(newSpec.Namespace).UpgradeStrategy
```
juliev0, line 75: is it possible for the `GetNamespaceConfig()` function to fail if there's no config for the namespace? Can we have some default value if that file doesn't exist?

internal/controller/config/usde_config.go
afugazzotto: fixed

internal/usde/usde.go
```diff
@@ -1,39 +1,103 @@
...
+
+		if userUpgradeStrategy == config.PPNDStrategyID {
+			return UpgradeStrategyPPND, &specsDiffer, nil
+		}
 
-	// Make sure the string is one of the possible strategy values
-	if strategyStr != progressiveStrategyID && strategyStr != ppndStrategyID {
-		return fmt.Errorf("invalid strategy %s (allowed values are: %s or %s)", string(data[:]), progressiveStrategyID, ppndStrategyID)
+		// TODO-PROGRESSIVE: return UpgradeStrategyProgressive instead of UpgradeStrategyPPND
+		return UpgradeStrategyPPND, &specsDiffer, nil
```
juliev0, line 89: am I correct that right now we're only doing PPND? so, basic CRUD is not an option for users?

juliev0: I believe this is what we discussed with Vigith - just want to confirm it

juliev0 (approved):

internal/usde/usde.go
afugazzotto: If I recall correctly, that is what we decided and we will use the feature flag to control enabling/disabling data loss prevention via PPND.

internal/controller/config/usde_config.go
```diff
@@ -0,0 +1,59 @@
...
+		return fmt.Errorf("invalid strategy '%s' (allowed value is: %s)", usdeUserStrategyStr, PPNDStrategyID)
+	}
+	// TODO-PROGRESSIVE: replace if-statement above for if-statement below
+	// if usdeUserStrategyStr != string(ProgressiveStrategyID) && usdeUserStrategyStr != string(PPNDStrategyID) {
+	// 	return fmt.Errorf("invalid strategy '%s' (allowed values are: %s or %s)", usdeUserStrategyStr, ProgressiveStrategyID, PPNDStrategyID)
+	// }
+
+	*s = USDEUserStrategy(usdeUserStrategyStr)
+
+	return nil
```
juliev0, line 58: nit: admittedly, I do find it a little confusing that the cluster-level configuration stuff and the user namespace-level configuration stuff are all in one file. I think this may confuse people?

juliev0: I think it would be less confusing to have all user namespace level stuff collocated

afugazzotto: I favored organizing by feature. In this case since the user strategy is for USDE, I put it in this file. If you think it's worth moving it, I can do that.

juliev0: no worries, I will look at possibly moving it later! thanks!

PR #234, opened 2024-09-06
feat: add pause metrics for ISBService/NumaflowController rollouts

### LINKED ISSUE #222: Add metrics for Numaflow Controllers or ISBServices that are taking a long time to get updated because they have paused the pipelines for a long time

dpadhiar:
Fixes #222 

### Modifications

Introduces Gauge metrics called numaflow_isbservice_paused_seconds and numaflow_controller_paused_seconds which reports the total time an ISBService or NumaflowController is requesting resources be paused when it is updating. This is in line with the previously added PipelineRollout pause metrics that we emit these when dataLossPrevention flag is set to True.

We add the PauseStatus struct to both the ISBService and NumaflowController Rollouts however we have changed the JSON name to `pauseRequestStatus` as we don't want users to assume the ISBService or controller themselves are being paused.

Verification

Manually testing the changes by continuously updating NumaflowController and ISBServiceRollouts with dataLossPrevention enabled to examine the pause seconds are accurate.

(no unit testing now as we don't have dataLossPrevention enabled in those cases)


xdevxy (approved):

xdevxy (commented):
@juliev0 and I discussed how to add a label for the metrics to distinguish if the pause request is coming from user or numaplane. However we can do as a followup.

juliev0:
I like your very complete github PR descriptions :)

pkg/apis/numaplane/v1alpha1/isbservicerollout_types.go
```diff
@@ -36,7 +36,8 @@ type InterStepBufferService struct {
 
 // ISBServiceRolloutStatus defines the observed state of ISBServiceRollout
 type ISBServiceRolloutStatus struct {
-	Status `json:",inline"`
+	Status      `json:",inline"`
+	PauseStatus `json:"pauseRequestStatus,omitempty"`
```
juliev0, line 40: it appears that `PauseStatus` is inline? I think I'd prefer that it not be. Can you change it to 
`PauseRequestStatus PauseStatus `json:"pauseRequestStatus,omitempty"`?

juliev0: I didn't catch that the `PipelineRolloutStatus.PauseStatus` is the same - can that be changed as well?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -696,17 +697,31 @@ func (r *NumaflowControllerRolloutReconciler) processNumaflowControllerStatus(co
 }
 
 func (r *NumaflowControllerRolloutReconciler) markRolloutPaused(ctx context.Context, rollout client.Object, paused bool) error {
+
 	controllerRollout := rollout.(*apiv1.NumaflowControllerRollout)
 
 	if paused {
+		if controllerRollout.Status.PauseRequestStatus.LastPauseBeginTime == metav1.NewTime(initTime) || !controllerRollout.Status.PauseRequestStatus.LastPauseBeginTime.After(controllerRollout.Status.PauseRequestStatus.LastPauseEndTime.Time) {
```
juliev0, line 704: please add comments

dpadhiar: Used same comments as PipelineRollout since they're the same logic.

juliev0 (approved):

PR #238, opened 2024-09-09
fix: checking Pipeline conditions for PPND, plus don't pause failed pipelines

### LINKED ISSUE #230: Don't pause Failed pipelines
# Summary

Based on the contract with Numaflow, we will only try to pause pipelines that are not failed - those are the ones that are guaranteed to either pause or timeout. 

Agreement with Derek is that we can do an "Apply" directly on Pipelines which are not paused.

---

juliev0:
Fixes #230 

#### Modification 1: Don't pause "Failed" pipelines:

1. I don't pause pipelines if they're "Failed"
2. I can update the pipeline spec without waiting for a "Failed" pipeline to be "Paused"
3. I can update NumaflowController and ISBService specs if Pipelines are either "Paused" or "Failed"

##### Verification

1. unit test checks ISBService (same code is used for Numaflow Controller so presumably that is being verified at the same time)
2. manually tested this sequence: create bad pipeline which failed validation; attempted to update failed pipeline with a change which didn't fix the problem; updated failed pipeline by fixing the validation error
3. manually tested updating ISBServiceRO and NCRO while pipeline was failed 

#### Modification 2: Fix Pipeline Condition checking for PPND

Was accidentally checking the `PipelineRollout.Conditions` rather than `Pipeline.Conditions` when waiting for the Pipeline to be fully reconciled. 

##### Verification

I observed in the log that at the time we resumed the Pipeline all of the Child Conditions were in a healthy state. I also observed that the daemon was up and running.

#### Modification 3: e2e test

Have incorporated latest Numaflow release into e2e test.
Have increased suite timeout and added a comment in case we need to increase further in the future.
Have fixed the Makefile which was setting environment variable for the test related to bringing up a separate kubernetes environment, which we aren't using.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,9 @@ test: codegen fmt vet envtest ## Run tests.
 
 .PHONY: test-e2e
 test-e2e: codegen fmt vet envtest ## Run e2e tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" GOFLAGS="-count=1" go test -v ./tests/e2e/... 
+	GOFLAGS="-count=1" go test -v ./tests/e2e/... 
+
+	
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 GOLANGCI_LINT_VERSION ?= v1.58.0
--- a/internal/controller/isbservicerollout_controller_test.go
+++ b/internal/controller/isbservicerollout_controller_test.go
@@ -325,6 +325,18 @@ func Test_reconcile_PPND(t *testing.T) {
 			expectedISBSvcSpec: createDefaultISBServiceSpec("2.10.11"),
 		},
 		{
+			name:                   "existing ISBService - new spec - pipelines failed",
+			newISBSvcSpec:          createDefaultISBServiceSpec("2.10.11"),
+			existingISBSvcDef:      createDefaultISBService("2.10.3", numaflowv1.ISBSvcPhaseRunning, true),
+			existingStatefulSetDef: createDefaultISBStatefulSet("2.10.3", true),
+			existingPipelinePhase:  numaflowv1.PipelinePhaseFailed,
+			expectedRolloutPhase:   apiv1.PhaseDeployed,
+			expectedConditionsSet: map[apiv1.ConditionType]metav1.ConditionStatus{
+				apiv1.ConditionChildResourceDeployed: metav1.ConditionTrue,
+			},
+			expectedISBSvcSpec: createDefaultISBServiceSpec("2.10.11"),
+		},
+		{
 			name:                   "existing ISBService - spec already updated - isbsvc reconciling",
 			newISBSvcSpec:          createDefaultISBServiceSpec("2.10.11"),
 			existingISBSvcDef:      createDefaultISBService("2.10.11", numaflowv1.ISBSvcPhaseRunning, false),
--- a/internal/controller/pause_requester.go
+++ b/internal/controller/pause_requester.go
@@ -48,7 +48,7 @@ func processChildObjectWithoutDataLoss(ctx context.Context, rollout client.Objec
 		// Don't do this yet if we just made a request - it's too soon for anything to have happened
 		if !pauseRequestUpdated && resourceNeedsUpdating {
 
-			// check if the pipelines are all paused
+			// check if the pipelines are all paused (or can't be paused)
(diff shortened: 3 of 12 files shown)


juliev0:
I will do some manual testing of pipeline changes before marking this "Ready for Review"

Makefile
```diff
@@ -115,7 +115,9 @@ test: codegen fmt vet envtest ## Run tests.
 
 .PHONY: test-e2e
 test-e2e: codegen fmt vet envtest ## Run e2e tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" GOFLAGS="-count=1" go test -v ./tests/e2e/... 
+	GOFLAGS="-count=1" go test -v ./tests/e2e/...
```
juliev0, line 118: don't think we need `KUBEBUILDER_ASSETS` which seems to set up a Kubernetes environment, which apparently sets up a Kubernetes environment. In the test we use our [existing](https://github.com/numaproj/numaplane/blob/main/tests/e2e/suite_test.go#L60) cluster.

internal/controller/isbservicerollout_controller_test.go
```diff
@@ -324,6 +324,18 @@ func Test_reconcile_PPND(t *testing.T) {
...
+			newISBSvcSpec:          createDefaultISBServiceSpec("2.10.11"),
+			existingISBSvcDef:      createDefaultISBService("2.10.3", numaflowv1.ISBSvcPhaseRunning, true),
+			existingStatefulSetDef: createDefaultISBStatefulSet("2.10.3", true),
+			existingPipelinePhase:  numaflowv1.PipelinePhaseFailed,
+			expectedRolloutPhase:   apiv1.PhaseDeployed,
+			expectedConditionsSet: map[apiv1.ConditionType]metav1.ConditionStatus{
+				apiv1.ConditionChildResourceDeployed: metav1.ConditionTrue,
+			},
+			expectedISBSvcSpec: createDefaultISBServiceSpec("2.10.11"),
+		},
```
juliev0, line 338: testing that ISBServiceRollout will ignore Failed pipelines for pausing

tests/e2e/common.go
```diff
@@ -21,12 +21,13 @@ import (
...
-	cancel        context.CancelFunc
-	suiteTimeout  = 5 * time.Minute
-	testTimeout   = 2 * time.Minute
+	dynamicClient       dynamic.DynamicClient
+	testEnv             *envtest.Environment
+	ctx                 context.Context
+	cancel              context.CancelFunc
+	suiteTimeout        = 30 * time.Minute
+	testTimeout         = 2 * time.Minute
+	testPollingInterval = 1 * time.Second
```
juliev0, line 30: I think it will be better to poll the K8S API once per second rather than the default once per 10ms just in case we ever do face a rate limiting issue

tests/manifests/default/controller_definitions.yaml
```diff
@@ -318,7 +318,7 @@ controllerDefinitions:
             - controller
             env:
             - name: NUMAFLOW_IMAGE
-              value: quay.io/numaproj/numaflow:v1.3.0
```
juliev0, line 321: this diff is weird - note that I didn't remove 1.3.0

internal/controller/pause_requester.go
```diff
@@ -49,12 +49,12 @@ func processChildObjectWithoutDataLoss(ctx context.Context, rollout client.Objec
 		if !pauseRequestUpdated && resourceNeedsUpdating {
 
 			// check if the pipelines are all paused
```
xdevxy, line 51: nit: update the comment accordingly

internal/controller/pipelinerollout_controller.go
```diff
@@ -568,9 +570,12 @@ func pipelineIsUpdating(pipelineRollout *apiv1.PipelineRollout, newPipelineDef *
 	}
 
 	// note if Pipeline's children are still being updated
-	_, pipelineChildResourceReason := getPipelineChildResourceHealth(pipelineRollout.Status.Conditions)
+	unhealthyOrProgressing, _ := checkChildResources(existingPipelineStatus.Conditions, func(c metav1.Condition) bool {
+		// TODO: once we have "inProgressStrategy" we can just look for c.Status == metav1.ConditionFalse while inProgressStrategy==PPND
```
xdevxy, line 574: is there a numaflow issue for this? can we add it here in the comment?

xdevxy (approved):

internal/controller/pipelinerollout_controller.go
juliev0: no longer a numaflow issue but rather just a numaplane issue which I can reference in the comment: https://github.com/numaproj/numaplane/issues/239

PR #244, opened 2024-09-10
chore: add a hash suffix based on the spec for pipeline

### LINKED ISSUE #243: Support multiple pipelines can coexist for a PipelineRollout
Currently, the name for pipeline and pipelineRollout are the same. When we need to have multiple pipelines map to the same PipelineRollout this become a problem.

xdevxy:
Fixes #243 

### Modifications

Since we may have multiple pipelines associate with a pipeline rollout soon for the no downtime upgrade feature. This PR adds the logic to uniquely identify a pipeline that associate with a pipeline rollout based on its spec. To deterministic identify a pipeline, added a hash of the pipeline spec as a suffix to the name of the pipeline. 

This PR also fixes a bug that when a pipeline with the same name already exist without a pipeline Rollout manage it. It should update the phase to failed without checking the child status. 

### Verification

```
make test
make test-e2e
```


juliev0:
hmm, looks like e2e failed?

internal/controller/pipelinerollout_controller.go
```diff
@@ -836,20 +844,17 @@ func updatePipelineSpec(ctx context.Context, restConfig *rest.Config, obj *kuber
...
-	}
-	if err := json.Unmarshal(pipelineRollout.Spec.Pipeline.Spec.Raw, &pipelineSpec); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal pipeline spec: %v", err)
-	}
+func pipelineLabels(pipelineSpec *numaflowv1.PipelineSpec) (map[string]string, error) {
+	labelMapping := make(map[string]string)
 	if pipelineSpec.InterStepBufferServiceName != "" {
 		labelMapping[common.LabelKeyISBServiceNameForPipeline] = pipelineSpec.InterStepBufferServiceName
+	} else {
+		labelMapping[common.LabelKeyISBServiceNameForPipeline] = "default"
```
juliev0, line 852: maybe doesn't need to be done in this PR, but are we going to add a label for PipelineRollout name? Then we can easily find all Pipelines using that PipelineRollout. I know the Pipeline name will be fixed with the PipelineRollout name, but I think this would enable easy selection both through kubectl and potentially in the code.

juliev0: unless we can do a query on OwnerReference I guess...but just to note that Numaflow includes labels like this

juliev0: FYI, I have just created [this](https://github.com/numaproj/numaplane/issues/245) issue which depends on this

xdevxy: yeah, that will be the follow up PR for this.

xdevxy:
Decided to postponed this PR as it can introduce the issue that multiple pipelines associate with a pipelineRollout after a spec change.

PR #247, opened 2024-09-11
test: check rollout conditions in e2e tests

### LINKED ISSUE #96: E2E Tests for Rollout Conditions
# Summary

Add E2E test cases to verify that the status conditions for all rollouts are set has expected.
Example:
```
Eventually(func() (metav1.ConditionStatus, error) {
	createdResource := &apiv1.NumaflowControllerRollout{}
	Expect(k8sClient.Get(ctx, resourceLookupKey, createdResource)).To(Succeed())

	for _, cond := range createdResource.Status.Conditions {
		if cond.Type == string(apiv1.ConditionChildResourceHealthy) {
			return cond.Status, nil
		}
	}

	return metav1.ConditionUnknown, nil
}, timeout, interval).Should(Equal(metav1.ConditionTrue))
```

# Use Cases

Validate correctness of status conditions.

---

dpadhiar:
Fixes #96 

### Modifications

Adds a common function to get a specific condition from a rollout's array of condition to then check the status of it.

For our e2e tests, we are checking the `ChildResourcesHealthyCondition` for all rollouts and extra for PipelineRollout the `PipelinePausedOrPausing` condition to be sure they are true or false in the correct cases.

TODO: We will want to include an env variable to signify when dataLossPrevention is enabled. This way we will appropriately check that pipelines are being paused during rollout updates.

```
func getRolloutCondition(conditions []metav1.Condition, conditionType apiv1.ConditionType) metav1.ConditionStatus {
	for _, cond := range conditions {
		if cond.Type == string(conditionType) {
			return cond.Status
		}
	}
	return metav1.ConditionUnknown
}
```

### Verification

Ran `make test-e2e` and Github CI.

### DIFF
--- a/tests/e2e/functional_test.go
+++ b/tests/e2e/functional_test.go
@@ -287,6 +287,17 @@ var _ = Describe("Functional e2e", Serial, func() {
 			return *retrievedPipelineSpec.Vertices[0].Source.Generator.RPU == int64(10)
 		})
 
+		document("Verifying that the PipelineRollout conditions are as expected")
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceHealthy)
+		}).WithTimeout(testTimeout).Should(Equal(metav1.ConditionTrue))
+
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
+		}).WithTimeout(testTimeout).Should(Equal(metav1.ConditionFalse))
+
 		verifyPipelineReady(Namespace, pipelineRolloutName, 2)
 
 	})
@@ -308,6 +319,12 @@ var _ = Describe("Functional e2e", Serial, func() {
 		verifyNumaflowControllerDeployment(Namespace, func(d appsv1.Deployment) bool {
 			return d.Spec.Template.Spec.Containers[0].Image == "quay.io/numaio/numaflow-rc:v0.0.6"
 		})
+
+		document("Verifying that the NumaflowControllerRollout conditions are as expected")
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceHealthy)
+		}).WithTimeout(testTimeout).Should(Equal(metav1.ConditionTrue))
 
 		verifyNumaflowControllerReady(Namespace)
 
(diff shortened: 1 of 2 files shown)


tests/e2e/functional_test.go
```diff
@@ -135,10 +135,16 @@ var _ = Describe("Functional e2e", Serial, func() {
...
-			_, err := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+			_, err = numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
 			return err
 		}, testTimeout, testPollingInterval).Should(Succeed())
 
+		document("Verifying that the NumaflowControllerRollout conditions are as expected")
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceHealthy)
+		}).WithTimeout(testTimeout).Should(Equal(metav1.ConditionTrue))
```
juliev0, line 146: debating whether to bother you with this or not: I just updated all of these `Eventually()` calls to include a `testPollingInterval` (1 sec) instead of using the default 10 ms.  I wasn't sure if we could potentially stress the kubernetes client rate limiter by hitting it every 10ms or not.

dpadhiar: Can make this change really quick, no worries there.

tests/e2e/functional_test.go
```diff
@@ -277,6 +324,12 @@ var _ = Describe("Functional e2e", Serial, func() {
 			return d.Spec.Template.Spec.Containers[0].Image == "quay.io/numaio/numaflow-rc:v0.0.13"
 		})
 
+		document("Verifying that the NumaflowControllerRollout conditions are as expected")
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceHealthy)
+		}).WithTimeout(testTimeout).Should(Equal(metav1.ConditionTrue))
+
```
juliev0, line 332: how about the `Condition.PipelinesPausingOrPaused` for this and ISBSvc?

dpadhiar: Ack, resigned those in my head to just the pausing test. Will add those in all these checks.

juliev0: Thanks. I didn't realize you had to use `ConditionUnknown`. But I guess that's the only thing that works for both feature flag on and feature flag off? With feature flag, it would only be unknown briefly and then hopefully switch to `ConditionTrue` and then `ConditionFalse`. If so, maybe you were right to postpone that for when you have the environment variable ready after all? (sorry, I didn't recognize this when I asked)

juliev0: I think checking `ConditionUnknown` is not very useful right now...

dpadhiar: That's right, I have to check for `Unknown` at this time as PausingPipelines condition isn't present unless the flag is up. As for the environment variable, I could do that in a follow up PR or continue to add to this initial PR.

juliev0: follow up PR seems fine!

dpadhiar: Awesome, will just turn these current checks for PausingPipelines into comments for now.

juliev0 (approved):

juliev0:
FYI, @dpadhiar will do a follow up PR to verify the `observedGeneration` in the Condition

tests/e2e/functional_test.go
```diff
@@ -139,6 +139,23 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceDeployed)
+		}, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
+
+		Eventually(func() metav1.ConditionStatus {
+			rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceHealthy)
+		}, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
+
```
xdevxy, line 152: can you create an issue to track this todo? Thanks!

xdevxy (approved):

xdevxy:
> FYI, @dpadhiar will do a follow up PR to verify the `observedGeneration` in the Condition

@dpadhiar do you mind also add an issue for this one if you are not picking up the follow up soon? Thanks!

PR #249, opened 2024-09-12
feat: support consecutive Updates to our dynamic objects; also remove use of dynamic client for Rollouts (not needed)

juliev0:
### Modifications

1. the `Update()` function in the controller-runtime framework always returns the new object with the correct resourceVersion in it (as mentioned in [this](https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332) article). I wanted to do the same thing for our dynamic client Update functions. This way if we call it consecutively, we are unlikely to have any resource version conflicts that cause us to return an error. **Verified this by:** writing a unit test for it to confirm the updated resourceVersion gets returned from an `UpdateCR()` call and that we can make a subsequent call to `UpdateCR()`.
2. I noticed that we were using the dynamic client to update the Rollout CRs, which wasn't necessary, so I changed it to use the standard controller-runtime framework, thereby making the code simpler. Verified the calls to update status are working by running e2e and unit tests, and also updating the e2e test to check for `Rollout.Status.Phase` which it previously wasn't.
3. Removed the requirement from the unit tests that the "sync failure" metrics be 0. Normal resource version conflicts can cause this to increment. Note that the change above does cause there to be some errors issued some of the time where there weren't before (the previous code just updated `spec` and `status` and not any metadata such as `resourceVersion` so there was never a conflict), but actually conflicts are a good thing because it means we won't ignore more recent updates that have been made on the object.

### DIFF
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -540,26 +540,7 @@ func (r *ISBServiceRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *ISBServiceRolloutReconciler) updateISBServiceRolloutStatus(ctx context.Context, isbServiceRollout *apiv1.ISBServiceRollout) error {
-	rawSpec := runtime.RawExtension{}
-	err := util.StructToStruct(&isbServiceRollout.Spec, &rawSpec)
-	if err != nil {
-		return fmt.Errorf("unable to convert ISBServiceRollout Spec to GenericObject Spec: %v", err)
-	}
-
-	rawStatus := runtime.RawExtension{}
-	err = util.StructToStruct(&isbServiceRollout.Status, &rawStatus)
-	if err != nil {
-		return fmt.Errorf("unable to convert ISBServiceRollout Status to GenericObject Status: %v", err)
-	}
-
-	obj := kubernetes.GenericObject{
-		TypeMeta:   isbServiceRollout.TypeMeta,
-		ObjectMeta: isbServiceRollout.ObjectMeta,
-		Spec:       rawSpec,
-		Status:     rawStatus,
-	}
-
-	return kubernetes.UpdateStatus(ctx, r.restConfig, &obj, "isbservicerollouts")
+	return r.client.Status().Update(ctx, isbServiceRollout)
 }
 
 func (r *ISBServiceRolloutReconciler) updateISBServiceRolloutStatusToFailed(ctx context.Context, isbServiceRollout *apiv1.ISBServiceRollout, err error) error {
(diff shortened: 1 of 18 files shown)


internal/controller/isbservicerollout_controller.go
```diff
@@ -540,26 +540,7 @@ func (r *ISBServiceRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-
-	obj := kubernetes.GenericObject{
-		TypeMeta:   isbServiceRollout.TypeMeta,
-		ObjectMeta: isbServiceRollout.ObjectMeta,
-		Spec:       rawSpec,
-		Status:     rawStatus,
-	}
-
-	return kubernetes.UpdateStatus(ctx, r.restConfig, &obj, "isbservicerollouts")
+	return r.client.Status().Update(ctx, isbServiceRollout)
```
juliev0, line 543: don't need dynamic client for this

tests/e2e/isbservice.go
```diff
@@ -50,6 +50,31 @@ func verifyISBServiceSpec(namespace string, name string, f func(numaflowv1.Inter
 	}, testTimeout, testPollingInterval).Should(BeTrue())
 }
 
+func verifyISBSvcRolloutReady(isbServiceRolloutName string) {
+	document("Verifying that the ISBServiceRollout is ready")
+
+	Eventually(func() bool {
+		rollout, _ := isbServiceRolloutClient.Get(ctx, isbServiceRolloutName, metav1.GetOptions{})
+		return rollout.Status.Phase == apiv1.PhaseDeployed
+	}, testTimeout, testPollingInterval).Should(BeTrue())
```
juliev0, line 59: added `Phase` check to all of the Rollouts

dpadhiar (commented):
Changes to e2e testing LGTM!

tests/e2e/isbservice.go
```diff
@@ -50,6 +50,31 @@ func verifyISBServiceSpec(namespace string, name string, f func(numaflowv1.Inter
...
+		rollout, _ := isbServiceRolloutClient.Get(ctx, isbServiceRolloutName, metav1.GetOptions{})
+		return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceDeployed)
+	}, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
+
+	Eventually(func() metav1.ConditionStatus {
+		rollout, _ := isbServiceRolloutClient.Get(ctx, isbServiceRolloutName, metav1.GetOptions{})
+		return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionChildResourceHealthy)
+	}, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
+
+	// Eventually(func() metav1.ConditionStatus {
```
xdevxy, line 71: nit: why this is commented out?

xdevxy (approved):

tests/e2e/isbservice.go
dpadhiar: This is being worked on in my current PR, this check requires dataLossPrevention flag to be on.

PR #253, opened 2024-09-12
fix: return error when a pipeline with the same name already exist without a pipeline Rollout manage it

### LINKED ISSUE #252: Return reconciliation error when a pipeline with the same name already exist without a corresponding rollout
# Summary

Currently, we swallowed the error and directly update the phase in the status when a pipeline with the same name already exist without a pipeline Rollout manage it. We should instead return the error as reconciliation failure so the error is recorded by the metrics via the common error handler.

---

xdevxy:
Fixes #252 

### Modifications

This PR returns the reconciliation error when a pipeline with the same name already exist without a pipeline Rollout manage it. So the error is recorded by the metrics via the common error handler.

### Verification

`make test`

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -370,8 +370,9 @@ func (r *PipelineRolloutReconciler) reconcile(
 	// Object already exists
 	// if Pipeline is not owned by Rollout, fail and return
 	if !checkOwnerRef(existingPipelineDef.OwnerReferences, pipelineRollout.UID) {
-		numaLogger.Debugf("PipelineRollout %s failed because Pipeline %s already exists in namespace", pipelineRollout.Name, existingPipelineDef.Name)
-		return false, fmt.Errorf("pipeline %s already exists in namespace", existingPipelineDef.Name)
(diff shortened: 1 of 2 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -370,8 +370,8 @@ func (r *PipelineRolloutReconciler) reconcile(
 	// Object already exists
 	// if Pipeline is not owned by Rollout, fail and return
 	if !checkOwnerRef(existingPipelineDef.OwnerReferences, pipelineRollout.UID) {
-		pipelineRollout.Status.MarkFailed(fmt.Sprintf("Pipeline %s already exists in namespace", pipelineRollout.Name))
-		return false, nil
+		numaLogger.Debugf("PipelineRollout %s failed because Pipeline %s already exists in namespace", pipelineRollout.Name, existingPipelineDef.Name)
+		return false, fmt.Errorf("pipeline %s already exists in namespace", existingPipelineDef.Name)
```
juliev0, line 374:
```suggestion
		errStr := fmt.Sprintf("Pipeline %s already exists in namespace, not owned by PipelineRollout", existingPipelineDef.Name)
		numaLogger.Debugf("PipelineRollout %s failed because %s", pipelineRollout.Name, errStr)
		return false, fmt.Error(errStr)
```

juliev0 (approved):

Issue #233, opened 2024-09-06
Add Rollout Health Metrics 


xdevxy:
# Summary

We should add Rollout Health Metrics for each type using the phase information in status.

# Use Cases

Health metrics will be used for incident indication.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

chandankumar4:
I had thought about the implementation about it, where the metrics data should looks like below, But if the phase changes then it will be a new metrics all together will be become problem.
```
numaplane_rollout_health_status{intuit_alert="true",phase="Deployed",type="isbservice-rollout-controller"} 0
numaplane_rollout_health_status{intuit_alert="true",phase="Deployed",type="pipeline-rollout-controller"} 0
numaplane_rollout_health_status{intuit_alert="true",phase="Deployed",type="numaflow-controller-rollout-controller"} 0
```

xdevxy:
We can refer to how numaflow is adding the health [metrics](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/metrics.go#L35-L46).

PR #256, opened 2024-09-13
feat: Add health metrics for monovertex

chandankumar4:
Fixes: https://github.com/numaproj/numaplane/issues/233

### Modifications

Add rollout health metrics for monovertex

### Verification
- verified in local k8s cluster
```bash
# HELP numaplane_monovertex_rollout_health A metric to indicate whether the MonoVertex is healthy. '1' means healthy, '0' means unhealthy
# TYPE numaplane_monovertex_rollout_health gauge
numaplane_monovertex_rollout_health{intuit_alert="true",monovertex="my-monovertex",namespace="example-namespace"} 1
```

### DIFF
--- a/pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
+++ b/pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
@@ -63,7 +63,7 @@ func init() {
 	SchemeBuilder.Register(&MonoVertexRollout{}, &MonoVertexRolloutList{})
 }
 
-// IsHealthy indicates whether the InterStepBufferService is healthy or not
+// IsHealthy indicates whether the MonoVertexRollout is healthy.
 func (mv *MonoVertexRolloutStatus) IsHealthy() bool {
 	if mv.Phase != PhaseDeployed {
 		return false
--- a/pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
+++ b/pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
@@ -65,5 +65,5 @@ func init() {
 
 // IsHealthy indicates whether the MonoVertexRollout is healthy.
 func (mv *MonoVertexRolloutStatus) IsHealthy() bool {
-	return mv.Phase == PhaseDeployed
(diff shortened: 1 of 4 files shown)


pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -201,3 +201,16 @@ func (s *Status) markTypeStatus(t ConditionType, status metav1.ConditionStatus,
 		ObservedGeneration: generation,
 	})
 }
+
+// IsReady returns true when all the conditions are true
+func (s *Status) IsReady() bool {
```
xdevxy, line 206: when we consider rollout object isHealthy, we only need to consider if the deployment/orchestration is successful. we definitely not want to include if the underlying pipeline is healthy or not. because numaflow already expose metrics for that. So I think we only need to consider `ChildResourceDeployed` condition.

chandankumar4: In that case checking only `phase` is enough, I believe we don't need to check for condition `ChildResourceDeployed` as phase status condition as getting set at same time [here](https://github.com/numaproj/numaplane/blob/main/pkg/apis/numaplane/v1alpha1/status_types.go#L138)

pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
```diff
@@ -62,3 +62,8 @@ type MonoVertexRolloutList struct {
 func init() {
 	SchemeBuilder.Register(&MonoVertexRollout{}, &MonoVertexRolloutList{})
 }
+
+// IsHealthy indicates whether the MonoVertexRollout is healthy.
+func (mv *MonoVertexRolloutStatus) IsHealthy() bool {
+	return mv.Phase == PhaseDeployed
```
xdevxy, line 68: When the phase is `pending`, it is progressing which should also be healthy.

chandankumar4: Updated

juliev0 (approved):

xdevxy (approved):

PR #257, opened 2024-09-13
feat: Add health status metrics for pipeline, numaflow controller and isb service

chandankumar4:
Fixes #TODO


pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -201,3 +201,16 @@ func (s *Status) markTypeStatus(t ConditionType, status metav1.ConditionStatus,
...
+func (s *Status) IsReady() bool {
+	if len(s.Conditions) == 0 {
+		return false
+	}
+	for _, c := range s.Conditions {
+		if c.Status != metav1.ConditionTrue {
+			return false
+		}
+	}
+	return true
```
juliev0, line 215: as you noticed, this is not the case with `PipelinePausingOrPaused`

pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go
```diff
@@ -88,3 +88,11 @@ func (status *PipelineRolloutStatus) SetUpgradeInProgress(upgradeStrategy string
 func (status *PipelineRolloutStatus) ClearUpgradeInProgress() {
 	status.UpgradeInProgress = ""
 }
+
+// IsHealthy indicates whether the PipelineRollout is healthy.
+func (status *PipelineRolloutStatus) IsHealthy() bool {
+	if status.Phase != PhaseDeployed {
+		return false
+	}
+	return status.IsReady()
```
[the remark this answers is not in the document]
juliev0: (this comment applies for isbsvc and Numaflow Controller as well btw)

pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go
```diff
@@ -88,3 +88,8 @@ func (status *PipelineRolloutStatus) SetUpgradeInProgress(upgradeStrategy string
 func (status *PipelineRolloutStatus) ClearUpgradeInProgress() {
 	status.UpgradeInProgress = ""
 }
+
+// IsHealthy indicates whether the PipelineRollout is healthy.
+func (status *PipelineRolloutStatus) IsHealthy() bool {
+	return status.Phase == PhaseDeployed
```
juliev0, line 94: note that the way that the code is currently, we call it "Pending" if the user manually writes `desiredPhase=Paused` to their spec and the feature flag is turned on. Then this metric will deem it unhealthy. Perhaps need to reconsider if it should be "Pending" or "Deployed" in this case. 

The other effect of it being "Pending" is that it means ArgoCD would continually call it "Progressing" in the Jenkins pipeline. 

Let me write an issue for this.

juliev0: Actually, I take that back. I believe it should work but we should test it. Let me delete the comment from here, especially because I don't think it has an effect on the correctness of this.

pkg/apis/numaplane/v1alpha1/numaflowcontrollerrollout_types.go
```diff
@@ -65,3 +65,8 @@ type NumaflowControllerRolloutList struct {
 func init() {
 	SchemeBuilder.Register(&NumaflowControllerRollout{}, &NumaflowControllerRolloutList{})
 }
+
+// IsHealthy indicates whether the NumaflowController rollout is healthy or not
+func (nc *NumaflowControllerRolloutStatus) IsHealthy() bool {
```
xdevxy, line 70: Pending should also be healthy

pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go
xdevxy, line 94: similarly here

pkg/apis/numaplane/v1alpha1/isbservicerollout_types.go
```diff
@@ -66,3 +66,8 @@ type ISBServiceRolloutList struct {
 func init() {
 	SchemeBuilder.Register(&ISBServiceRollout{}, &ISBServiceRolloutList{})
 }
+
+// IsHealthy indicates whether the InterStepBufferService rollout is healthy or not
+func (isb *ISBServiceRolloutStatus) IsHealthy() bool {
+	return isb.Phase == PhaseDeployed
```
xdevxy, line 72: and here

chandankumar4: Updated in all places

internal/util/metrics/metrics.go
```diff
@@ -35,6 +39,10 @@ type CustomMetrics struct {
 	MonoVerticesSyncFailed *prometheus.CounterVec
 	// MonoVerticesSynced is the counter for the total number of monovertices synced.
 	MonoVerticesSynced *prometheus.CounterVec
+	// NumaflowControllersHealth is the gauge for the health of Numaflow controller.
+	NumaflowControllersHealth *prometheus.GaugeVec
+	// NumaflowControllerVersionCounter contains the information of all running numaflow controllers with "version" as a key and "name_namespace" as a value.
```
juliev0, line 44: there are a few comments which say "name_namespace" - should they?

chandankumar4: Ohh, It should not be. Might left the comment by mistake.
Fixed, Thanks

juliev0:
Looks like for some reason e2e test failed due to isbsvc updated and not healthy. Not related to your changes. Once we have debuggability working, we can better understand these. And hopefully Sidhant will fix some of the other issues in the meantime.

juliev0 (approved):

xdevxy (approved):

PR #260, opened 2024-09-13
test: add environment variable to enable dataLossPrevention for e2e testing

### LINKED ISSUE #259: e2e test - adding an envVar for testing with dataLossPrevention flag
# Summary

Currently we are only testing for basic CRUD in our e2e tests. In order to test with pausing enabled, we need an environment variable to allow us to install the specific pause manifests (dataLossPrevention flag is enabled) and run certain tests only when this flag is set.

# Use Cases

e2e testing

---

dpadhiar:
Fixes #259 

This pull request introduces the DATA_LOSS_PREVENTION environment variable. When this is set, the user will be able to:

- run make start to install the manifests that set the `dataLossPrevention` flag to `true`
- run make test-e2e to run specific cases to check that resources are pausing when being updated (PPND)

With this change, we are also able to split the currently long e2e test run on the Github CI into two separate runs using a matrix. These two runs are for when this dataLossPrevention flag is enabled or not to accurately test functionality.

### Modifications

Modify Makefile to use DATA_LOSS_PREVENTION flag to install pause manifests.

e2e tests now includes logic that is only checked when this environment variable is set.

Changed the update Pipeline e2e test to be a more substantial change (adding a vertex) for future testing later where we determine whether or not an upgrade truly requires a pause.

### Verification

Ran make start and make test-e2e with the variable set or not multiple times to check both scenarios.

### DIFF
--- a/tests/e2e/functional_test.go
+++ b/tests/e2e/functional_test.go
@@ -331,10 +331,16 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		if dataLossPrevention == "true" {
 			document("Verify that Pipelines are being paused by checking rollout condition")
-			Eventually(func() metav1.ConditionStatus {
-				rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
-				return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPausingPipelines)
-			}, testTimeout).Should(Equal(metav1.ConditionTrue))
+			Eventually(func() bool {
+				ncRollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+				ncCondStatus := getRolloutCondition(ncRollout.Status.Conditions, apiv1.ConditionPausingPipelines)
+				plRollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
+				plCondStatus := getRolloutCondition(plRollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
+				if ncCondStatus != metav1.ConditionTrue || plCondStatus != metav1.ConditionTrue {
+					return false
+				}
+				return true
+			}, testTimeout).Should(BeTrue())
 		}
 
 		// TODO: update this controller image when Numaflow v1.3.1 is released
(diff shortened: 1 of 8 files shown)


tests/e2e/functional_test.go
```diff
@@ -251,13 +291,13 @@ var _ = Describe("Functional e2e", Serial, func() {
...
-		// 	rollout, _ := pipelineRolloutClient.Get(ctx, pipelineRollouName, metav1.GetOptions{})
-		//
-		// return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
-		// }, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
+		if dataLossPrevention == "true" {
+			document("Verify that child Pipeline is paused by checking rollout condition")
+			Eventually(func() metav1.ConditionStatus {
+				rollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
+				return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
+			}, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
```
juliev0, line 299: maybe we should actually reduce the interval for this one, since the value will go from True to False, we need to make sure we catch it.

dpadhiar: Could we use the default 10ms for just these cases then?

tests/e2e/functional_test.go
```diff
@@ -289,6 +329,14 @@ var _ = Describe("Functional e2e", Serial, func() {
 			return rollout, nil
 		})
 
+		if dataLossPrevention == "true" {
+			document("Verify that Pipelines are being paused by checking rollout condition")
+			Eventually(func() metav1.ConditionStatus {
+				rollout, _ := numaflowControllerRolloutClient.Get(ctx, numaflowControllerRolloutName, metav1.GetOptions{})
+				return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPausingPipelines)
+			}, testTimeout, testPollingInterval).Should(Equal(metav1.ConditionTrue))
+		}
```
juliev0, line 338: do you want to confirm that this goes to "False" later? 

Also, in the same `Eventually` statement here, you could also check that the Pipeline's Condition is True here

tests/e2e/functional_test.go
juliev0: let's do that for any test that's checking that pausing=true

tests/e2e/functional_test.go
dpadhiar: We confirm its false later by calling `verifyNumaflowControllerReady` on L346.

For checking the Pipeline's condition, do you want to check the rollout or the child Pipeline?

juliev0: sorry, I just saw you did check that it goes to "False" later

dpadhiar:
```
if dataLossPrevention == "true" {
			document("Verify that Pipelines are being paused by checking rollout condition")
			Eventually(func() bool {
				isbRollout, _ := isbServiceRolloutClient.Get(ctx, isbServiceRolloutName, metav1.GetOptions{})
				isbStatus := getRolloutCondition(isbRollout.Status.Conditions, apiv1.ConditionPausingPipelines)
				plRollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
				plStatus := getRolloutCondition(plRollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
				if isbStatus != metav1.ConditionTrue || plStatus != metav1.ConditionTrue {
					return false
				}
				return true
			}, testTimeout).Should(BeTrue())
		}
```

dpadhiar: Just realized there is also the `verifyPipelineStatus` func that would probably be better for this. My only hesitation is that there is a call to document every time that function is run so we might be printing the state of the cluster many many times if this is called quite often. I could move that call out of that function and into `verifyPipelineReady` where it's being called.

Thoughts? @juliev0

juliev0: > For checking the Pipeline's condition, do you want to check the rollout or the child Pipeline?
Hmm, I'm okay either way.

juliev0: > Just realized there is also the verifyPipelineStatus func that would probably be better for this. My only hesitation is that there is a call to document every time that function is run so we might be printing the state of the cluster many many times if this is called quite often. I could move that call out of that function and into verifyPipelineReady where it's being called.

Good calls. Yeah, we could move the `document` call out of `verifyPipelineStatus()`

dpadhiar: This will mean we would wrap an Eventually inside another. Maybe it would be best we just check the status of the Pipeline before or after we check the rollout condition.

juliev0 (approved):

PR #265, opened 2024-09-14
fix: functional test for checking correct number of pipeline vertices running

juliev0:
### Modifications

Function in test that was supposed to be checking for a certain number of vertices running wasn't using the parameter passed in. Also, since the test switches from using 2 to 3 vertices, need to pass in 3 once the spec gets changed.

### DIFF
--- a/tests/e2e/functional_test.go
+++ b/tests/e2e/functional_test.go
@@ -358,7 +358,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		verifyNumaflowControllerReady(Namespace)
 
-		verifyPipelineReady(Namespace, pipelineRolloutName, 2)
+		verifyPipelineReady(Namespace, pipelineRolloutName, 3)
 
 	})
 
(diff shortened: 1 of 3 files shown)


tests/e2e/functional_test.go
```diff
@@ -358,7 +358,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		verifyNumaflowControllerReady(Namespace)
 
-		verifyPipelineReady(Namespace, pipelineRolloutName, 2)
+		verifyPipelineReady(Namespace, pipelineRolloutName, 3)
```
juliev0, line 361: at this point the PipelineRollout spec has been modified to 3

chandankumar4: E2E test should be failing without this test?

juliev0: no, first of all the parameter wasn't being used in the function (see other fix below). Plus, the `verifyPodsRunning()` function checks that at least this number are running. And at least 2 were running, so it passed

juliev0: okay, after this comment, decided to update from `>=` to `==`

chandankumar4 (approved):
LGTM

xdevxy (approved):

Issue #239, opened 2024-09-09
Determine how to handle ambiguous Child health conditions to determine if reconciliation is still happening (PPND)


juliev0:
**Describe the bug**
PipelineRollout, ISBServiceRollout, and NumaflowController Rollout need to know whether the child resources of the Pipeline and ISBSvc, and whether the Numaflow Controller Deployment are done Progressing so they can safely unpause pipelines.

Pipeline and ISBsvc have Conditions which can indicate if their children are progressing: `Condition.Reason=Progressing`. 

However, there are certain cases in which the Condition is set to false but it was ambiguous as to whether it was a case of "Progressing" or just general failure, so the Condition was not marked as "Progressing", namely I see this one:

```
{Type:DaemonServiceHealthy Status:False Reason:GetDaemonServiceFailed Message:Deployment not found, might be still under creation}
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

juliev0:
@chandankumar4 would you mind checking if there are other cases where we are saying some other reason than "Progressing", but it could be a progressing case?

juliev0:
Since some Conditions will inevitably be ambiguous, I think the way I will deal with it instead is this: once we are setting `inProgressStrategy` to "PPND", then I will concern myself with all unhealthy Child Conditions and not worry about their particular "Reason". I will make sure we're pausing if there's still unhealthiness during the process of PPND and not outside of it.

Issue #248, opened 2024-09-11
Integrate USDE logic into Rollout controllers to support PPND - perform selected strategy for isbsvc fields


juliev0:
# Summary

Integrate the USDE logic to derive an upgrade strategy into the reconciliation process of the ISBServiceRollout for the purpose of selectively handling different strategies for **isbsvc field changes.**

Relates to [this](https://github.com/numaproj/numaplane/issues/228)

Similar to PipelineRolloutController, the ISBServiceRolloutController should have an `InProgressManager`, so that it can maintain the current strategy that's in Progress (Progressive, PPND, or none) such that if 2 subsequent requests come in, or if the user's preferred strategy changes, it will finish what it's in the middle of doing first.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
@xdevxy we are currently not doing any field-specific checks for isbsvc at all, so that's why this issue is here

juliev0:
We want to make the isbservicerollout_controller more similar to [pipelinerollout_controller](https://github.com/numaproj/numaplane/blob/decision-engine/internal/controller/pipelinerollout_controller.go#L443-L529). 

The pseudocode should be:

```
`isbServiceNeedsToUpdate, upgradeStrategyType, err := usde.ResourceNeedsUpdating` (determine if we're trying to update the ISBService spec, and if so, is it a simple change that would be "direct apply" or is it a non-trivial change that would require PPND or Progressive?)

Is there currently an in-progress-strategy set for this ISBServiceRollout?
If not, do we need to set it?:
  If the user-prefered strategy is PPND && upgradeStrategyType==PPND: set in-progress-strategy to PPND
  If the user-prefered strategy is Progressive && upgradeStrategyType==Progressive: set in-progress-strategy to Progressive

Now do whatever the in-progress-strategy says to do
```

PR #270, opened 2024-09-16
feat: Upgrade Decision engine incorporated into reconciliation logic

### LINKED ISSUE #228: Integrate USDE logic into Rollout controllers to support PPND - perform selected strategy for Pipeline fields
# Summary

Integrate the USDE logic to derive an upgrade strategy into the reconciliation process of the PipelineRollout, ISBServiceRollout, and NumaflowControllerRollout for the purpose of selectively handling different strategies for Pipeline field changes.

Note [this](https://github.com/numaproj/numaplane/issues/248) similar issue but to selectively handle **isbsvc** fields

---

juliev0:
Fixes  #228  #239 

### Modifications

From PipelineRollout Reconciler, if user has set a preference for PPND, we determine which strategy we need to do based on 1. whether/which fields changed, 2. was there an external request to pause from NumaflowController or ISBService? If so, we put ourselves in the PPND InProgressStrategy and set that value in memory and in the Rollout Status. 

From NumaflowControllerRO Reconciler and ISBServiceRO reconciler, if user has set a preference for PPND, we perform the PPND strategy always (later we will make it dependent on which isbsvc fields changed). 

### Verification

Added unit test for PPND to PipelineRollout, testing many different cases, including whether the InProgressStrategy is set correctly. 

Enhanced e2e test to include verification of InProgressStrategy, and also now includes verification of the cases of:
1. user sets `desiredPhase=Paused`
2. user sets `desiredPhase=Running` when it was previously paused

### To Do
1. perhaps e2e test could be updated to include an actual user config
2. address any TODOs in the code
3. Would like to move PPND specific functions into a separate file


internal/controller/config/usde_config.go
```diff
@@ -1,17 +1,5 @@
 package config
 
-import (
-	"encoding/json"
-	"fmt"
-)
-
-type USDEUserStrategy string
```
juliev0, line 8: moved the stuff from this file that pertains to the user config to its own file

internal/usde/usde.go
```diff
@@ -53,51 +35,86 @@ func DeriveUpgradeStrategy(ctx context.Context, newSpec *kubernetes.GenericObjec
...
 	if err != nil {
-		return UpgradeStrategyError, nil, err
+		return false, apiv1.UpgradeStrategyError, err
 	}
 
-	numaLogger.WithValues("newSpecOnlyApplyPaths", newSpecOnlyApplyPaths, "newSpecWithoutApplyPaths", newSpecWithoutApplyPaths).Debug("split new spec")
+	numaLogger.WithValues(
+		"newSpecOnlyApplyPaths", newSpecOnlyApplyPaths,
+		"newSpecWithoutApplyPaths", newSpecWithoutApplyPaths,
+	).Debug("split new spec")
```
juliev0, line 46: most of the code here is actually from Tony's other branch

internal/util/util.go
```diff
@@ -85,18 +85,18 @@ func removeNullValuesFromMap(m map[string]any) {
 
 // SplitObject returns 2 maps from a given object as bytes array and a slice of paths.
 // One of the 2 output maps will include only the paths from the slice while the second returned map will include all other paths.
-func SplitObject(obj []byte, paths []string, pathSeparator string) (map[string]any, map[string]any, error) {
+func SplitObject(obj []byte, paths []string, excludedPaths []string, pathSeparator string) (map[string]any, map[string]any, error) {
```
juliev0, line 88: the code in this file is from Tony's other branch

tests/e2e/functional_test.go
```diff
@@ -260,23 +264,30 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		// update child Pipeline
 		updatePipelineSpecInK8S(Namespace, pipelineRolloutName, func(pipelineSpec numaflowv1.PipelineSpec) (numaflowv1.PipelineSpec, error) {
-			rpu := int64(10)
-			pipelineSpec.Vertices[0].Source.Generator.RPU = &rpu
+			pipelineSpec.Watermark.Disabled = true
```
juliev0, line 267: currently this is one of the fields that causes a DirectApply from USDE

tests/e2e/functional_test.go
```diff
@@ -306,16 +313,71 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+		verifyPipelineRolloutDeployed(pipelineRolloutName)
+		verifyPipelineRolloutHealthy(pipelineRolloutName)
+
+		verifyPipelineRunning(Namespace, pipelineRolloutName, 3)
+
+	})
+
+	currentPipelineSpec = updatedPipelineSpec
+
+	It("Should pause the Pipeline if user requests it", func() {
```
juliev0, line 329: added tests for user setting desiredPhase

config/install.yaml
```diff
@@ -1398,6 +1398,7 @@ apiVersion: v1
 data:
   config.yaml: |-
     logLevel: 3
+    defaultUpgradeStrategy: ""
```
juliev0, line 1401: this replaces the `dataLossPrevention` feature flag, and has 3 possible states: 

- "" means no op
- "pause-and-drain"
- "progressive"

internal/controller/config/config.go
```diff
@@ -69,8 +69,9 @@ func (*ConfigManager) GetControllerDefinitionsMgr() *NumaflowControllerDefinitio
 type GlobalConfig struct {
 	LogLevel          int    `json:"logLevel" mapstructure:"logLevel"`
 	IncludedResources string `json:"includedResources" mapstructure:"includedResources"`
-	// Feature flag - if enabled causes pipeline(s) to be paused when pipeline, numaflow controller, or ISB Service gets updated
-	DataLossPrevention bool `json:"dataLossPrevention" mapstructure:"dataLossPrevention"`
+	// If user's config doesn't exist or doesn't specify strategy, this is the default
+	// TODO: should we put this here or in usde config?
+	DefaultUpgradeStrategy USDEUserStrategy `json:"defaultUpgradeStrategy" mapstructure:"defaultUpgradeStrategy"`
```
juliev0, line 74: not sure if we want this to be in the USDE config instead

internal/controller/pause.go
```diff
@@ -115,12 +114,27 @@ func (pm *PauseModule) updatePipelineLifecycle(ctx context.Context, restConfig *
...
+	if resultObj == nil {
+		return fmt.Errorf("error converting unstructured %+v to object, result is nil?", unstruc.Object)
+	}
+
+	err = kubernetes.UpdateCR(ctx, restConfig, resultObj, "pipelines")
+	if err != nil {
+		return err
+	}
+	*pipeline = *resultObj
+	return nil
```
juliev0, line 136: By calling `UpdateCR()`, it will update `resultObj` to the returned value from K8S and we can then set `pipeline` to it, meaning any subsequent update call that might get made will get the latest value and won't incur a resource version conflict.

internal/controller/pause.go
```diff
@@ -109,18 +108,42 @@ func (pm *PauseModule) runPipelineIfSafe(ctx context.Context, restConfig *rest.C
...
+	err := withDesiredPhase(pipeline, phase)
+	if err != nil {
+		return err
+	}
+
+	err = kubernetes.UpdateCR(ctx, restConfig, pipeline, "pipelines")
+	if err != nil {
+		return err
+	}
+	return nil
```
juliev0, line 122: by calling `UpdateCR()` here, we get the latest `pipeline` with the revised ResourceVersion from K8S, so any subsequent call to do an update within this reconcile call will use the latest ResourceVersion and won't result in a conflict error.

internal/controller/pipelinerollout_controller.go
```diff
@@ -408,49 +433,105 @@ func mergePipeline(existingPipeline *kubernetes.GenericObject, newPipeline *kube
...
+	// is there currently an inProgressStrategy for the pipeline? (This will override any new decision)
+	inProgressStrategy := r.inProgressStrategyMgr.getStrategy(ctx, pipelineRollout)
+
+	// if not, should we set one?
+	// inProgressStrategy is used for PPND and Progressive strategies (i.e. any strategies which require multiple reconciliations to perform)
+	if userPreferredStrategy == config.PPNDStrategyID && inProgressStrategy != apiv1.UpgradeStrategyPPND {
+		// if the preferred strategy is PPND, do we need to start the process for PPND (if we haven't already)?
+		needPPND := false
+		if userPreferredStrategy == config.PPNDStrategyID {
+			ppndRequired, err := r.needPPND(ctx, pipelineRollout, newPipelineDef, upgradeStrategyType == apiv1.UpgradeStrategyPPND)
```
xdevxy, line 473: We should not alter the runtime in progress upgrade strategy right? As that implies there is an upgrade with a certain strategy already going on.

internal/controller/in_progress_strategy.go
```diff
@@ -0,0 +1,103 @@
...
+		inProgressUpgradeStrategies: map[string]apiv1.UpgradeStrategy{},
+	}
+}
+
+func (mgr *inProgressStrategyMgr) getStrategy(ctx context.Context, rollout client.Object) apiv1.UpgradeStrategy {
+	return mgr.synchronize(ctx, rollout)
+}
+
+// make sure in-memory value and Rollout Status value are syncronized
+// if in-memory value is set, make sure Rollout Status value gets set the same
```
xdevxy, line 51: nit: synchronized

internal/controller/isbservicerollout_controller.go
```diff
@@ -300,13 +310,19 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 			r.customMetrics.ReconciliationDuration.WithLabelValues(ControllerISBSVCRollout, "update").Observe(time.Since(syncStartTime).Seconds())
 			return nil
 		})
-	} else if isbServiceNeedsUpdating {
-		// update ISBService
-		err = r.updateISBService(ctx, isbServiceRollout, newISBServiceDef)
-		if err != nil {
-			return false, err
+	case config.NoStrategyID:
```
xdevxy, line 313: Hmm, we also needs to track for ISBService if there is an ongoing upgrade happening now, right?

juliev0: yes. But in the current code, we are doing PPND for all updates if that's the user strategy for the namespace. Once we incorporate the logic that looks at the individual fields for isbsvc to do apply vs PPND/Progressive, then we need InProgressStrategy here too.

internal/controller/pipelinerollout_controller.go
juliev0: I guess you are referring to the case in which a user changes their preferred strategy on their namespace while something is already happening. You're right - I didn't take care of that case. 

Perhaps I can just change like 469 above to:
`if userPreferredStrategy == config.PPNDStrategyID && inProgressStrategy == apiv1.UpgradeStrategyNoOp`

internal/controller/config/config.go
juliev0: sure. In the future, I think we should try to put most everything in our `numaplane-controller-config` - do you know why we have a separate file?

tests/e2e/functional_test.go
```diff
@@ -306,18 +315,79 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+		// update the PipelineRollout
+		updatePipelineRolloutInK8S(Namespace, pipelineRolloutName, func(rollout apiv1.PipelineRollout) (apiv1.PipelineRollout, error) {
+			rollout.Spec.Pipeline.Spec.Raw = rawSpec
+			return rollout, nil
+		})
+		document("verifying PipelineRollout spec deployed")
+		verifyPipelineRolloutDeployed(pipelineRolloutName)
+
+		// Give it a little while to get to Paused and then verify that it stays that way
+		// TODO: add back after Numaflow fixes this to not go from Paused to Pausing
```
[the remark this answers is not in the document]
juliev0: Sidhant actually fixed it [here](https://github.com/numaproj/numaflow/pull/2065) - I just need for us to incorporate it

internal/controller/isbservicerollout_controller.go
juliev0: I see - maybe in this case you are also thinking about when the user changes from one strategy to another in their configmap? Is it okay if we include this as part of the work Dillen will do for isbsvc?

juliev0: I have updated the description in that [issue](https://github.com/numaproj/numaplane/issues/248) to make sure this is included. I will also add a comment for that issue for what the pseudocode should be.

xdevxy: sounds good to me. Thanks!

internal/controller/config/config.go
xdevxy: this is not for the namespaced user upgrade strategy config but cluster wise for USDE config, right? then I'm not aware of reasons it has to be a separate file.

juliev0: right. In the future, we can try to just have the one file for stuff like this unless there's a good reason why it should be separate

internal/usde/usde.go
```diff
@@ -53,51 +35,80 @@ func DeriveUpgradeStrategy(ctx context.Context, newSpec *kubernetes.GenericObjec
...
+			"existingSpecWithoutApplyPaths", existingSpecWithoutApplyPaths,
+		).Debug("the specs without the 'apply' paths are different")
+
+		switch userUpgradeStrategy {
+		case config.PPNDStrategyID:
+			return true, apiv1.UpgradeStrategyPPND, nil
+		case config.ProgressiveStrategyID:
+			return true, apiv1.UpgradeStrategyProgressive, nil
+		case config.NoStrategyID:
+			return true, apiv1.UpgradeStrategyApply, nil
```
xdevxy, line 78: This block of code according to the comment at L59 should only return `PPND or Progressive`?

xdevxy (approved):

internal/controller/pipelinerollout_controller.go
```diff
@@ -408,49 +433,108 @@ func mergePipeline(existingPipeline *kubernetes.GenericObject, newPipeline *kube
...
 
-	// Get the fields we need from both the Pipeline spec we have and the one we want
-	// TODO: consider having one struct which include our GenericObject plus our PipelineSpec so we can avoid multiple repeat conversions
+	numaLogger := logger.FromContext(ctx)
 
-	var existingPipelineSpec PipelineSpec
-	if err := json.Unmarshal(existingPipelineDef.Spec.Raw, &existingPipelineSpec); err != nil {
-		return fmt.Errorf("failed to convert existing Pipeline spec %q into PipelineSpec type, err=%v", string(existingPipelineDef.Spec.Raw), err)
+	// what is the preferred strategy for this namespace?
+	userPreferredStrategy, err := usde.GetUserStrategy(ctx, newPipelineDef.Namespace)
```
xdevxy, line 443: I feel the logic from L443 to L528 to determine which strategy to be used is too interleaved between code blocks and be further condensed. But I'm good to leave it to a follow up.

internal/usde/usde.go
juliev0: Good catch that that's a mismatch. I am going to change the comment. Perhaps in the future, those would be the only strategies allowed, but for now "no strategy" is also allowed.

internal/controller/pipelinerollout_controller.go
juliev0: can you clarify what you mean by "interleaved between code blocks"? 

I specifically wanted to design the overall flow as:
1. figure out what "In progress strategy" should be and set it if it's not set
2. Do whatever the "In progress strategy" is

Do you disagree with that approach?

juliev0: Feel free to refactor if you see a potential improvement

juliev0: Okay, I thought about this a little more, and I can kind of see what you're saying. I think I can see keep the PPND logic collocated, and keep the Progressive logic collocated.

juliev0: On second thought, I started to do this and it didn't seem to be better with what I was doing so I decided not to spend time on it. But do feel free to refactor it if you want to. :)

PR #275, opened 2024-09-17
test: enable debuggability for e2e tests

### LINKED ISSUE #266: Enable debuggability for CI e2e failures
# Summary

If we see the e2e test fail in the CI, currently there is very little information for what happened. It would be good if possible to have the numaplane log and the numaflow log (potential to add more later).

[This](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow) article describes publishing artifact in a Github Workflow. I would like to see the numaplane log and the numaflow log both published there. I think it would be fine to either just publish it for every run, or just for failed runs. (We should configure a [retention period](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#configuring-a-custom-artifact-retention-period) of say 7 days.)

The other aspect of this is: the e2e test needs to be able to fail fast if any test fails. This will enable us to have the numaflow log before the numaflow controller is deleted. Can we do this somehow through Ginkgo framework?

---

dpadhiar:
Fixes #266 

### Modifications

If an e2e test fails on the CI currently there is no easy way to check the logs of the controllers to see what the issue might be and begin debugging.

This PR enables this option by outputting controller logs for both Numaplane and Numaflow after unsuccessful runs and Numaplane only after successful runs (the Numaflow controller is deleted in a normal test). These logs are available for download on the Github CI as an artifact and can be found locally under `tests/e2e/output` where the output directory is created during a test run if not already existing. Currently these artifacts will persist for 7 days.

Testing has also been updated so that after a test spec fails, the test ends immediately as opposed to trying to complete the rest. This will save time by not having to wait for the full test to run if an early test fails and we can preserve Numaflow controller logs.

### Verification

Running make test-e2e locally with a test set incorrectly to test failures and see that logs are being output.

Submitted a commit with a failing test to check that both sets of artifacts are downloadable.

Check artifacts here: https://github.com/numaproj/numaplane/pull/275/commits/6c7576e6df635b6c64a929c3d598bbc1e139f929

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -3,6 +3,8 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"io"
+	"os"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -19,6 +21,7 @@ import (
 	"github.com/numaproj/numaplane/internal/util/kubernetes"
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
 	planepkg "github.com/numaproj/numaplane/pkg/client/clientset/versioned/typed/numaplane/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 var (
@@ -41,6 +44,12 @@ var (
 
 const (
 	Namespace = "numaplane-system"
+
+	NumaplaneCtrlLogs = "output/numaplane-controller.log"
+	NumaflowCtrlLogs  = "output/numaflow-controller.log"
+
+	NumaplaneLabel = "app.kubernetes.io/part-of=numaplane"
+	NumaflowLabel  = "app.kubernetes.io/part-of=numaflow, app.kubernetes.io/component=controller-manager"
 )
 
 // document for Ginkgo framework and print to console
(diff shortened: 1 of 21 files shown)


tests/e2e/common.go
```diff
@@ -107,3 +116,33 @@ func printPodLogs(client clientgo.Interface, namespace, podName, containerName s
...
 }*/
+
+func getPodLogs(client clientgo.Interface, namespace, labelSelector, containerName, fileName string) {
+
+	ctx := context.Background()
+	podLogOptions := &corev1.PodLogOptions{Container: containerName}
+
+	podList, err := client.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	if err != nil {
+		// TODO: improve error reporting
```
juliev0, line 127: Do you want to just do this now instead of needing to have a TODO? Maybe you could just return an error? or did you have something else in mind?

tests/e2e/common.go
```diff
@@ -107,3 +116,33 @@ func printPodLogs(client clientgo.Interface, namespace, podName, containerName s
 	logBytes, _ := io.ReadAll(stream)
 	fmt.Printf("Printing Log for namespace=%q, pod=%q, container=%q:\n%s\n", namespace, podName, containerName, string(logBytes))
 }*/
+
```
juliev0, line 119: can you delete my commented out function above? (now that you are essentially doing almost the same thing)

tests/e2e/common.go
dpadhiar: Ack, I just left this as a reminder to add in the PR while I was just testing things out quickly. I think we could just return the error as well. We just want to be sure we get to stopping the test env or aborting the suite.

juliev0 (approved):

xdevxy (approved):

PR #277, opened 2024-09-17
fix: bug with "Pipeline Status" icons

calvinx408:
1. looking at array[0] was wrong logic. need to loop through array of object 
image


calvinx408:
opening new after rebase

PR #281, opened 2024-09-19
test: enhance e2e debuggability by capturing resource changes and storing them as artifacts

### LINKED ISSUE #276: e2e debuggability - capture resource changes and store artifacts
# Summary

Another artifact if the e2e fails that would be useful is to understand how the Resources change over time (spec, status, etc). This would essentially be a "watch" (which shows each time the resource changes) plus a timestamp for each change. 

I'm debating which Resources to ask for. We have a lot of potential Resources:

- 3 different types of Rollout
- Numaflow resources: Pipeline, isbsvc, Vertex
- StatefulSet, Deployment, Pod

Of these, I'd say the most important ones are the first two rows. Maybe we could do those, or if you want to start small, just do Pipeline. We can either have a single file for each type, or for each resource if that's easier.

---

dpadhiar:
Fixes #276

### Modifications

To add to the controller artifacts that we currently have, it is also useful to track information of all the resources and their child resources. The best way to do this would be to store the output of `kubectl get <resource> -o yaml` anytime there is a change. We can achieve this by using a `watch.Interface` to watch each resource and generate an output file `<resource>.yaml` that continually receives the updated resource YAML every time a change occurs.

### Verification

Ran make test-e2e locally to verify resource file are output correctly.

Verified artifacts are uploaded when test fails on Github CI.

Failed commit: https://github.com/numaproj/numaplane/pull/281/commits/218e6b4300be90cc53621e77563bd8de7079f2d9


juliev0:
Looks like we watch the entire resource Kind for each file generated. That seems fine. In the future, perhaps we will generate multiple files per resource name. But right now we don't have that need, since for these resources at least we only have one of each.

juliev0 (approved):

dpadhiar:
> Looks like we watch the entire resource Kind for each file generated. That seems fine. In the future, perhaps we will generate multiple files per resource name. But right now we don't have that need, since for these resources at least we only have one of each.

Sounds good, this would be a pretty simple change when we need it: just adding the resource name as a parameter and using it in the label selector.

PR #286, opened 2024-09-20
feat: Add support for multiple pipelines can coexist with a PipelineRollout 

### LINKED ISSUE #243: Support multiple pipelines can coexist for a PipelineRollout

xdevxy:
Fixes #243 

### Modifications

Add name count property to the PipelineRollout Status to be used as the name count suffix for the pipeline names, so that multiple pipelines can coexist for a PipelineRollout.

### Verification

`make test`
`make test-e2e`


internal/controller/pipelinerollout_controller_test.go
```diff
@@ -947,17 +954,19 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
...
 			assert.Len(t, pipelineList.Items, 0)
 
-			// create Pipeline definition
 			rollout := createPipelineRollout(tc.newPipelineSpec)
+			_ = numaplaneClient.Delete(ctx, rollout)
+
 			rollout.Status.Phase = tc.initialRolloutPhase
 			if tc.initialInProgressStrategy != nil {
+				rollout.Status.UpgradeInProgress = *tc.initialInProgressStrategy
 				rollout.Status.UpgradeInProgress = *tc.initialInProgressStrategy
```
juliev0, line 970: is this a duplicate line?

internal/controller/pipelinerollout_controller_test.go
```diff
@@ -968,6 +977,9 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
 			// the Reconcile() function does this, so we need to do it before calling reconcile() as well
 			rollout.Status.Init(rollout.Generation)
 
+			err = numaplaneClient.Create(ctx, rollout)
+			assert.NoError(t, err)
```
juliev0, line 981: This is interesting. So, it didn't work if the PipelineRollout didn't already exist on the cluster?

(I mean, I suppose in real life, the PipelineRollout _would_ already exist if the Pipeline already exists)

xdevxy: yeah, if not exist the `update()` will not work as it is not created yet.

internal/controller/pipelinerollout_controller_test.go
xdevxy: ah yeah! Thanks for catching, must be from the rebase.

juliev0:
let me take a look at the e2e test failing....

juliev0:
> let me take a look at the e2e test failing....

Okay, I actually saw this yesterday as well. Remember when I mentioned that the tricky thing about checking for pausing in the e2e test is that we need to catch it before it goes back to "Running"? We need to revisit that I guess.

juliev0:
> > let me take a look at the e2e test failing....
> 
> Okay, I actually saw this yesterday as well. Remember when I mentioned that the tricky thing about checking for pausing in the e2e test is that we need to catch it before it goes back to "Running"? We need to revisit that I guess.

Sorry, actually, this seems to be different. This is the case in which we explicitly set `desiredPhase=Paused` in the Rollout, so it doesn't have the issue I mentioned above (because it only goes back to `desiredPhase=Running` after the test explicitly sets it back).

In this case, if I download the artifacts I see that we keep `desiredPhase=Paused` but Numaflow is taking the Status from "Paused" into "Pausing", which sounds like the [fix](https://github.com/numaproj/numaflow/pull/2065) that Sidhant made so let's see if this happens after I incorporate the newest Numaflow version in (which I am working on now).

juliev0:
Actually, I just revisited the other intermittent failure I saw in the CI. I think that one also may be fixed by Sidhant's fix, as I can see the Pipeline going from "Paused" to "Pausing".

PR #287, opened 2024-09-23
feat: incorporate latest numaflow, and incorporate various fixes and features as a result of that 

juliev0:
### Modifications

Incorporate the latest numaflow in, including:

- latest RC versions of numaflow that were created, 0.0.18 and 0.0.19
- version of numaflow used in go.mod, which I'm using in the e2e test to reference `PipelineSpec`/`PipelineStatus`

Having latest numaflow PipelineStatus means I can check for [drainedOnPause](https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/pipeline_types.go#L639) to confirm that there was nothing that failed as far as pausing.

### Verification

Ran e2e test multiple times

### To do

Now that I have the latest Numaflow version, which no longer does an Update on Pipeline, I will try removing the logic which eliminates the zero values from the comparison.


internal/controller/pipelinerollout_controller.go
```diff
@@ -468,7 +468,7 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 	// if not, should we set one?
 	if !inProgressStrategySet {
 		// inProgressStrategy is used for PPND and Progressive strategies (i.e. any strategies which require multiple reconciliations to perform)
-		if userPreferredStrategy == config.PPNDStrategyID && inProgressStrategy != apiv1.UpgradeStrategyPPND {
+		if userPreferredStrategy == config.PPNDStrategyID {
```
juliev0, line 471: realized these parts of the "if statement" were no longer needed now that this is enclosed within a check that `inProgressStrategy` isn't set at all

internal/controller/suite_test.go
```diff
@@ -86,7 +86,7 @@ var _ = BeforeSuite(func() {
 		"https://raw.githubusercontent.com/numaproj/numaflow/main/config/base/crds/minimal/numaflow.numaproj.io_interstepbufferservices.yaml",
 		"https://raw.githubusercontent.com/numaproj/numaflow/main/config/base/crds/minimal/numaflow.numaproj.io_pipelines.yaml",
 		"https://raw.githubusercontent.com/numaproj/numaflow/main/config/base/crds/minimal/numaflow.numaproj.io_vertices.yaml",
-		"https://raw.githubusercontent.com/numaproj/numaflow/main/config/base/crds/full/numaflow.numaproj.io_monovertices.yaml",
+		"https://raw.githubusercontent.com/numaproj/numaflow/main/config/base/crds/minimal/numaflow.numaproj.io_monovertices.yaml",
```
juliev0, line 89: unit test for monovertex was failing due to this

tests/e2e/functional_test.go
```diff
@@ -291,8 +290,8 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		if dataLossPrevention == "true" {
 			document("Verify that child Pipeline is not paused when an update not requiring pause is made")
-			verifyPipelineStatusConsistently(Namespace, pipelineRolloutName, func(retrievedPipelineSpec numaflowv1.PipelineSpec, retrievedPipelineStatus kubernetes.GenericStatus) bool {
-				return retrievedPipelineStatus.Phase != string(numaflowv1.PipelinePhasePaused)
+			verifyPipelineStatusConsistently(Namespace, pipelineRolloutName, func(retrievedPipelineSpec numaflowv1.PipelineSpec, retrievedPipelineStatus numaflowv1.PipelineStatus) bool {
```
juliev0, line 293: if we use the `PipelineStatus` instead of `GenericStatus`, we can check for Pipeline Status-specific fields

tests/manifests/default/kustomization.yaml
```diff
@@ -3,6 +3,8 @@ kind: Kustomization
 
 resources:
   - ../../../config/default
+  - ./controller_def_rc0.0.18.yaml
+  - ./controller_def_rc0.0.19.yaml
```
juliev0, line 7: it's easier to create these controller definitions that are single files rather than have one big long file of all of the definitions

I think in the future it might be nice if we can just copy the entire manifest from the open source file and not even worry about needing to remove anything by hand. (I noticed that when I did this, the `Service` Type wasn't an [includedResource](https://github.com/numaproj/numaplane/blob/main/config/install.yaml#L1402) so that failed, but maybe we can change that?)

xdevxy (approved):

PR #294, opened 2024-09-25
feat: Remove logic which eliminates zero values from spec comparison

### LINKED ISSUE #246: Remove nil checking from pipeline comparison after Numaflow changes its Finalizer updating to patch
**Describe the bug**
We currently have logic which checks if the Pipeline or ISBSvc has changed in order to know what to do. When Numaflow Controller adds the finalizer it can add additional fields. This could result in an incorrect comparison result.

Document here: https://docs.google.com/document/d/1BfaDP4mf5QxBhttNP32p2X09YyhbS-9KsuyjR_Qt8c8/edit#heading=h.9032mlsdcybz

Numaflow will resolve this by patching the finalizer rather than updating the Pipeline directly. Once it does, we can remove all of this nil checking.

---

juliev0:
Fixes #246 

### Modifications

Now that we have incorporated the latest version of Numaflow, which no longer performs updates on Pipelines (now just Patches), we no longer need to include the logic which tries to eliminate the zero values that were being added from our comparisons.

There are also a couple of other changes described inline.

### Verification

- Ran e2e test of course
- Updated USDE unit test to include more cases of spec changes.
- Did some manual changes of updating pipeline, including those which involved different combinations of: {desiredPhase=Paused/Running/unset, ppnd-required field, apply-field} since I also made a change in this area of the code


internal/controller/pipelinerollout_controller.go
```diff
@@ -661,30 +662,35 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // pipelineSpecNeedsUpdating() tests for essential equality, with any irrelevant fields eliminated from the comparison
 func pipelineSpecNeedsUpdating(ctx context.Context, a *kubernetes.GenericObject, b *kubernetes.GenericObject) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
-	// remove lifecycle field from comparison, as well as any nulls to test for equality
-	pipelineWithoutLifecycleA, err := pipelineWithoutLifecycle(a)
+	// remove lifecycle.desiredPhase field from comparison to test for equality
+	pipelineWithoutDesiredPhaseA, err := pipelineWithoutDesiredPhase(a)
```
juliev0, line 666: this is just a name change - really we were only excluding "desiredPhase" as opposed to the entire "lifecycle" struct

internal/controller/pipelinerollout_controller.go
```diff
@@ -499,7 +500,7 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 	switch inProgressStrategy {
 	case apiv1.UpgradeStrategyPPND:
 		numaLogger.Debug("processing pipeline with PPND")
-		done, err := r.processExistingPipelineWithPPND(ctx, pipelineRollout, existingPipelineDef, newPipelineDef, pipelineNeedsToUpdate)
+		done, err := r.processExistingPipelineWithPPND(ctx, pipelineRollout, existingPipelineDef, newPipelineDef)
```
juliev0, line 503: this is calculated within the function instead - the calculation within the function eliminates the "lifecycle" field from the comparison (more information below in the function)

internal/controller/pipelinerollout_ppnd.go
```diff
@@ -24,11 +24,17 @@ import (
...
+	existingPipelineDef, newPipelineDef *kubernetes.GenericObject) (bool, error) {
 
 	numaLogger := logger.FromContext(ctx)
 
-	needsPaused, err := r.shouldBePaused(ctx, pipelineRollout, existingPipelineDef, newPipelineDef)
+	// this will check for any update required except for "desiredPhase", which PPND
+	pipelineNeedsToUpdate, err := pipelineSpecNeedsUpdating(ctx, existingPipelineDef, newPipelineDef)
+	if err != nil {
+		return false, err
+	}
```
juliev0, line 35: we calculate this here instead of using the value that was calculated by the USDE earlier - this eliminates the "lifecycle" field from the comparison (it was still mostly working the other way but the log was a little confusing to read, and we were making an unnecessary change to override `desiredPhase: Running` with ``).

internal/usde/usde_test.go
```diff
@@ -242,6 +260,52 @@ func Test_ResourceNeedsUpdating(t *testing.T) {
...
+			newSpec:      pipelineDefn,
+			existingSpec: pipelineDefn,
+			usdeConfig: config.USDEConfig{
+				DefaultUpgradeStrategy:    config.PPNDStrategyID,
+				PipelineSpecExcludedPaths: []string{"vertices.source.something"},
+			},
+			namespaceConfig:       &config.NamespaceConfig{UpgradeStrategy: "pause-and-drain"},
+			expectedNeedsUpdating: false,
+			expectedStrategy:      apiv1.UpgradeStrategyNoOp,
+		},
```
juliev0, line 308: Added some more tests over here to verify different kinds of changes in spec

config/install.yaml
```diff
@@ -900,7 +900,7 @@ spec:
     - jsonPath: .status.phase
       name: Phase
       type: string
-    - jsonPath: .spec.replicas
+    - jsonPath: .status.desiredReplicas
       name: Desired
```
juliev0, line 904: not related to my change but codegen was failing - I guess numaflow changed its CRD

internal/util/util_test.go
```diff
@@ -200,10 +198,10 @@ func Test_SplitMap(t *testing.T) {
 				},
 				"name": "John",
 				"projects": []msa{
-					{"name": "Project2", "other": []msa{{"y": "y2"}}, "status": "completed", "vals": []any{1, 2, 3}},
+					{"name": "Project2", "other": []msa{{}, {"y": "y2"}}, "status": "completed", "vals": []any{1, 2, 3}},
 					{"name": "Project3", "other": []msa{{"y": "y3"}}},
-					{"name": "Project4"},
-					{"name": "Project1", "other": []msa{{"y": "y1"}}, "status": "in progress", "vals": []any{4, 5, 6, 7}},
+					{"name": "Project4", "other": []msa{{}}},
+					{"name": "Project1", "other": []msa{{}, {"y": "y1"}, {}, {}}, "status": "in progress", "vals": []any{4, 5, 6, 7}},
```
juliev0, line 204: At first, I thought the `SplitMap()` function should not return these empty maps in an array like that, but it's actually better that they do - it essentially maintains all of the array entries in here, which enables the caller to correctly get the diff between the 2 arrays. (if some entries were missing, the diff would not be correct)

tests/e2e/functional_test.go
```diff
@@ -374,19 +374,18 @@ var _ = Describe("Functional e2e", Serial, func() {
...
 		verifyPipelineRolloutDeployed(pipelineRolloutName)
 
-		// Give it a little while to get to Paused and then verify that it stays that way
-
+		// Give it a little while to get to Paused and then verify that it stays in Paused (or otherwise Pausing)
 		verifyPipelinePaused(Namespace, pipelineRolloutName, pipelineName)
-		// TODO: add back after Numaflow fixes this to not go from Paused to Pausing
-		//document("verifying Pipeline stays paused")
-		/*Consistently(func() bool {
+		document("verifying Pipeline stays in paused or otherwise pausing")
```
juliev0, line 379: I've noticed that it goes to "Pausing" - @kohlisid - is this understood that this may happen if we simply set `desiredPhase=Paused` and then don't make any other spec change?

tests/e2e/functional_test.go
```diff
@@ -374,19 +374,18 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+			_, _, retrievedPipelineStatus, err := getPipelineFromK8S(Namespace, pipelineName)
 			if err != nil {
-				return false
+				return false // TODO: if this is a transient error, return true; otherwise false?
 			}
-			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused) == metav1.ConditionTrue && retrievedPipelineStatus.Phase == string(numaflowv1.PipelinePhasePaused)
-		}, 1*time.Minute, testPollingInterval).Should(BeTrue())*/
+			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused) == metav1.ConditionTrue &&
+				(retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePaused || retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePausing)
+		}, 1*time.Minute, testPollingInterval).Should(BeTrue())
```
juliev0, line 388: decided to put this test back in, except allow it to also be in "Pausing" (mainly we don't want it to be in "Running" or "Failed")

tests/e2e/functional_test.go
juliev0: (note that this is after we confirmed it went to "Paused", it switched back to "Pausing")

internal/controller/pipelinerollout_controller.go
```diff
@@ -661,30 +662,35 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		return nil, err
 	}
+	// remove "lifecycle.desiredPhase"
 	comparisonExcludedPaths := []string{"lifecycle.desiredPhase"}
 	util.RemovePaths(pipelineAsMap, comparisonExcludedPaths, ".")
+	// if "lifecycle" is there and empty, remove it
+	lifecycleMap, found := pipelineAsMap["lifecycle"].(map[string]interface{})
+	if found && len(lifecycleMap) == 0 {
+		util.RemovePaths(pipelineAsMap, []string{"lifecycle"}, ".")
+	}
```
juliev0, line 693: now that we no longer remove zero values in the 2 specs being compared, in order to effectively eliminate the `desiredPhase` field from the comparison, we also need to remove `lifecycle` parent if it has no other children, so:
- `"lifecycle{desiredPhase: x}" == "lifecycle{}" == ""`

xdevxy (approved):

internal/controller/isbservicerollout_controller.go
```diff
@@ -394,7 +395,7 @@ func (r *ISBServiceRolloutReconciler) isISBServiceUpdating(ctx context.Context,
 		return false, false, err
 	}
 	// TODO: see if we can just use DeepEqual or if there may be null values we need to ignore
```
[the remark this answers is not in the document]
juliev0: good call, will update

PR #297, opened 2024-09-25
test: e2e debuggability improvements

dpadhiar:
Follow up for #281 

### Modifications

This PR includes a few improvements for the current e2e debuggability which are:

- outputting pod changes for each resource and sorting all output into separate folders
- adding the option to set the env variable DISABLE_TEST_ARTIFACTS to disable watches for all resources and controllers
- refining the output to include fractional seconds and format better overall

### Verification

Ran make test-e2e locally to view the new output artifacts.


tests/e2e/suite_test.go
```diff
@@ -114,6 +104,9 @@ var _ = BeforeSuite(func() {
 	Expect(dynamicClient).NotTo(BeNil())
 	Expect(err).NotTo(HaveOccurred())
 
+	wg.Add(1)
+	go watchPods()
```
juliev0, line 108: We could do this in a separate PR, but what if we watched all resource types just like this instead of having to create the watch after the individual resource gets created?

I'm also wondering if there's any shared code for these various watches. What do you think?

juliev0: also, is there no `disableWatch` check around this?

dpadhiar: Ack, missed that one when updating all the others will add.

> We could do this in a separate PR, but what if we watched all resource types just like this instead of having to create the watch after the individual resource gets created?

As in do we want to start all watches before the particular resource is created? That would be fine, let me know if I'm mixing this up otherwise.

> I'm also wondering if there's any shared code for these various watches. What do you think?

There is a fair bit of shared logic for each one aside from file path and the pod logic is slightly different. The main issue is these lines where we cast the return event object as the actual type we need:

```
if event.Type == watch.Modified {
	if rollout, ok := event.Object.(*apiv1.ISBServiceRollout); ok {
		rollout.ManagedFields = nil
		rl := Output{
			APIVersion: NumaplaneAPIVersion,
			Kind:       "ISBServiceRollout",
			Metadata:   rollout.ObjectMeta,
			Spec:       rollout.Spec,
			Status:     rollout.Status,
		}
```

Haven't thought of the best way to consolidate this yet.

juliev0: > As in do we want to start all watches before the particular resource is created?

Yes, I think it would be consistent to have the approach be the same for all resources. Plus, it would mean not needing to specifically start the watch at the time when the Resource is created. Let me know if the request doesn't make sense for any reason.

juliev0:
Oh, sorry I didn't clarify earlier but can we have the disableWatch flag also prevent the capture of the numaplane and numaflow log files?

juliev0:
> Oh, sorry I didn't clarify earlier but can we have the disableWatch flag also prevent the capture of the numaplane and numaflow log files?

and maybe instead of calling it `disableWatch`, we could just call it `disableTestArtifacts` or something?

tests/e2e/suite_test.go
juliev0: > Haven't thought of the best way to consolidate this yet.

Feel free to do this and the other request in a follow up. If you get stuck on the consolidation part, let me know. 

If two functions have the same general structure but within that structure they need to take some different action, one approach you can take is to pass a function into the generic function.

juliev0 (approved):

xdevxy (approved):

PR #298, opened 2024-09-26
chore: remove redundant check for in progress upgrade strategy

xdevxy:
### Modifications
This PRs remove a redundant check for in progress upgrade strategy

### Verification
`make test`

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -46,7 +46,6 @@ import (
 	numaflowv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaplane/internal/common"
 	"github.com/numaproj/numaplane/internal/controller/config"
-
 	"github.com/numaproj/numaplane/internal/usde"
 	"github.com/numaproj/numaplane/internal/util"
 	"github.com/numaproj/numaplane/internal/util/kubernetes"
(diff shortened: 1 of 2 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -467,21 +466,18 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
...
-				needPPND = *ppndRequired
+			ppndRequired, err := r.needPPND(ctx, pipelineRollout, newPipelineDef, upgradeStrategyType == apiv1.UpgradeStrategyPPND)
+			if err != nil {
+				return err
+			}
+			if ppndRequired == nil { // not enough information
+				// TODO: mark something in the Status for why we're remaining in "Pending" here
+				return nil
 			}
+			needPPND = *ppndRequired
```
juliev0, line 480: oops, thanks for noticing this!

internal/controller/pipelinerollout_controller.go
```diff
@@ -467,21 +466,18 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 
 	// if not, should we set one?
 	if !inProgressStrategySet {
-		// inProgressStrategy is used for PPND and Progressive strategies (i.e. any strategies which require multiple reconciliations to perform)
```
juliev0, line 470: you don't think we need a comment like this? (although maybe above)

xdevxy: I feel this comment is a little bit irrelevant and redundant, so removed it. And the code above seems well documented to me. :)

juliev0 (approved):

PR #305, opened 2024-09-27
feat: use the `promoted` label to identify the running pipeline

### LINKED ISSUE #304: Identify the running pipeline by utilizing the `promoted` label

xdevxy:
Fixes #304 

### Modifications
This PR Utilize the `promoted` label to identify the running pipeline managed by a PipelineRollout. If such pipeline does not exist, the name should be calculated using the `NameCount` suffix.

### Verification

`make test`

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -757,7 +757,7 @@ func (r *PipelineRolloutReconciler) getPipelineName(ctx context.Context, pipelin
 }
 
 // calPipelineNameSuffix calculates the suffix of the pipeline name by utilizing the `NameCount`
-// field. 
+// field.
 func (r *PipelineRolloutReconciler) calPipelineNameSuffix(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) (string, error) {
 	if pipelineRollout.Status.NameCount == nil {
 		pipelineRollout.Status.NameCount = new(int32)
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -374,7 +374,7 @@ func (r *PipelineRolloutReconciler) reconcile(
 		controllerutil.AddFinalizer(pipelineRollout, finalizerName)
 	}
 
-	newPipelineDef, err := r.makePipelineDefinition(ctx, pipelineRollout)
+	newPipelineDef, err := r.getPipelineDefinition(ctx, pipelineRollout)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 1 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -731,28 +731,62 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
...
-			return nil, statusUpdateErr
+			return "", statusUpdateErr
 		}
 	}
 
-	curNameCount := *pipelineRollout.Status.NameCount
+	return "-" + fmt.Sprint(*pipelineRollout.Status.NameCount), nil
+}
+
+func (r *PipelineRolloutReconciler) getPipelineDefinition(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) (*kubernetes.GenericObject, error) {
```
juliev0, line 773: nit: do we rename it? The new name makes it sound like we're finding the definition of the existing pipeline when we're actually creating a new definition - do you agree? I'm okay with adding a comment to clarify if you prefer this name.

internal/controller/pipelinerollout_controller.go
```diff
@@ -731,28 +731,62 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
 	return r.updatePipelineRolloutStatus(ctx, pipelineRollout)
 }
 
-func (r *PipelineRolloutReconciler) makePipelineDefinition(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) (*kubernetes.GenericObject, error) {
-	labels, err := pipelineLabels(pipelineRollout)
+// getPipelineName retrieves the name of the current running pipeline managed by the given
+// pipelineRollout through the `promoted` label. Unless there is non such pipeline exist, then
+// construct the name by calculate the suffix and append to the PipelineRollout name.
```
juliev0, line 736:
```suggestion
// pipelineRollout through the `promoted` label. If no such pipeline exists, then it 
// constructs the name by calculating the suffix and appending to the PipelineRollout name.
```

juliev0:
I'm curious about the overall plan as far as using both "promoted" label and count suffix. Should I refer to your design document for that?

internal/controller/pipelinerollout_controller.go
xdevxy: yeah, let me keep the existing naming

internal/controller/pipelinerollout_controller.go
```diff
@@ -731,28 +731,62 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
...
+func (r *PipelineRolloutReconciler) getPipelineName(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) (string, error) {
+	pipelines, err := kubernetes.ListCR(
+		ctx, r.restConfig, common.NumaflowAPIGroup, common.NumaflowAPIVersion, "pipelines",
+		pipelineRollout.Namespace, fmt.Sprintf(
+			"%s=%s,%s=%s", common.LabelKeyPipelineRolloutForPipeline, pipelineRollout.Name,
+			common.LabelKeyPipelineRolloutForPipeline, common.LabelValueUpgradePromoted,
```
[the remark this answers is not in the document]
xdevxy: thanks for catching that!

xdevxy:
> m thinking about how

The high level contract is if we cannot find from the promoted label, then we should be able to reconstruct deterministically from the `NameCount` in the status field.

juliev0 (approved):

PR #313, opened 2024-10-02
fix: don't wait for pipeline to finish updating because we can't distinguish failure from progressing

juliev0:
### Modifications

This makes it so that we no longer require to be paused simply because pipeline is still reconciling. 

The original process of updating was:
1. pause pipeline
2. wait for it to be paused
3. apply new pipeline spec with `desiredPhase=Paused`
4. wait for pipeline change to be reconciled based on `pipeline.ObservedGeneration` and pipeline Conditions, which indicate whether daemon, vertices, side inputs are fully reconciled
5. set `desiredPhase=Running`

I have to remove step 4 because it's impossible to know based on Pipeline Conditions whether Pipeline is in fact still "progressing" or failed. This becomes clear when we actually deploy a bad pipeline.

Why I think this is okay to remove: After the Pipeline has been paused, it goes down to 0 vertices. After we apply the spec, Numaflow Controller updates the Vertex specs. Then we set it to `desiredPhase=Running`, and it scales those Vertices back up. My original motivation to wait for it to fully reconcile was out of paranoia that we would somehow be "Running" an older Vertex but now this seems unnecessary.

### Verification

Ran e2e test about 6 times with success every time

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -822,16 +822,6 @@ func getPipelineChildResourceHealth(conditions []metav1.Condition) (metav1.Condi
 	return "True", ""
 }
 
-/*
-func checkChildResources(conditions []metav1.Condition, f func(metav1.Condition) bool) (bool, metav1.Condition) {
-	for _, cond := range conditions {
-		if f(cond) {
-			return true, cond
-		}
-	}
-	return false, metav1.Condition{}
-}*/
-
 func (r *PipelineRolloutReconciler) ErrorHandler(pipelineRollout *apiv1.PipelineRollout, err error, reason, msg string) {
 	r.customMetrics.PipelinesSyncFailed.WithLabelValues().Inc()
 	r.recorder.Eventf(pipelineRollout, corev1.EventTypeWarning, reason, msg+" %v", err.Error())
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -822,6 +822,7 @@ func getPipelineChildResourceHealth(conditions []metav1.Condition) (metav1.Condi
 	return "True", ""
 }
 
+/*
 func checkChildResources(conditions []metav1.Condition, f func(metav1.Condition) bool) (bool, metav1.Condition) {
 	for _, cond := range conditions {
 		if f(cond) {
@@ -829,7 +830,7 @@ func checkChildResources(conditions []metav1.Condition, f func(metav1.Condition)
 		}
 	}
 	return false, metav1.Condition{}
-}
+}*/
 
 func (r *PipelineRolloutReconciler) ErrorHandler(pipelineRollout *apiv1.PipelineRollout, err error, reason, msg string) {
 	r.customMetrics.PipelinesSyncFailed.WithLabelValues().Inc()
--- a/internal/controller/pipelinerollout_controller_test.go
+++ b/internal/controller/pipelinerollout_controller_test.go
@@ -669,14 +669,9 @@ func createPipelineRollout(isbsvcSpec numaflowv1.PipelineSpec) *apiv1.PipelineRo
 	}
 }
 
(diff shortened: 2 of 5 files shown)


tests/e2e/pipeline.go
```diff
@@ -110,15 +110,16 @@ func verifyPipelinePaused(namespace string, pipelineRolloutName string, pipeline
 	}, testTimeout).Should(Equal(metav1.ConditionTrue))
 
 	document("Verify that Pipeline is paused and fully drained")
-	verifyPipelineStatusEventually(Namespace, pipelineName,
+	verifyPipelineStatusEventually(namespace, pipelineName,
```
juliev0, line 113: we weren't using the parameter passed in to the function

tests/e2e/functional_test.go
```diff
@@ -251,7 +251,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		verifyPipelineRolloutDeployed(pipelineRolloutName)
 		verifyPipelineRolloutHealthy(pipelineRolloutName)
-		verifyInProgressStrategy(Namespace, pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+		verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
```
juliev0, line 254: remove unused parameter from this function

internal/controller/pipelinerollout_controller.go
```diff
@@ -822,15 +822,6 @@ func getPipelineChildResourceHealth(conditions []metav1.Condition) (metav1.Condi
...
 
-func checkChildResources(conditions []metav1.Condition, f func(metav1.Condition) bool) (bool, metav1.Condition) {
-	for _, cond := range conditions {
-		if f(cond) {
-			return true, cond
-		}
-	}
-	return false, metav1.Condition{}
-}
-
```
juliev0, line 833: no longer used

internal/controller/pipelinerollout_controller_test.go
```diff
@@ -669,15 +669,10 @@ func createPipelineRollout(isbsvcSpec numaflowv1.PipelineSpec) *apiv1.PipelineRo
 	}
 }
 
-func createPipelineOfSpec(spec numaflowv1.PipelineSpec, phase numaflowv1.PipelinePhase, fullyReconciled bool) *numaflowv1.Pipeline {
+func createPipelineOfSpec(spec numaflowv1.PipelineSpec, phase numaflowv1.PipelinePhase) *numaflowv1.Pipeline {
```
juliev0, line 672: not needed in the test anymore

internal/controller/pipelinerollout_ppnd.go
```diff
@@ -237,25 +222,6 @@ func (r *PipelineRolloutReconciler) setPipelineLifecycle(ctx context.Context, pa
...
-	}
-
-	// note if Pipeline's children are still being updated
-	unhealthyOrProgressing, _ := checkChildResources(existingPipelineStatus.Conditions, func(c metav1.Condition) bool {
-		return c.Status == metav1.ConditionFalse
-	})
-
-	return unhealthyOrProgressing, nil
-
-}
```
juliev0, line 258: no longer used

tests/e2e/pipeline.go
```diff
@@ -110,15 +110,16 @@ func verifyPipelinePaused(namespace string, pipelineRolloutName string, pipeline
...
 		func(retrievedPipelineSpec numaflowv1.PipelineSpec, retrievedPipelineStatus numaflowv1.PipelineStatus) bool {
 			return retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePaused && retrievedPipelineStatus.DrainedOnPause
 
 		})
-	verifyPodsRunning(namespace, 0, getVertexLabelSelector(pipelineName))
+	// this happens too fast to verify it:
+	//verifyPodsRunning(namespace, 0, getVertexLabelSelector(pipelineName))
```
juliev0, line 119: this can happen too fast now to verify 0 Pods now but note that Numaflow has in fact indicated phase=Paused and drainedOnPause=true in the previous step, which it would only do if there were 0 Pods

xdevxy (approved):

PR #315, opened 2024-10-02
feat: enable Numaplane to deploy Secrets and Services for Numaflow Controller; add new Numaflow Controller versions

juliev0:
updated `includedResources` and the numaplane ClusterRole to be able to deploy Secret and Service types. Replaces old Numaflow manifests with 2 which both represent Numaflow 1.3.3 (which has fixes for pausing).


tests/e2e/functional_test.go
```diff
@@ -290,6 +290,8 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 	})
 
+	time.Sleep(2 * time.Second)
```
juliev0, line 293: there's a 2 second sleep between any changes we do and I'd previously forgotten it in this spot - this is just to have a little more "real world timing" in the test - this was thought to make things a little easier on numaflow - we can potentially take it out later

tests/e2e/functional_test.go
```diff
@@ -724,7 +726,7 @@ func createNumaflowControllerRolloutSpec(name, namespace string) *apiv1.Numaflow
 			Namespace: namespace,
 		},
 		Spec: apiv1.NumaflowControllerRolloutSpec{
-			Controller: apiv1.Controller{Version: "0.0.18"},
+			Controller: apiv1.Controller{Version: "1.3.3"},
```
kohlisid, line 729: Maybe keep these defined as constants

tests/e2e/functional_test.go
```diff
@@ -466,7 +468,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 		// TODO: update this controller image when Numaflow v1.3.1 is released
```
kohlisid, line 468: Do we need the TODO now?

juliev0: good catch

tests/e2e/functional_test.go
juliev0: replaced these and the Jetstream versions both with constants

kohlisid (approved):

PR #317, opened 2024-10-02
feat: perform selected strategy for isbsvc fields to support PPND

### LINKED ISSUE #248: Integrate USDE logic into Rollout controllers to support PPND - perform selected strategy for isbsvc fields

dpadhiar:
Fixes #248 

### Modifications

Integrates the USDE logic to derive an upgrade strategy (that we currently have for PipelineRollouts) into the ISBServiceRollout reconciliation process.

### Verification

Ran `make test` to verify unit tests are passing.

Ran `make test-e2e` locally and on CI to see that updating with and without excluded fields works properly.


tests/e2e/functional_test.go
```diff
@@ -524,6 +541,56 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+			document("Verify that in-progress-strategy gets set to NoOp")
+			verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyNoOp)
+			verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+
+			if dataLossPrevention == "true" {
+				document("Verify that dependent Pipeline is not paused when an update to ISBService not requiring pause is made")
+				verifyPipelineStatusConsistently(Namespace, pipelineName, func(retrievedPipelineSpec numaflowv1.PipelineSpec, retrievedPipelineStatus numaflowv1.PipelineStatus) bool {
+					return retrievedPipelineStatus.Phase != numaflowv1.PipelinePhasePaused
+				})
+			}
```
juliev0, line 566: looks like `if dataLossPrevention == "true"` is in here twice

dpadhiar: Oops, fixing that now.

tests/e2e/functional_test.go
```diff
@@ -524,6 +541,56 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+		updateISBServiceRolloutInK8S(isbServiceRolloutName, func(rollout apiv1.ISBServiceRollout) (apiv1.ISBServiceRollout, error) {
+			rollout.Spec.InterStepBufferService.Spec.Raw = rawSpec
+			return rollout, nil
+		})
+
+		if dataLossPrevention == "true" {
+
+			document("Verify that in-progress-strategy gets set to NoOp")
+			verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyNoOp)
+			verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
```
juliev0, line 559: hmm, just thinking that these may have no effect, as they may have already been "NoOp"

dpadhiar: This is true, is there no chance it could be set to PauseAndDrain incorrectly? If not, can remove these checks.

tests/e2e/functional_test.go
```diff
@@ -524,6 +541,54 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+			Eventually(func() bool {
+				isbRollout, _ := isbServiceRolloutClient.Get(ctx, isbServiceRolloutName, metav1.GetOptions{})
+				isbCondStatus := getRolloutCondition(isbRollout.Status.Conditions, apiv1.ConditionPausingPipelines)
+				plRollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
+				plCondStatus := getRolloutCondition(plRollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
+				if isbCondStatus == metav1.ConditionTrue || plCondStatus == metav1.ConditionTrue {
+					return false
+				}
+				return true
+			}, testTimeout).Should(BeTrue())
```
juliev0, line 575: We really need to do this "consistently", not "eventually", right? We kind of need to verify that lines 562-575 are happening consistently.

internal/controller/isbservicerollout_controller.go
```diff
@@ -278,32 +302,60 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 	r.processISBServiceStatus(ctx, existingISBServiceDef, isbServiceRollout)
 
 	// if I need to update or am in the middle of an update of the ISBService, then I need to make sure all the Pipelines are pausing
```
juliev0, line 304: can you update the comment to remove "need to update or"?

tests/e2e/functional_test.go
dpadhiar: That sounds right, how long should we check - 30 seconds? 1 minute?

tests/e2e/functional_test.go
juliev0: or maybe you can group it with the same "consistently" code I mention below?

juliev0:
nice job

tests/e2e/functional_test.go
juliev0: How about 20 sec or something around there?

dpadhiar: So it looks like consistently will immediately fail as soon as the pausing pipelines condition is true unlike eventually which waits till they're both false. That's why it keeps failing quickly.

dpadhiar: Taking a look at the logs.. the Pipeline does pause. That shouldn't be happening in this case, hm.

juliev0: > So it looks like consistently will immediately fail as soon as the pausing pipelines condition is true unlike eventually which waits till they're both false

Yes, when we check `Consistently`, we do need to check `Eventually` first so we can make sure it got to that state in the first place

juliev0: Ideally we can first "Eventually" and then "Consistently" check:

- verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyNoOp)
- verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
- PipelineStatus != Pausing or Paused
- PausingConditions false

To avoid repeating a large chunk of code, if we can't think of good functions, we can always create a temporary function to call each time within this function like:
```
verifyNotPausing := func(str string) {
       .....
    }

Consistently(....
   verifyNotPausing
}

Eventually(...
   verifyNotPausing
}
```

tests/e2e/functional_test.go
```diff
@@ -524,6 +541,50 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+		// new ISBService spec
+		rawSpec, err := json.Marshal(ISBServiceSpecExcludedField)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		updateISBServiceRolloutInK8S(isbServiceRolloutName, func(rollout apiv1.ISBServiceRollout) (apiv1.ISBServiceRollout, error) {
+			rollout.Spec.InterStepBufferService.Spec.Raw = rawSpec
+			return rollout, nil
+		})
+
+		if dataLossPrevention == "true" {
```
juliev0, line 555: nit: I believe the block of code should apply whether or not `dataLossPrevention == true`?

juliev0:
Spoke to @dpadhiar and when he gets time he'll explore why the previous test for updating isbsvc with PPND is sometimes failing before we merge this. (Merging this is not critically important to be done by Q1)

juliev0:
Actually, I went through the artifacts I had when I saw the failure locally - not sure if numaplane is really doing anything wrong - we ultimately need to get the fixes from numaflow and try with that. I'm still a little nervous about merging this in since we weren't seeing this failure before.

internal/controller/isbservicerollout_controller.go
```diff
@@ -277,33 +301,61 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
...
-	case config.PPNDStrategyID:
+	// is there currently an inProgressStrategy for the isbService? (This will override any new decision)
+	inProgressStrategy := r.inProgressStrategyMgr.getStrategy(ctx, isbServiceRollout)
+	inProgressStrategySet := (inProgressStrategy != apiv1.UpgradeStrategyNoOp)
 
-		return processChildObjectWithPPND(ctx, r.client, isbServiceRollout, r, isbServiceNeedsUpdating, isbServiceIsUpdating, func() error {
+	// if not, should we set one?
+	if !inProgressStrategySet {
+		if userPreferredStrategy == config.PPNDStrategyID {
+			if upgradeStrategyType == apiv1.UpgradeStrategyPPND {
```
juliev0, line 343: nit: I know you copied this from my redundant code in PipelineRolloutController, so my bad, but I believe the check for `userPreferredStrategy` here and below may be unnecessary as it's happening within `ResourceNeedsUpdating()`. We probably don't need to get the value for `userPreferredStrategy` above at all.

internal/controller/isbservicerollout_controller_test.go
```diff
@@ -260,119 +261,133 @@ func Test_reconcile_isbservicerollout_PPND(t *testing.T) {
...
-			existingStatefulSetDef: createDefaultISBStatefulSet("2.10.11", true),
-			existingPipeline:       createDefaultPipelineOfPhase(numaflowv1.PipelinePhasePaused, map[string]string{}),
-			expectedRolloutPhase:   apiv1.PhaseDeployed,
+			name:                       "existing ISBService - spec already updated - isbsvc done reconciling",
+			newISBSvcSpec:              createDefaultISBServiceSpec("2.10.11"),
+			existingISBSvcDef:          createDefaultISBService("2.10.11", numaflowv1.ISBSvcPhaseRunning, true),
+			existingStatefulSetDef:     createDefaultISBStatefulSet("2.10.11", true),
+			existingPipeline:           createDefaultPipelineOfPhase(numaflowv1.PipelinePhasePaused, map[string]string{}),
+			expectedRolloutPhase:       apiv1.PhaseDeployed,
+			existingInProgressStrategy: nil,
```
juliev0, line 385: nit: I'm thinking in this case the incoming InProgressStrategy would probably be PPND

tests/e2e/functional_test.go
```diff
@@ -524,6 +547,50 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+		})
+
+		verifyISBSvcRolloutReady(isbServiceRolloutName)
+
+		verifyISBSvcReady(Namespace, isbServiceRolloutName, 3)
+
+		verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyNoOp)
+		verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+		verifyPipelineRunning(Namespace, pipelineName, 3)
+
```
juliev0, line 591: some of these are being checked in the `Consistently` block above I think, and we don't need them ?

tests/e2e/functional_test.go
```diff
@@ -524,6 +547,50 @@ var _ = Describe("Functional e2e", Serial, func() {
...
+			return rollout, nil
+		})
+
+		document("Verify that dependent Pipeline is not paused when an update to ISBService not requiring pause is made")
+		verifyNotPausing := func() bool {
+			verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyNoOp)
+			verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+			verifyPipelineStatusConsistently(Namespace, pipelineName, func(retrievedPipelineSpec numaflowv1.PipelineSpec, retrievedPipelineStatus numaflowv1.PipelineStatus) bool {
+				return retrievedPipelineStatus.Phase != numaflowv1.PipelinePhasePaused
+			})
```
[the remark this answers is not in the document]
dpadhiar: Maybe if we just use Eventually instead of the Consistently version of this function it should work.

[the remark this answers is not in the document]
juliev0: yeah, but then it's not truly doing a "consistently" check. The purpose is to verify at every time interval that the whole thing is try. Imagine that the eventually statement inside is blocking for a long period of time.

juliev0:
FYI, Sidhant says he can get his numaflow changes ready to merge into our testing soon, so then we can see if that intermittent failure with the CI that we were seeing before goes away (and if not, we may have something for him to look at). Thanks for your patience.

tests/e2e/functional_test.go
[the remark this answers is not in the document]
dpadhiar: Ack, will make this a single check.

tests/e2e/functional_test.go
dpadhiar: Removed checking the in progress strategy after the Consistently block.

dpadhiar:
@juliev0 Addressed all review changes, I did see that intermittent failure a few times (I believe its when the ISBService is ready-ing) so I suppose we will still wait to merge on this.

tests/e2e/functional_test.go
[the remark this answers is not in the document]
juliev0: What I was saying was that inside the `Consistently` which calls this block, what it should do is not contain any `Eventually` statements at all (or `Consistently` statements). It should simply do a `Get()` on each of the different resources, and check that they are correct every iteration.

tests/e2e/functional_test.go
juliev0: I just realized the two statements above `verifyInProgressStrategyISBService` and `verifyInProgressStrategy` are `Eventually` statements themselves - we want to avoid those being within a `Consistently`.

juliev0 (approved):

Issue #295, opened 2024-09-25
Add capability to force reapply Pipeline when it's stuck draining


juliev0:
# Summary

See here: https://github.com/numaproj/numaplane/issues/74#issuecomment-2349945359

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #323, opened 2024-10-04
feat: Allow annotation to specify that a pipeline should temporarily allow data loss

juliev0:
Fixes: #295 

### Modifications

I have actually written the ArgoCD extension locally. I have added to it the capability to apply an "allow-data-loss" annotation to the PipelineRollout while the PipelineRollout is in the middle of the PPND Upgrade strategy, and to remove it any time that the annotation is present, whether or not in the middle of PPND Upgrade.

This introduces that annotation. The annotation is observed whether the Pipeline is pausing due to its own need to update or if it's pausing because the Numaflow Controller or isbsvc are updating.

### Verification

Tested end to end with ArgoCD. Tested with a slow Pipeline in the middle of draining, in which I added the annotation in the middle of "Pausing" - it did not complete the drain and updated the spec and went back to "Running" at that time.

Also tested as part of a Numaflow Controller update (with the same slow pipeline). In the middle of the Numaflow Controller trying to update and waiting for the pause, I added the annotation on the PipelineRollout. The pipeline did not get drained fully and went back to "Running".

### DIFF
--- a/internal/controller/pipelinerollout_ppnd.go
+++ b/internal/controller/pipelinerollout_ppnd.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	numaflowv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+	"github.com/numaproj/numaplane/internal/common"
 	"github.com/numaproj/numaplane/internal/util/kubernetes"
 	"github.com/numaproj/numaplane/internal/util/logger"
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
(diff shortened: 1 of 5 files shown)


internal/controller/pipelinerollout_ppnd.go
```diff
@@ -53,7 +54,7 @@ func (r *PipelineRolloutReconciler) processExistingPipelineWithPPND(ctx context.
 	// if it's safe to Update and we need to, do it now
 	if pipelineNeedsToUpdate {
 		pipelineRollout.Status.MarkPending()
-		if !shouldBePaused || (shouldBePaused && isPipelinePausedOrUnpausible(ctx, existingPipelineDef)) {
+		if !shouldBePaused || (shouldBePaused && checkPipelineStatus(ctx, existingPipelineDef, numaflowv1.PipelinePhasePaused)) {
```
juliev0, line 57: I actually don't need to check for the case of "won't pause/unpausible" here as it has already been checked in the derivation of `shouldBePaused`, which will be set to false here for the case of "won't pause/unpausible".

PR #327, opened 2024-10-07
feat: Pipeline No Downtime Upgrade

### LINKED ISSUE #225: PipelineRollout Controller Support for Pipeline No Downtime Update
# Summary

See the section for pipeline upgrade in the [design doc](https://docs.google.com/document/d/1UH_0DEc9EpiTwmnSHLFwnP6GRFXiAb-gYv1Eyvi63DY/edit?usp=sharing)

# Use Cases

Upgrade pipeline without downtime

- [x] #243
- [x] #254
- [x] #272
- [x] #304
- [ ] #334

---

xdevxy:
Fixes #225

### Modifications

Pipeline No Downtime upgrade which creates a second pipeline when upgrade strategy is `progressive`. A follow up  will do clean up for the old pipeline, e.g. pause the pipeline and delete after drain is completed.

### Verification

`make test`
Run in local cluster, apply a namespace config to set the upgrade strategy to `progressive` and see a new pipeline running and `PipelineProgressiveUpgradeSucceed` in the PipelineRollout to be true.

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -374,7 +374,7 @@ func (r *PipelineRolloutReconciler) reconcile(
 		controllerutil.AddFinalizer(pipelineRollout, finalizerName)
 	}
 
-	newPipelineDef, err := r.makePipelineDefinition(ctx, pipelineRollout)
+	newPipelineDef, err := r.makeRunningPipelineDefinition(ctx, pipelineRollout)
 	if err != nil {
 		return false, err
 	}
@@ -506,15 +506,13 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 
 	case apiv1.UpgradeStrategyProgressive:
 		if pipelineNeedsToUpdate {
-			numaLogger.Error(errors.New("Progressive not supported"), "Progressive not supported")
-			//TODO (just an idea of what it might look like below...)
-			//	done, err := r.processExistingPipelineWithProgressive(...)
-			//if err != nil {
-			//	return err
-			//}
-			//if done {
-			//	r.unsetInProgressStrategy(namespacedName)
-			//}
+			done, err := r.processExistingPipelineWithProgressive(ctx, pipelineRollout, newPipelineDef, pipelineNeedsToUpdate)
+			if err != nil {
+				return err
+			}
+			if done {
+				r.inProgressStrategyMgr.unsetStrategy(ctx, pipelineRollout)
+			}
 		}
 	default:
 		if pipelineNeedsToUpdate && upgradeStrategyType == apiv1.UpgradeStrategyApply {
@@ -539,7 +537,7 @@ func pipelineObservedGenerationCurrent(generation int64, observedGeneration int6
 func (r *PipelineRolloutReconciler) processPipelineStatus(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) error {
 	numaLogger := logger.FromContext(ctx)
 
-	pipelineDef, err := r.makePipelineDefinition(ctx, pipelineRollout)
+	pipelineDef, err := r.makeRunningPipelineDefinition(ctx, pipelineRollout)
 	if err != nil {
 		return err
 	}
@@ -705,7 +703,7 @@ func updatePipelineSpec(ctx context.Context, restConfig *rest.Config, obj *kuber
 	return kubernetes.UpdateCR(ctx, restConfig, obj, "pipelines")
 }
 
-func pipelineLabels(pipelineRollout *apiv1.PipelineRollout) (map[string]string, error) {
+func pipelineLabels(pipelineRollout *apiv1.PipelineRollout, upgradeState string) (map[string]string, error) {
 	var pipelineSpec PipelineSpec
 	labelMapping := map[string]string{
 		common.LabelKeyISBServiceNameForPipeline: "default",
@@ -718,7 +716,7 @@ func pipelineLabels(pipelineRollout *apiv1.PipelineRollout) (map[string]string, 
 	}
 
 	labelMapping[common.LabelKeyPipelineRolloutForPipeline] = pipelineRollout.Name
-	labelMapping[common.LabelKeyUpgradeState] = string(common.LabelValueUpgradePromoted)
+	labelMapping[common.LabelKeyUpgradeState] = upgradeState
 
 	return labelMapping, nil
 }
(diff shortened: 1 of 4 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -767,19 +770,34 @@ func (r *PipelineRolloutReconciler) calPipelineNameSuffix(ctx context.Context, p
 		}
```
juliev0, line 770: Just realizing we're calling `updatePipelineRolloutStatus()` above. Generally, we've updated the in-memory value of `Status` during the reconciliation but only at the end posted the Update to K8S - is there a good reason to do it differently here?

internal/controller/pipelinerollout_controller.go
```diff
@@ -732,20 +731,24 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
 }
 
 // getPipelineName retrieves the name of the current running pipeline managed by the given
-// pipelineRollout through the `promoted` label. If no such pipeline exists, then it
-// constructs the name by calculating the suffix and appending to the PipelineRollout name.
-func (r *PipelineRolloutReconciler) getPipelineName(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) (string, error) {
+// pipelineRollout through the `promoted` label. Unless there is non such pipeline exist, then
```
juliev0, line 734:
```suggestion
// pipelineRollout through the `promoted` label. Unless no such pipeline exists, then
```

internal/controller/pipelinerollout_controller.go
```diff
@@ -732,20 +731,24 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
...
+// construct the name by calculate the suffix and append to the PipelineRollout name.
```
juliev0, line 735:
```suggestion
// construct the name by calculating the suffix and appending to the PipelineRollout name.
```

internal/controller/pipelinerollout_progressive.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		pipelineRollout.Status.MarkPipelineProgressiveUpgradeSucceeded("New Pipeline Running", pipelineRollout.Generation)
+		pipelineRollout.Status.MarkDeployed(pipelineRollout.Generation)
+
+		// TODO: pause old pipeline
+		return true, nil
+	} else {
```
juliev0, line 125: I like this logic - it handles the race condition of the existing upgrading Pipeline entering "Failed" state or "Running" state at the same time as there's a new update in the PipelineRollout spec. In that case we mark as "promoted" or failed, and then in a subsequent reconciliation actually process the latest desired PipelineRollout spec.

juliev0:
Did you test going back and forth between "DirectApply" field changes and Progressive changes?

internal/controller/pipelinerollout_progressive.go
juliev0: Although, just thinking that theoretically, in the case that there are 2 back to back updates prior to the pipeline entering "Running", we may notice it "Running" and mark it as such prior to Numaflow having reconciled the very latest update that we wrote. That's unlikely in this current state to happen, but once the Progressive process becomes longer, we may need to be cognizant of the Pipeline's ObservedGeneration.

juliev0: In the future, we may even want to stop the current upgrade which is in progress and have a whole new one if this happens.

internal/controller/pipelinerollout_controller.go
```diff
@@ -732,20 +731,24 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
...
+func (r *PipelineRolloutReconciler) getPipelineName(
+	ctx context.Context,
+	pipelineRollout *apiv1.PipelineRollout,
+	upgradeState string,
+) (string, error) {
```
juliev0, line 740: This naming seems good. So, if I'm following along correctly:
1. we create our very first Pipeline for the PipelineRollout as "<pipeline>-1" with Label "promoted". 
2. Then say we make a Direct apply change: we simply update "<pipeline>-1" since it is the one with Label "promoted".
3. Then say we need to make a Progressive change: we don't find a Pipeline labeled "upgrading", so we create a new one "<pipeline>-2" with label "upgrading".
4. If for some reason somebody makes a change to the Pipeline spec before "<pipeline>-2" is "Running" or "Failed", then we update that spec again. 
5. Once "<pipeline>-2" is "Running", we mark it "promoted". 
6. Next time we need a Progressive update, we notice that there's no "upgrading" state so we create a new Pipeline "<pipeline>-3"

internal/controller/pipelinerollout_controller.go
xdevxy: I'm following the deployment reconciliation model here. We want to avoid the naming collision as much as possible, if `NameCount` is already updated we want to know that and return error here.

internal/controller/pipelinerollout_progressive.go
xdevxy: I don't think stopping the current upgrade is needed, as we always check if there is a new spec change needs to happen by compare the rollout spec and the `promoted` pipeline spec (which is a new pipeline).

xdevxy:
> Did you test going back and forth between "DirectApply" field changes and Progressive changes? (and maybe even trying to change the user strategy between PPND and Progressive? (although, I'm fine to merge before that one))

No I don't. I think there definitely can be more testing (especially e2e)added as a follow up

internal/controller/pipelinerollout_controller.go
xdevxy: yes, that's correct.

internal/controller/pipelinerollout_controller.go
juliev0: Got it

internal/controller/pipelinerollout_progressive.go
juliev0: Well, in the future when we have "Progressive", I'm thinking say we're analyzing how successfully that new Pipeline is running - I don't want to be in a situation in which we replace the new Pipeline manifest and we still use the original data. That's why I thought one possible way to deal with it is to stop the current upgrade and start up a new one.

juliev0:
> > Did you test going back and forth between "DirectApply" field changes and Progressive changes? (and maybe even trying to change the user strategy between PPND and Progressive? (although, I'm fine to merge before that one))
> 
> No I don't. I think there definitely can be more testing (especially e2e)added as a follow up

I'm good to merge it, but do you think you may have time to do the test of going between "DirectApply" and Progressive, since that one's a very typical case? Certainly, we could test this after you're gone, but would be good data to have if you have time.

internal/controller/pipelinerollout_progressive.go
xdevxy: ah got it, that could be needed in that case

xdevxy:
> > > Did you test going back and forth between "DirectApply" field changes and Progressive changes? (and maybe even trying to change the user strategy between PPND and Progressive? (although, I'm fine to merge before that one))
> > 
> > 
> > No I don't. I think there definitely can be more testing (especially e2e)added as a follow up
> 
> I'm good to merge it, but do you think you may have time to do the test of going between "DirectApply" and Progressive, since that one's a very typical case? Certainly, we could test this after you're gone, but would be good data to have if you have time.

yeah sure, I can do the testing when I get time.

xdevxy:
> > > > Did you test going back and forth between "DirectApply" field changes and Progressive changes? (and maybe even trying to change the user strategy between PPND and Progressive? (although, I'm fine to merge before that one))
> > > 
> > > 
> > > No I don't. I think there definitely can be more testing (especially e2e)added as a follow up
> > 
> > 
> > I'm good to merge it, but do you think you may have time to do the test of going between "DirectApply" and Progressive, since that one's a very typical case? Certainly, we could test this after you're gone, but would be good data to have if you have time.
> 
> yeah sure, I can do the testing when I get time.

Verified the case is working, update the pipeline spec with watermark field will just do `DirectApply` and update the spec of vertices will trigger progressive upgrade.

PR #328, opened 2024-10-07
feat: Use informer cache for monovertex, pipeline and ISB Services

### LINKED ISSUE #95: Not using Informer cache for Pipelines and ISBServices
**Describe the bug**

We made a decision that in Numaplane we don't need to define the Pipeline/ISBService specs - therefore we are using the Kubernetes Dynamic client to create/update/get resources. 

It looks like currently, we are not using an Informer cache and instead are making multiple repeated calls to GET from the Kubernetes API directly.

Assuming that we stick with the decision to use dynamic client (which may first be worth a deep dive analysis), we can incorporate a Dynamic client Informer described [here](https://firehydrant.com/blog/dynamic-kubernetes-informers/). 

See slack [thread](https://intuit-teams.slack.com/archives/C076Q9SD4BF/p1720039913538109). 

**Expected behavior**
When calling Get() for a given resource multiple times in a row when the resource hasn't changed, I shouldn't hit the K8S API every time. Instead, my resource should be updated asynchronously into my local cache.

---

chandankumar4:
Fixes #95

### DIFF
--- a/internal/controller/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout_controller.go
@@ -30,9 +30,12 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/selection"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -191,9 +194,9 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 
 	defer func() {
 		if isbServiceRollout.Status.IsHealthy() {
-			r.customMetrics.ISBServicesHealth.WithLabelValues(isbServiceRollout.Namespace, isbServiceRollout.Name).Set(1)
+			r.customMetrics.ISBServicesRolloutHealth.WithLabelValues(isbServiceRollout.Namespace, isbServiceRollout.Name).Set(1)
 		} else {
-			r.customMetrics.ISBServicesHealth.WithLabelValues(isbServiceRollout.Namespace, isbServiceRollout.Name).Set(0)
+			r.customMetrics.ISBServicesRolloutHealth.WithLabelValues(isbServiceRollout.Namespace, isbServiceRollout.Name).Set(0)
 		}
 	}()
 
@@ -210,7 +213,7 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		// generate metrics for ISB Service deletion.
 		r.customMetrics.DecISBServiceMetrics(isbServiceRollout.Name, isbServiceRollout.Namespace)
 		r.customMetrics.ReconciliationDuration.WithLabelValues(ControllerISBSVCRollout, "delete").Observe(time.Since(startTime).Seconds())
-		r.customMetrics.ISBServicesHealth.DeleteLabelValues(isbServiceRollout.Namespace, isbServiceRollout.Name)
+		r.customMetrics.ISBServicesRolloutHealth.DeleteLabelValues(isbServiceRollout.Namespace, isbServiceRollout.Name)
 		return ctrl.Result{}, nil
 	}
 
@@ -238,15 +241,16 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		Spec: isbServiceRollout.Spec.InterStepBufferService.Spec,
 	}
 
-	existingISBServiceDef, err := kubernetes.GetCR(ctx, r.restConfig, newISBServiceDef, "interstepbufferservices")
+	existingISBServiceDef, err := kubernetes.GetResource(ctx, r.client, newISBServiceDef.GroupVersionKind(),
+		k8stypes.NamespacedName{Namespace: newISBServiceDef.Namespace, Name: newISBServiceDef.Name})
 	if err != nil {
 		// create an object as it doesn't exist
 		if apierrors.IsNotFound(err) {
 			numaLogger.Debugf("ISBService %s/%s doesn't exist so creating", isbServiceRollout.Namespace, isbServiceRollout.Name)
 			isbServiceRollout.Status.MarkPending()
 
-			if err := kubernetes.CreateCR(ctx, r.restConfig, newISBServiceDef, "interstepbufferservices"); err != nil {
-				return ctrl.Result{}, err
+			if err = kubernetes.CreateResource(ctx, r.client, newISBServiceDef); err != nil {
+				return ctrl.Result{}, fmt.Errorf("error creating ISBService: %v", err)
 			}
 
 			isbServiceRollout.Status.MarkDeployed(isbServiceRollout.Generation)
@@ -258,11 +262,10 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 	} else {
 		// Object already exists
 		// perform logic related to updating
-
 		newISBServiceDef = mergeISBService(existingISBServiceDef, newISBServiceDef)
 		needsRequeue, err := r.processExistingISBService(ctx, isbServiceRollout, existingISBServiceDef, newISBServiceDef, syncStartTime)
 		if err != nil {
-			return ctrl.Result{}, err
+			return ctrl.Result{}, fmt.Errorf("error processing existing ISBService: %v", err)
 		}
 		if needsRequeue {
 			return common.DefaultDelayedRequeue, nil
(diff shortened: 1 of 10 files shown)


internal/controller/isbservicerollout_controller.go
```diff
@@ -400,8 +413,19 @@ func (r *ISBServiceRolloutReconciler) isISBServiceUpdating(ctx context.Context,
...
+	}
+	objects := make([]*kubernetes.GenericObject, len(pipelineList.Items))
+	for i, unstruc := range pipelineList.Items {
+		obj, err := kubernetes.UnstructuredToObject(&unstruc)
+		if err != nil {
+			return nil, err
+		}
+		objects[i] = obj
+	}
+	return objects, nil
```
juliev0, line 428: So, I assume that since we aren't adding Unstructured type to the cache for Pipelines yet, then this is going to K8S API?

internal/controller/isbservicerollout_controller.go
```diff
@@ -330,7 +339,11 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 }
 
 func (r *ISBServiceRolloutReconciler) updateISBService(ctx context.Context, isbServiceRollout *apiv1.ISBServiceRollout, newISBServiceDef *kubernetes.GenericObject) error {
-	err := kubernetes.UpdateCR(ctx, r.restConfig, newISBServiceDef, "interstepbufferservices")
+	newISBServiceDefUns, err := kubernetes.ObjectToUnstructured(newISBServiceDef)
+	if err != nil {
+		return err
+	}
+	err = r.client.Update(ctx, newISBServiceDefUns)
```
juliev0, line 346: this was a good finding that it seems like we can maybe just use `r.client` across the board now - maybe our functions in `unstructured_util.go` will get replaced. Perhaps, the one thing we could do over there is have convenience functions that can take `GenericObject` so that we could reduce the code above for example to a single call?

internal/controller/isbservicerollout_controller.go
chandankumar4: Yes, For pipeline currently it's going to k8s API

chandankumar4:
> Hey, @chandankumar4 - how were you saying you were able to confirm that you were using the cache out of curiosity? (vs going to K8S API)

For `Get()` call, [this](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.3/pkg/client/client.go#L351) is the method called from controller-runtime package, So I added some logger there and watched the log in numaplane-controller pod

juliev0:
> For Get() call, [this](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.3/pkg/client/client.go#L351) is the method called from controller-runtime package, So I added some logger there and watched the log in numaplane-controller pod

nice, thanks for the thoroughness!

internal/controller/isbservicerollout_controller.go
```diff
@@ -215,14 +217,18 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		Spec: isbServiceRollout.Spec.InterStepBufferService.Spec,
 	}
 
-	existingISBServiceDef, err := kubernetes.GetCR(ctx, r.restConfig, newISBServiceDef, "interstepbufferservices")
+	existingISBServiceDefUns, err := kubernetes.ObjectToUnstructured(newISBServiceDef)
```
juliev0, line 220: "existing" should only mean "isbsvc that exists in K8S". "new" should only mean the new one which is defined by the ISBServiceRollout

internal/controller/isbservicerollout_controller.go
```diff
@@ -280,7 +281,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 	// if I need to update or am in the middle of an update of the ISBService, then I need to make sure all the Pipelines are pausing
 	isbServiceNeedsUpdating, isbServiceIsUpdating, err := r.isISBServiceUpdating(ctx, isbServiceRollout, existingISBServiceDef)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("error determining if ISBService is updating: %v", err)
```
juliev0, line 284: thanks for expanding on the errors with more info - I know I didn't do that in a lot of places

cmd/main.go
```diff
@@ -144,6 +144,10 @@ func main() {
 		numaLogger.Fatal(err, "Failed to start configmap watcher")
 	}
 
+	if err := kubernetes.SetDynamicClient(newRawConfig); err != nil {
+		numaLogger.Fatal(err, "Failed to set dynamic client set")
+	}
```
chandankumar4, line 149: Creating the raw client initially, instead of generating the clientset all the time

internal/util/kubernetes/unstructured_util.go
```diff
@@ -339,3 +294,64 @@ func (obj *GenericObject) DeepCopy() *GenericObject {
...
+			return err
+		}
+		*obj = *result
+	}
+
+	return nil
+}
+
+// ListResources retrieves the list of resources from the informer cache, if it's not found then it fetches from the API server.
+func ListResources(ctx context.Context, c client.Client, obj *GenericObject, opts ...client.ListOption) ([]*GenericObject, error) {
```
juliev0, line 339: I'm not sure if we did this somewhere else in the code, but I'd prefer to not pass in `obj *GenericObject` - I'd like that kind of parameter to truly represent an object as opposed to just being somewhere to hold a `GroupVersionKind`

internal/controller/monovertexrollout_controller.go
```diff
@@ -192,13 +192,13 @@ func (r *MonoVertexRolloutReconciler) reconcile(ctx context.Context, monoVertexR
 		Spec: monoVertexRollout.Spec.MonoVertex.Spec,
 	}
 
-	existingMonoVertexDef, err := kubernetes.GetCR(ctx, r.restConfig, newMonoVertexDef, "monovertices")
+	existingMonoVertexDef, err := kubernetes.GetLiveResource(ctx, newMonoVertexDef, "monovertices")
```
juliev0, line 195: oh, sorry. I guess the original issue said to just use the cache for getting `pipeline` and `isbsvc`, but actually I had just forgotten about `monovertex` - that is the same. And like `isbsvc`, I think we can just get it from the cache and not worry about getting a live version.

internal/util/kubernetes/unstructured_util.go
juliev0: the `GetResource()` and `GetCR()` functions are kind of like that too - the `GenericObject` is where the name and namespace are coming from, but theoretically we're not actually using the object itself - could consider changing that too, although I know it's been like that for the history of this codebase

internal/controller/pipelinerollout_controller.go
```diff
@@ -492,6 +493,18 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 		}
 	}
 
+	if inProgressStrategy != apiv1.UpgradeStrategyNoOp {
```
juliev0, line 496:
```suggestion
	// don't risk out-of-date cache while performing PPND or Progressive strategy - get 
	// the most current version of the Pipeline just in case
	if inProgressStrategy != apiv1.UpgradeStrategyNoOp {
```

internal/controller/pipelinerollout_controller.go
```diff
@@ -545,7 +558,7 @@ func (r *PipelineRolloutReconciler) processPipelineStatus(ctx context.Context, p
 	}
 
 	// Get existing Pipeline
-	existingPipelineDef, err := kubernetes.GetCR(ctx, r.restConfig, pipelineDef, "pipelines")
+	existingPipelineDef, err := kubernetes.GetLiveResource(ctx, pipelineDef, "pipelines")
```
juliev0, line 561: Hmm, we need to think about what to do here. If we call `GetLiveResource()` every time here, then we are effectively making an API call with every reconciliation of the Pipeline. I kind of wonder if we should stop doing any kind of "Get" here and simply get the Pipeline from the previous `reconcile()` call and pass it into here ?

internal/util/metrics/metrics.go
```diff
@@ -92,7 +92,7 @@ var (
 	// pipelinesHealth indicates whether the pipeline rollouts are healthy (from k8s resource perspective).
 	pipelinesHealth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
juliev0, line 93: should we change the name fo `pipelineRolloutsHealth`?

internal/util/kubernetes/unstructured_util.go
chandankumar4: How about having it as `gvk schema.GroupVersionKind` instead of `obj *GenericObject` ? like below
```
ListResources(ctx context.Context, c client.Client, gvk schema.GroupVersionKind, opts ...client.ListOption)
```

juliev0: perfect!

juliev0:
Do you want to update the name of this PR since it's doing a lot more than it used to?

internal/controller/pipelinerollout_controller.go
chandankumar4: Instead of getting it from previous `reconcile()`, how about getting it from cache instead? because getting from previous `reconsile` is neither or less same as getting from cache.
I ran the unit test with it and don't see any failure.

chandankumar4: Yaa, it might be a problem!

chandankumar4: Updated

chandankumar4: After change, Unit test and E2E is started getting NPE, I tried to look but didn't found the root issue in time-bound, So thinking to look it as a separate issue. what do you think?

juliev0: Do you mean to leave the code making the K8S API call here for now? And then doing a follow up to avoid making the K8S call?

chandankumar4: Yes, But still one Unit test case if failing, which i'm fixing

juliev0: okay, let me know when ready

juliev0: and just add a "TODO" here to indicate that it will be changed to not call the K8S API

cmd/main.go
chandankumar4: I have reverted this change, as it was causing test case failure somehow, I'll raise followup PR for it.

internal/controller/pipelinerollout_controller.go
chandankumar4: Fixed the test cases

PR #335, opened 2024-10-15
feat: Pipeline No Downtime Upgrade Clean Up

### LINKED ISSUE #334: Clean up old pipeline when drained

xdevxy:
Fixes #334 

### Modifications
This PR adds the logic for draining the old pipeline and delete drained pipeline during clean up.

### Verification

`make test`
Verified in local cluster with edit the pipeline rollout spec and saw the old pipeline gets drained and deleted.

### DIFF
--- a/internal/controller/pipelinerollout_progressive.go
+++ b/internal/controller/pipelinerollout_progressive.go
@@ -2,10 +2,7 @@ package controller
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	numaflowv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/rest"
@@ -100,6 +97,7 @@ func (r *PipelineRolloutReconciler) processUpgradingPipelineStatus(
 	pipelinePhase := pipelineStatus.Phase
 	if pipelinePhase == numaflowv1.PipelinePhaseFailed {
 		// Mark the failed new pipeline recyclable.
+		// TODO: pause the failed new pipeline so it can be drained.
 		err = r.updatePipelineLabel(ctx, r.restConfig, existingUpgradingPipelineDef, string(common.LabelValueUpgradeRecyclable))
 		if err != nil {
 			return false, err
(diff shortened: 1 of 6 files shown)


internal/controller/pipelinerollout_progressive.go
```diff
@@ -155,5 +165,86 @@ func (r *PipelineRolloutReconciler) updatePipelineLabel(
...
+	if len(status.Conditions) == 0 {
+		return false
+	}
+	for _, c := range status.Conditions {
+		if c.Status != metav1.ConditionTrue {
+			return false
+		}
+	}
+	return true
 }
```
juliev0, line 250: nit: do you mind if we move `parsePipelineStatus()` and `isPipelineReady()` into pipelinerollout_controller.go since those will be potentially useful outside of progressive strategy?

internal/controller/pipelinerollout_progressive.go
```diff
@@ -90,20 +92,25 @@ func (r *PipelineRolloutReconciler) processUpgradingPipelineStatus(
...
 	if err != nil {
 		return false, fmt.Errorf("failed to parse Pipeline Status from pipeline CR: %+v, %v", existingUpgradingPipelineDef, err)
 	}
 
-	pipelinePhase := numaflowv1.PipelinePhase(pipelineStatus.Phase)
+	pipelinePhase := pipelineStatus.Phase
 	if pipelinePhase == numaflowv1.PipelinePhaseFailed {
-		err = r.updatePipelineLabel(ctx, r.restConfig, existingUpgradingPipelineDef, "")
+		// Mark the failed new pipeline recyclable.
+		err = r.updatePipelineLabel(ctx, r.restConfig, existingUpgradingPipelineDef, string(common.LabelValueUpgradeRecyclable))
```
juliev0, line 103: I am seeing the value of not deleting the Failed one, so the user can observe what happened after the fact (check Pipeline logs, etc). 

I am okay to merge, but maybe we change that in the future or make it configurable? what are your thoughts?

juliev0: oh, I see. Is it simply marked "recyclable", but then nothing causes it to pause and drain so it never actually gets deleted?

internal/controller/pipelinerollout_progressive.go
```diff
@@ -90,20 +92,25 @@ func (r *PipelineRolloutReconciler) processUpgradingPipelineStatus(
...
 		if err != nil {
 			return false, err
 		}
 		pipelineRollout.Status.MarkPipelineProgressiveUpgradeFailed("New Pipeline Failed", pipelineRollout.Generation)
 		return false, nil
 	} else if pipelinePhase == numaflowv1.PipelinePhaseRunning {
+		if !isPipelineReady(pipelineStatus.Status) {
+			//continue (re-enqueue)
+			return false, nil
+		}
```
juliev0, line 113: I know this has the ambiguity of the Child Conditions and whether they mean "Progressing" or actually unhealthy, which we can't really do anything about. So, I guess if they are actually unhealthy, then the PipelineRollout will simply remain in a Pending state with this upgrading Pipeline, and then the user can figure out how to resolve it. Does that sound right?

xdevxy: that's right

internal/controller/pipelinerollout_progressive.go
xdevxy: that's right. I forgot to pause the pipeline. I will add a todo here. so your guys can pick it up.

xdevxy: its a simple change however the reason why I didn't add it, is I didn't cover the test for it.

juliev0: I do actually think that we may not want to necessarily delete the Failed pipeline, as it could be good to diagnose after the fact what happened. But then I guess somebody will need the way to perform the delete maybe through ArgoCD Action. 
Maybe this will be configurable ultimately.

juliev0 (approved):

PR #336, opened 2024-10-15
test: change MonoVertexRollout update test to be more stable

### LINKED ISSUE #314: MonoVertex pods often unhealthy in e2e test
**Describe the bug**
I'm not sure that this is any issue on our side, but would be worth investigating. Ultimately, could be something to hand over to Numaflow team to look at after some analysis on our side.

I was seeing that the MonoVertex pod was in a crash loop at the very end of the e2e test. I'm not sure if it's consistent or not, but I've seen it more than once. (Perhaps it's okay and it eventually fixes itself?)

This is the CI log from the test I ran locally:
[ci.log.txt](https://github.com/user-attachments/files/17219806/ci.log.txt)
These are the outputs from `tests/e2e/outputs` directory:
[output.zip](https://github.com/user-attachments/files/17219787/output.zip)

If you look at `outputs/resources/monovertexrollouts/pods` you can see many Pods in there, which seems to indicate that the Pods restarted a lot. 

**To Reproduce**

1. DATA_LOSS_PREVENTION=true make start
2. DATA_LOSS_PREVENTION=true make test-e2e 

I assume this also happens for DATA_LOSS_PREVENTION=false, but I didn't try it.

---

dpadhiar:
Fixes #314 

### Modifications

MonoVertex currently ends up unhealthy in testing since we are updating the source image to an unstable and outdated version that is incompatible with Numaflow. This changes it to update to the most stable Python image instead.

### Verification

Ran `make test-e2e` locally and checked MonoVertex logs to see that it is still healthy.

### DIFF
--- a/tests/e2e/functional_test.go
+++ b/tests/e2e/functional_test.go
@@ -528,7 +528,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		// new MonoVertex spec
 		updatedMonoVertexSpec := monoVertexSpec
-		updatedMonoVertexSpec.Source.UDSource.Container.Image = "quay.io/numaio/numaflow-java/source-simple-source:v0.6.0"
+		updatedMonoVertexSpec.Source.UDSource.Container.Image = "quay.io/numaio/numaflow-python/simple-source:stable"
 		rawSpec, err := json.Marshal(updatedMonoVertexSpec)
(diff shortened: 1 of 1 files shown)


juliev0 (approved):

juliev0:
thanks for fixing!

PR #345, opened 2024-10-21
feat: Generalize progressive logic

juliev0:
### Modifications

This generalizes the Progressive logic so it can at least be used by PipelineRollout and MonoVertexRollout (possibly ISBServiceRollout but more changes may need to be made for that).

It introduces 2 new interfaces:
`progressiveController` - PipelineRolloutController and MonoVertexRolloutController can both implement this
`RolloutObject` - all of our Rollout objects can implement this

The `pipelinerollout_controller_progressive.go` code was replaced with `progressive.go`. Functions in there take the 2 interfaces mentioned above.

For the most part, the logic was kept the same. 2 exceptions:
1. Calls to "Update()" were replaced with "Patch()" as discussed before to reduce resource version conflict errors
2. No longer mark Recyclable Failed Pipelines: this causes the code to perpetually think that it needs to create a new "Upgrading" pipeline (verified that this happens in the old code). Will consider separately the best ways to deal with Failed children.

### Verification
If I create a PipelineRollout, a new pipeline named "mypipeline0" gets created. Then I modify the PipelineRollout spec, a new Pipeline "mypipeline1" gets created. Once it is fully up and running, "mypipeline0" pauses, and once fully drained, it's deleted.


.github/workflows/ci.yaml
```diff
@@ -113,7 +114,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        datalossprevention: ["true", "false"]
+        ppnd: ["true", "false"]
```
juliev0, line 117: this more accurately reflects what we're doing - either testing with "ppnd" or no strategy

internal/controller/pipelinerollout_controller.go
```diff
@@ -685,23 +690,6 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		return false, err
-	}
-	pipelineWithoutDesiredPhaseB, err := pipelineWithoutDesiredPhase(b)
-	if err != nil {
-		return false, err
-	}
-	numaLogger.Debugf("comparing specs: pipelineWithoutDesiredPhaseA=%v, pipelineWithoutDesiredPhaseB=%v\n", pipelineWithoutDesiredPhaseA, pipelineWithoutDesiredPhaseB)
-
-	return !reflect.DeepEqual(pipelineWithoutDesiredPhaseA, pipelineWithoutDesiredPhaseB), nil
-}
```
juliev0, line 703: replaced this with `childNeedsUpdating()` which is one of the functions of the `rolloutObject` interface

internal/controller/pipelinerollout_controller.go
```diff
@@ -844,6 +802,88 @@ func (r *PipelineRolloutReconciler) makePipelineDefinition(
...
+	currentNameCount, found := r.getCurrentChildCount(rolloutObject)
+	if !found {
+		currentNameCount = int32(0)
+		err := r.updateCurrentChildCount(ctx, rolloutObject, int32(0))
+		if err != nil {
+			return int32(0), err
+		}
+	}
+
+	// TODO: why in PipelineRolloutReconciler.calPipelineName() do we only update the Status subresource when it's 0?
```
juliev0, line 848: was wondering why we do the Update in Kubernetes [here](https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout_controller.go#L797) but not [here](https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout_controller.go#L805)

Maybe we always want to increment the `NameCount` in the Status before we create a Pipeline which is based off that value? Otherwise, we can look at the scenario in which we update the Pipeline name first and then fail to update the `NameCount` in Kubernetes, could we create a new Pipeline with the wrong name next time around?

internal/controller/pipelinerollout_controller.go
```diff
@@ -889,15 +929,3 @@ func parsePipelineStatus(obj *kubernetes.GenericObject) (numaflowv1.PipelineStat
...
-	if len(status.Conditions) == 0 {
-		return false
-	}
-	for _, c := range status.Conditions {
-		if c.Status != metav1.ConditionTrue {
-			return false
-		}
-	}
-	return true
-}
```
juliev0, line 903: replaced with generic

.github/workflows/ci.yaml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - "main"
+      - "ged-oct-2024" # todo: remove later
```
juliev0, line 10: need to run CI for all of our GED PRs

internal/controller/pipelinerollout_controller.go
```diff
@@ -761,65 +747,20 @@ func (r *PipelineRolloutReconciler) updatePipelineRolloutStatusToFailed(ctx cont
...
-		if statusUpdateErr != nil {
-			return "", statusUpdateErr
-		}
-	}
-
-	preNameCount := *pipelineRollout.Status.NameCount
-	*pipelineRollout.Status.NameCount++
-
-	return "-" + fmt.Sprint(preNameCount), nil
-}
```
juliev0, line 808: moved into generic code

xdevxy (approved):

PR #347, opened 2024-10-21
feat: allow for multiple instances of the NumaflowController

afugazzotto:
### Modifications

Added instanceID to the NumaflowControllerRollout spec and templated the definitions config to include the instanceID.

### Verification

Manually tested that multiple instances are deployed and "tagged" (resource name suffixed) by the instanceID defined in the NumaflowControllerRollout spec. Added unit tests for func `resolveManifestTemplate`.

### DIFF
--- a/internal/controller/numaflowcontrollerrollout_controller.go
+++ b/internal/controller/numaflowcontrollerrollout_controller.go
@@ -454,6 +454,10 @@ func ownerExists(existingRefs []interface{}, ownerRef map[string]interface{}) bo
 }
 
 func resolveManifestTemplate(manifest string, rollout *apiv1.NumaflowControllerRollout) ([]byte, error) {
+	if rollout == nil {
+		return []byte(manifest), nil
+	}
+
 	tmpl, err := template.New("manifest").Parse(manifest)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse manifest: %v", err)
--- a/internal/controller/numaflowcontrollerrollout_controller.go
+++ b/internal/controller/numaflowcontrollerrollout_controller.go
@@ -463,10 +463,16 @@ func resolveManifestTemplate(manifest string, rollout *apiv1.NumaflowControllerR
 		return nil, fmt.Errorf("unable to parse manifest: %v", err)
 	}
 
+	instanceID := rollout.Spec.Controller.InstanceID
+	instanceSuffix := ""
+	if strings.TrimSpace(instanceID) != "" {
+		instanceSuffix = fmt.Sprintf("-%s", instanceID)
+	}
+
 	data := struct {
-		InstanceID string
+		InstanceSuffix string
 	}{
-		InstanceID: rollout.Spec.Controller.InstanceID,
+		InstanceSuffix: instanceSuffix,
 	}
 
 	var buf bytes.Buffer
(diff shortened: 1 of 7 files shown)


internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -607,8 +639,11 @@ func getDeploymentCondition(status appsv1.DeploymentStatus, condType appsv1.Depl
 // - whether it exists
 // - error if any
 func (r *NumaflowControllerRolloutReconciler) getNumaflowControllerDeployment(ctx context.Context, controllerRollout *apiv1.NumaflowControllerRollout) (*appsv1.Deployment, bool, error) {
+	instaceID := controllerRollout.Spec.Controller.InstanceID
```
afugazzotto, line 642: Not checking for nil, assuming the rollout is always available at this point.

juliev0: I'm fine with that

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -607,8 +639,11 @@ func getDeploymentCondition(status appsv1.DeploymentStatus, condType appsv1.Depl
...
+	numaflowControllerDeploymentInstanceName := fmt.Sprintf("%s%s", NumaflowControllerDeploymentName, instaceID)
```
juliev0, line 643: nit: spelling of "instaceID"

afugazzotto: you have no idea how many times I misspelled that today and still missed some, thanks for catching this

juliev0 (approved):

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -452,6 +453,31 @@ func ownerExists(existingRefs []interface{}, ownerRef map[string]interface{}) bo
...
+	tmpl, err := template.New("manifest").Parse(manifest)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse manifest: %v", err)
+	}
+
+	data := struct {
+		InstanceID string
+	}{
+		InstanceID: rollout.Spec.Controller.InstanceID,
+	}
```
juliev0, line 470: nit: So, what i had in mind was this:
the templates themselves don't include "if" statements in them. Something like this:
`name: numaflow-dex-server{{ .InstanceSuffix }}`

And up here `ResourceSuffix` can either be set to `"-" + rollout.Spec.Controller.InstanceID` if it's defined and otherwise be set to `""`. What do you think? I think it would simplify the manifests some.

juliev0:
Also, just thinking - it may be easy to miss a reference? I wonder if you can verify that Controller and Server are working fine at reconciling/showing the resources.

internal/controller/numaflowcontrollerrollout_controller.go
afugazzotto: I opted for this solution to allow us to just change the manifest without having to do code changes. I can move this logic to the code side and add 2 if-statements if you prefer that.

juliev0: Yeah, I would prefer to make the manifests simpler. Thanks for the flexibility!

juliev0 (approved):

PR #349, opened 2024-10-22
feat: propagate annotations/labels set in rollout to the child resource

dpadhiar:
### Modifications

If annotations or labels are set in the `spec.crd.metadata` of the rollout, they will be carried over to the child resource.

### Verification

Creating an ISBService, Pipeline and MonoVertex rollout with annotations and labels and verifying the resource had the correct annotations and labels as well as any previous ones.
 
 This was checked under no strategy, PPND and Progressive all set using `make start`


internal/controller/isbservicerollout_controller.go
```diff
@@ -261,6 +269,16 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 
 	} else {
 		// Object already exists
+		// copy rollout annotations and labels to child resource
+		for val, key := range isbServiceRollout.Spec.InterStepBufferService.Annotations {
+			existingISBServiceDef.Annotations[val] = key
+		}
+		newISBServiceDef.Annotations = existingISBServiceDef.Annotations
+		for val, key := range isbServiceRollout.Spec.InterStepBufferService.Labels {
+			existingISBServiceDef.Labels[val] = key
```
juliev0, line 278: so, "existingISBServiceDef" is supposed to mean the isbsvc that exists in Kubernetes, whereas "newISBServiceDef" is supposed to mean the definition we intend to write. We don't want to update the "existingISBServiceDef" (same applies to all rollouts)

dpadhiar: Ack, I can copy the old labels/annotations first and handle them the same way I handle the case where the resource doesn't exist:
```newISBServiceDef.Annotations = existingISBServiceDef.Annotations
		for val, key := range isbServiceRollout.Spec.InterStepBufferService.Annotations {
			newISBServiceDef.Annotations[val] = key
		}
		newISBServiceDef.Labels = existingISBServiceDef.Labels
		for val, key := range isbServiceRollout.Spec.InterStepBufferService.Labels {
			newISBServiceDef.Labels[val] = key
		}
```

internal/controller/isbservicerollout_controller.go
```diff
@@ -236,6 +236,8 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            isbServiceRollout.Name,
 			Namespace:       isbServiceRollout.Namespace,
+			Labels:          map[string]string{},
+			Annotations:     map[string]string{},
```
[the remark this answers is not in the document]
juliev0: (same for other rollouts)

juliev0 (approved):

PR #350, opened 2024-10-22
feat: Incorporate metadata into the USDE check

juliev0:
### Modifications

This updates the USDE `ResourceNeedsUpdating()` call to compare not just specs but also metadata (Labels and Annotations).

Note that this code will be improved later, but for now what it does is:
- consider just the "Instance" annotation as one which requires PPND or Progressive (I think PPND and Progressive both apply if you're changing which Controller to use)
- if any other Label or Annotation changes, return DirectApply

Then take the most conservative strategy returned from comparing `Metadata` and from comparing `Spec`.

Note that the caller merges the new and existing child definitions together to create the updated "new definition" (i.e. existing Labels/Annotations with overriding Labels/Annotations coming from the Rollout definition). Therefore, it should work to compare that "new definition" with the existing definition from Kubernetes.

### Verification

I've tested this through unit test only. Once Dillen incorporates the Labels and Annotations from the Rollout specs, this can be tested more thoroughly.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -67,6 +67,9 @@ const (
 	// LabelValueUpgradeRecyclable is the label value indicating that the resource managed by a NumaRollout is recyclable
 	// after an upgrade.
 	LabelValueUpgradeRecyclable UpgradeState = "recyclable"
+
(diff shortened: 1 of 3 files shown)


afugazzotto (approved):

dpadhiar (commented):
LGTM

PR #354, opened 2024-10-23
feat: Propagate labels and annotations from Pipelinerollout specified metadata into pipeline

juliev0:
### Modifications

Propagate labels and annotations from `Pipelinerollout`-specified metadata into pipeline.

For `PipelineRollout`, we need to make sure that both the "promoted" pipeline and the "upgrading" pipeline include these Labels and Annotations.

For each of these Pipelines, we always store the "new" (desired) definition and the "existing" definition that we get from Kubernetes. Our process is to take the "new" definition and merge it with the existing definition to create a more complete "new" definition (i.e. if somebody besides us set a Label or an Annotation, we want to maintain that, but we do want to replace the Spec with the one from the Rollout, and we also want to include any of our Labels and Annotations). 

### Verification

Ran `PROGRESSIVE=true make start`. Saw that a new Pipeline `pipeline-1` was created when I updated the instance ID annotation. Saw that if I updated a different Label or Annotation, then the Progressive path was not taken - just "Direct Apply" path, and the Label/Annotation was replaced.


internal/controller/pipelinerollout_controller.go
```diff
@@ -687,22 +704,22 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	util.RemovePaths(specAsMap, comparisonExcludedPaths, ".")
 	// if "lifecycle" is there and empty, remove it
-	lifecycleMap, found := pipelineAsMap["lifecycle"].(map[string]interface{})
+	lifecycleMap, found := specAsMap["lifecycle"].(map[string]interface{})
 	if found && len(lifecycleMap) == 0 {
-		util.RemovePaths(pipelineAsMap, []string{"lifecycle"}, ".")
+		util.RemovePaths(specAsMap, []string{"lifecycle"}, ".")
 	}
-	return pipelineAsMap, nil
+	return specAsMap, nil
```
juliev0, line 722: I am using the same function in `MonoVertexRolloutReconciler` (in my other not-yet-merged branch that I forked this from) which is why I renamed this. Ideally, maybe it should be moved to another file.

internal/controller/isbservicerollout_controller.go
```diff
@@ -287,13 +287,19 @@ func (r *ISBServiceRolloutReconciler) getChildTypeString() string {
 }
 
 // take the existing ISBService and merge anything needed from the new ISBService definition
-func mergeISBService(existingISBService, newISBService *kubernetes.GenericObject) *kubernetes.GenericObject {
+func (r *ISBServiceRolloutReconciler) merge(existingISBService, newISBService *kubernetes.GenericObject) *kubernetes.GenericObject {
 	resultISBService := existingISBService.DeepCopy()
 	resultISBService.Spec = *newISBService.Spec.DeepCopy()
 
+	if resultISBService.Annotations == nil {
```
juliev0, line 294: prevent panic if nil

internal/controller/pipelinerollout_controller.go
```diff
@@ -69,13 +68,20 @@ var (
...
+	Lifecycle                  Lifecycle `json:"lifecycle,omitempty"`
 }
 
-type lifecycle struct {
+func (pipeline PipelineSpec) getISBSvcName() string {
+	if pipeline.InterStepBufferServiceName == "" {
+		return "default"
+	}
+	return pipeline.InterStepBufferServiceName
+}
```
juliev0, line 82: This function existed previously but was not a function of `PipelineSpec` - thought this made more sense

internal/controller/isbservicerollout_controller.go
dpadhiar: Good catch!

dpadhiar (approved):
LGTM

afugazzotto (approved):

PR #357, opened 2024-10-23
feat: Progressive Rollout strategy for MonoVertex

juliev0:
Fixes #TODO

### Modifications

Incorporated Progressive rollout strategy for MonoVertexRollout. Like Pipeline, it will create a `mono-vertex-0` and a `mono-vertex-1` and follow all of the same core logic.

If we're in "Progressive" mode, it will do Progressive strategy if either spec or metadata is changed, as we don't have any "excludedFields". If we're in PPND, it just does basic apply since there's no concern for data loss.

### Verification

Created MonoVertexRollout and then edited it. Observed original one paused and deleted. (pause is very fast and I couldn't find a way for it to be at all slow by increasing `generator.rpu`)

Also verified that Progressive still works for PipelineRollout.


internal/controller/pipelinerollout_controller.go
```diff
@@ -557,14 +557,11 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
...
-			err = garbageCollectChildren(ctx, pipelineRollout, r, r.restConfig)
-			if err != nil {
-				return err
-			}
-		}
+	}
+	// clean up recyclable pipelines
+	err = garbageCollectChildren(ctx, pipelineRollout, r, r.restConfig)
+	if err != nil {
+		return err
```
juliev0, line 564: it seems to me that we can do this at any time regardless of whether we're in the middle of a strategy or not

internal/controller/progressive.go
```diff
@@ -242,7 +246,7 @@ func recycle(ctx context.Context,
 		return err
 	}
 	if isDrained {
-		err = kubernetes.DeleteCR(ctx, restConfig, childObject, "pipelines")
+		err = kubernetes.DeleteCR(ctx, restConfig, childObject, childPluralName)
```
juliev0, line 249: accidentally had "pipelines" in there before

dpadhiar (approved):

PR #359, opened 2024-10-24
chore: create better isolation between isbsvc test cases by adding initial UpgradeInProgress value

juliev0:
### Modifications

I wanted to make the test cases within the PPND unit test for isbsvc isolated so that each one is not dependent on the one that occurred before. I noticed that we weren't setting the initial `UpgradeInProgress` value, so when I'd tried to just test one of the cases in the middle it failed, but then it passed when I uncommented out the others. 

Also, made a couple of other minor mods to other PPND tests.

### Verification

Adding initial `UpgradeInProgress` makes it so that I can comment/uncomment them separately (I verified this).

### DIFF
--- a/internal/controller/isbservicerollout_controller_test.go
+++ b/internal/controller/isbservicerollout_controller_test.go
@@ -35,6 +35,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/record"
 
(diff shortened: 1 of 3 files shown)


internal/controller/pipelinerollout_controller_test.go
```diff
@@ -763,7 +762,7 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
 		pipelineRolloutAnnotations     map[string]string
 		existingPipelineDef            numaflowv1.Pipeline
 		initialRolloutPhase            apiv1.Phase
-		initialInProgressStrategy      *apiv1.UpgradeStrategy
+		initialInProgressStrategy      apiv1.UpgradeStrategy
```
juliev0, line 765: simplifying the code a little - no reason this needs to be a pointer

internal/controller/isbservicerollout_controller_test.go
```diff
@@ -273,48 +274,51 @@ func Test_reconcile_isbservicerollout_PPND(t *testing.T) {
...
 	testCases := []struct {
-		name                    string
-		newISBSvcSpec           numaflowv1.InterStepBufferServiceSpec
-		existingISBSvcDef       *numaflowv1.InterStepBufferService
-		existingStatefulSetDef  *appsv1.StatefulSet
-		existingPipelineRollout *apiv1.PipelineRollout
-		existingPipeline        *numaflowv1.Pipeline
-		existingPauseRequest    *bool // was ISBServiceRollout previously requesting pause?
-		expectedPauseRequest    *bool // after reconcile(), should it be requesting pause?
-		expectedRolloutPhase    apiv1.Phase
```
juliev0, line 284: not sure why it's marking these entire blocks as changed, all I did was add the `initialInProgressStrategy`

afugazzotto (approved):

dpadhiar (approved):

PR #366, opened 2024-10-28
feat: Merge feature branch from GED; Enable Multiple Numaflow Controllers and enable Pipelines and MonoVertices to use them progressively

juliev0:
This merges the GED branch into main.

It does what's in [this](https://docs.google.com/document/d/1t1sCDaIj-m3bxQuLmTLrE6k_sY83f76UbjzOpKcb_Og/edit?tab=t.0#heading=h.sr4vhpd3l9o2) document.

Enable multiple Numaflow Controller definitions to coexist in the namespace, each the child of a different NumaflowControllerRollout. Enable PipelineRollout, ISBServiceRollout, MonoVertexRollout to reference the Instance-ID through Annotation. 

Enable Progressive rollout for MonoVertexRollout. This includes both when the annotation is changed as well as when any field is changed. 

## Temporary changes that will be changed later

Note that for now in Progressive mode, we create a separate `ISBServiceRollout` - in the future, we will use one `ISBServiceRollout` which will have multiple `isbsvc` children. (We may also do the same for the `NumaflowControllerRollout` itself, if not too complicated.)

## Changes to USDE

We now have the idea that Labels and Annotations can be used by the USDE to determine if a given change should result in "DirectApply" or "PPND/Progressive" - for now, the "numaflow.numaproj.io/instance" annotation is the only one considered to require "PPND/Progressive".

## Changes to e2e test

Note with this change we run `make start` either as:
- `PPND=true make start` (do ppnd)
- `PROGRESSIVE=true make start` (do progressive)
- `make start` (no strategy)

To run the e2e test either with PPND or not (we don't have e2e for Progressive yet), do:
- `PPND=true make test-e2e` or
- `PPND=false make test-e2e` or `make test-e2e` (no strategy)

### Verification

I tried running 2 Numaflow Controllers with different instance IDs, 2 isbsvc on the different numaflow controllers, and 1 MV and 1 Pipeline originally on the first Numaflow Controller. Then I switched the MV and Pipeline to the new one and confirmed it worked.


chandankumar4 (approved):
looks good

PR #375, opened 2024-10-30
fix: Metrics rename

juliev0:
### Modifications

Renamed the names used in the code as well as the metric names for a few things:
- our counters were named after the child (i.e. `pipeline`) when in the code they really seemed to represent the counts of Rollouts
- changed the names of the counters representing "syncs" and "sync errors" to better describe what they represent


chandankumar4 (approved):
Thanks

PR #381, opened 2024-10-31
chore: print message when configmaps change; remove outdated fields from configmap that's run locally

juliev0:
### Modifications

I noticed we weren't printing anything when the configmaps change. I'm guessing people decided not to because there's a cyclical dependency caused if you try to use the logger, which depends on the `config` module. I decided to just add `fmt.Printf` statements in that case.

Second, I also noticed there were some outdated fields in the configmap we're applying when we run `make start` (from original Numaplane), so I removed them.

### Verification

Updated ConfigMaps and verified I saw the changes reflected in the log.

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -95,6 +95,8 @@ func (cm *NumaflowControllerDefinitionsManager) UpdateNumaflowControllerDefiniti
 	// Add or update the controller definition config based on a version
 	for _, controller := range config.ControllerDefinitions {
 		cm.rolloutConfig[controller.Version] = controller.FullSpec
+
+		fmt.Printf("Added/Updated Controller definition Config, version: %s\n", controller.Version)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/controller/config/config.go
```diff
@@ -95,6 +95,8 @@ func (cm *NumaflowControllerDefinitionsManager) UpdateNumaflowControllerDefiniti
 	// Add or update the controller definition config based on a version
 	for _, controller := range config.ControllerDefinitions {
 		cm.rolloutConfig[controller.Version] = controller.FullSpec
+
+		fmt.Printf("Added/Updated Controller definition Config, version: %s\n", controller.Version)
```
chandankumar4, line 99: shouldn't we use the logger from util package, instead of fmt ?

juliev0: See the PR description. There’s a circular dependency if you try to do that since “logger” module imports “config” module. I tried a little to move directories around to avoid that but it seemed to be complicated. If someone wants to revisit it later, they can.

chandankumar4 (approved):

PR #387, opened 2024-11-04
feat: distinguish pause metrics between Paused and Pausing phases

### LINKED ISSUE #307: Pause Metric should distinguish between Pausing and Paused
# Summary

The metric we have for "Pipeline paused" includes the entire interval between when the Pipeline started pausing and when it finished being Paused. 

We should have a way to distinguish Pipelines which stayed in "Pausing" for a really long time from Pipelines which were in "Paused" for a very long time.

---

dpadhiar:
Fixes #307 

### Modifications

Splits the current PipelinesPausing metric into two separate metrics: one for when the Pipeline is in the Pausing phase and another for when it in the Paused phase. By splitting these metrics, we will be able to accurately track if a Pipeline is stuck in a certain phase and an issue may be occurring as opposed to just being in the Paused state like we assume.

We also now only track system pauses as opposed to differentiating between user and system pauses. For user pauses, we expect to not to have to alert on users leaving their Pipelines paused. By tracking system pauses, we will be able to see errors occurring with the PPND strategy.

Within discussion for this implementation, we also will not add MonoVertex pausing metrics as there is no PPND strategy for the CRD at this time.

```
# HELP numaflow_pipeline_system_paused_seconds Duration a pipeline is paused for
# TYPE numaflow_pipeline_system_paused_seconds gauge
numaflow_pipeline_system_paused_seconds{intuit_alert="true",name="my-pipeline",namespace="example-namespace"} 0
# HELP numaflow_pipeline_system_pausing_seconds Duration a pipeline is pausing for
# TYPE numaflow_pipeline_system_pausing_seconds gauge
numaflow_pipeline_system_pausing_seconds{intuit_alert="true",name="my-pipeline",namespace="example-namespace"} 0
```

### Verification

Testing locally with PPND causing changes to a PipelineRollout and changing the desiredPhase to show user pausing.

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -646,32 +646,41 @@ func (r *PipelineRolloutReconciler) setChildResourcesPauseCondition(pipelineRoll
 		}
 		reason := fmt.Sprintf("Pipeline%s", string(pipelinePhase))
 		msg := fmt.Sprintf("Pipeline %s", strings.ToLower(string(pipelinePhase)))
-		r.updatePauseMetric(pipelineRollout)
+		r.updatePauseMetric(pipelineRollout, pipelinePhase)
 		pipelineRollout.Status.MarkPipelinePausingOrPaused(reason, msg, pipelineRollout.Generation)
 	} else {
 		// only set EndTime if BeginTime has been previously set AND EndTime is before/equal to BeginTime
 		// EndTime is either just initialized or the end of a previous pause which is why it will be before the new BeginTime
 		if (pipelineRollout.Status.PauseStatus.LastPauseBeginTime != metav1.NewTime(initTime)) && !pipelineRollout.Status.PauseStatus.LastPauseEndTime.After(pipelineRollout.Status.PauseStatus.LastPauseBeginTime.Time) {
 			pipelineRollout.Status.PauseStatus.LastPauseEndTime = metav1.NewTime(time.Now())
-			r.updatePauseMetric(pipelineRollout)
+			r.updatePauseMetric(pipelineRollout, pipelinePhase)
 		}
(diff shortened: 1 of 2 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -646,34 +646,43 @@ func (r *PipelineRolloutReconciler) setChildResourcesPauseCondition(pipelineRoll
...
 
 	var pipelineSpec PipelineSpec
 	_ = json.Unmarshal(pipelineRollout.Spec.Pipeline.Spec.Raw, &pipelineSpec)
 
 	timeElapsed := time.Since(pipelineRollout.Status.PauseStatus.LastPauseBeginTime.Time)
 	if r.isSpecBasedPause(pipelineSpec) {
-		r.customMetrics.PipelinePausedSeconds.WithLabelValues(pipelineRollout.Namespace, pipelineRollout.Name, "user_pause").Set(timeElapsed.Seconds())
+		r.setMetric(pipelineRollout.Namespace, pipelineRollout.Name, float64(0), float64(0))
+	} else if pipelinePhase == numaflowv1.PipelinePhasePaused {
+		r.setMetric(pipelineRollout.Namespace, pipelineRollout.Name, timeElapsed.Seconds(), float64(0))
```
juliev0, line 672: this time doesn't represent "paused" time - it represents "pausing+paused" time, right?

dpadhiar: Ack, this is the case since our pause status only track start to end for the pause and unpause request.

We would probably need to include a `PausingBeginTime` and `PausingEndTime` to Pause status to fully accurately report this info.

dpadhiar:
Converted to Draft and marked as WIP as we discussed on Slack to update our pause status to also accurately track Pausing time as opposed to what we track currently which is the time the Pause request occurs to when it finishes.

PR #389, opened 2024-11-04
feat: add option to override defaultStrategy with noStrategy

### LINKED ISSUE #386: Be able to override default Upgrade Strategy with "no strategy"
# Summary

Currently, we [only allow](https://github.com/numaproj/numaplane/blob/main/internal/controller/config/user_config.go#L39-L46) a user to specify "pause-and-drain" or "progressive" in their user namespace config. We need to be able to set "no strategy" as well. 

The thing is that we need to be able to distinguish empty string from "no strategy", so we may need to change the constant for [NoStrategyID](https://github.com/numaproj/numaplane/blob/v0.7.1/internal/controller/config/user_config.go#L13). 

---

dpadhiar:
Fixes #386 

### Modifications

Allows user to specify `no-strategy` as a valid option for upgrade strategy in the user namespace config.

Previously we viewed empty string as specifying no-strategy but now we look for this to be specifically to override the current setting.

### Verification

Running `PPND=true make start` and testing with a namespace config that says `no-strategy`

Watched that the Pipeline we updated did not pause when changes were applied to the rollout.

### DIFF
--- a/internal/usde/usde.go
+++ b/internal/usde/usde.go
@@ -190,7 +190,7 @@ func getDataLossUpggradeStrategy(ctx context.Context, namespace string) (apiv1.U
 		return apiv1.UpgradeStrategyPPND, nil
 	case config.ProgressiveStrategyID:
 		return apiv1.UpgradeStrategyProgressive, nil
-	case config.NoStrategyID:
+	case config.NoStrategyID, "":
 		return apiv1.UpgradeStrategyApply, nil
 	default:
 		return apiv1.UpgradeStrategyError, fmt.Errorf("invalid Upgrade Strategy: %v", userUpgradeStrategy)
--- a/internal/controller/config/user_config.go
+++ b/internal/controller/config/user_config.go
@@ -10,7 +10,7 @@ type USDEUserStrategy string
 const (
 	ProgressiveStrategyID USDEUserStrategy = "progressive"
 	PPNDStrategyID        USDEUserStrategy = "pause-and-drain"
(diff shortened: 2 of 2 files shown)


internal/usde/usde.go
```diff
@@ -207,7 +207,11 @@ func GetUserStrategy(ctx context.Context, namespace string) (config.USDEUserStra
 		if !namespaceConfig.UpgradeStrategy.IsValid() {
 			numaLogger.WithValues("upgrade strategy", namespaceConfig.UpgradeStrategy).Warnf("invalid Upgrade strategy for namespace %s", namespace)
 		} else {
-			userUpgradeStrategy = namespaceConfig.UpgradeStrategy
+			if namespaceConfig.UpgradeStrategy == "" {
+				userUpgradeStrategy = "no-strategy"
+			} else {
+				userUpgradeStrategy = namespaceConfig.UpgradeStrategy
+			}
```
juliev0, line 214: The case I was worried about was the case of `namespaceConfig == nil` and `config.GetConfigManagerInstance().GetUSDEConfig().DefaultUpgradeStrategy==""`.  I wanted to make sure we use `config.NoStrategyID` in that case.

juliev0: In the case that the `namespaceConfig.UpgradeStrategy == ""`, that should mean that they prefer to use the `config.GetConfigManagerInstance().GetUSDEConfig().DefaultUpgradeStrategy`

Issue #143, opened 2024-07-25
Refactor: Avoid too many conversions between GenericObject, Unstructured, PipelineSpec


juliev0:
# Summary

I'm thinking maybe we could at the beginning of reconciliation create a struct which can include all the various forms of our current child object and our desired object. Then we could pass that into the various functions, without having to repeatedly do the same translations over and over. 

For the Pipeline, we sometimes use:

- GenericObject
- [PipelineSpec ](https://github.com/numaproj/numaplane/blob/v0.5.0/internal/controller/pipelinerollout_controller.go#L68), which contains just the fields we need that should not ever change
- [GenericStatus](https://github.com/numaproj/numaplane/blob/v0.5.0/internal/util/kubernetes/unstructured_util.go#L34) 
- we don't have a `PipelineStatus` at the moment, but I think we will need it for fields that are not in `GenericStatus`
- `map[string]interface{}`  

For the isbsvc and monovertex, we could theoretically do the same.

For each of them, at the beginning of the reconciliation we create a `GenericObject` from the existing object that we find in Kubernetes and we create another `GenericObject` from the new one based on the Rollout. Then we make modifications to it for example to update the `desiredPhase` field. 

For both "new" and "existing", I'd like to be able to create a single Object at the beginning that stores any of the forms we may need. And if I need to update the "new" one, it should update all of the forms. 

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

chandankumar4:
I'm checking the feasibility to completely eliminate the `kubernetes.GenericObject` struct and directly work on `unstructured.Unstructured`, So that we don't need any conversion in between. If it's not possible then probably I'll work on design to make the conversion in more efficient way. How it's sound?

PR #401, opened 2024-11-11
fix: Refactor isbservice_controller to eliminate use of GenericObject

chandankumar4:
Fixes
ISBService Controller:  https://github.com/numaproj/numaplane/issues/143

### Modifications

Removed the usages of Generic Object struct and use Unstructured.

### DIFF
--- a/internal/util/kubernetes/unstructured_util.go
+++ b/internal/util/kubernetes/unstructured_util.go
@@ -51,6 +51,26 @@ func ParseStatus(obj *GenericObject) (GenericStatus, error) {
 	return status, nil
 }
 
+// TODO: This is a temporary function which will be removed once all the controller are migrated to use Unstructured Object
+func ParseStatusUnstructured(obj *unstructured.Unstructured) (GenericStatus, error) {
+	statusRaw, found, err := unstructured.NestedFieldNoCopy(obj.Object, "status")
+	if !found || err != nil {
+		return GenericStatus{}, nil
+	}
+
+	var status GenericStatus
+	objByte, err := json.Marshal(statusRaw)
+	if err != nil {
+		return GenericStatus{}, err
+	}
+	err = json.Unmarshal(objByte, &status)
+	if err != nil {
+		return GenericStatus{}, err
+	}
+
+	return status, nil
+}
+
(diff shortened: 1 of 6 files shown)


internal/util/kubernetes/unstructured_util.go
```diff
@@ -261,6 +281,12 @@ func CreateResource(ctx context.Context, c client.Client, obj *GenericObject) er
 	return c.Create(ctx, unstructuredObj)
 }
 
+// CreateResourceUnstructured creates the resource in the kubernetes cluster
+// TODO: This is a temporary function which will be removed once all the controller are migrated to use Unstructured Object
```
juliev0, line 285: when you say these are "temporary functions", do you mean that the logic for them will still exist, but basically they will be renamed as the original function names? (e.g. `CreateResource()`)

internal/util/kubernetes/unstructured_util.go
```diff
@@ -51,6 +51,26 @@ func ParseStatus(obj *GenericObject) (GenericStatus, error) {
...
+
+	var status GenericStatus
+	objByte, err := json.Marshal(statusRaw)
+	if err != nil {
+		return GenericStatus{}, err
+	}
+	err = json.Unmarshal(objByte, &status)
+	if err != nil {
+		return GenericStatus{}, err
+	}
```
juliev0, line 69: nit: can you use the [util.StructToStruct() ](https://github.com/numaproj/numaplane/blob/main/internal/util/util.go#L11)method here?

juliev0 (approved):

internal/util/kubernetes/unstructured_util.go
juliev0: I'm okay if you want to merge this PR and do this as a follow up

juliev0: Actually, in the interest of moving forward, I think I'll just merge this so I can merge main into my PR and resolve the conflicts.

internal/util/kubernetes/unstructured_util.go
chandankumar4: Right

Issue #407, opened 2024-11-13
USDE: specify "data loss" fields 


juliev0:
# Summary

Update the existing USDE code to specify "data loss" fields.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
I believe this should be closed now

PR #410, opened 2024-11-14
feat: USDE with data loss fields config

### LINKED ISSUE #330: USDE - if excludedPath is not already present/set in original spec, upgrade decision result will be incorrect
**Describe the bug**

If an excludedPath or path to a field does not currently exist in the original spec of a Rollout object, the USDE method `ResourceNeedsUpdating` will not return the strategy `DirectApply` like is expected and instead will return `PauseAndDrain`.

Note that excluding the parent field of a path does currently work.

**To Reproduce**

1. Create default NumaflowController, ISBService and PipelineRollouts
2. Set an excludedPath for one of the objects ie: 
```
  isbServiceSpecExcludedPaths: |
    - "jetstream.containerTemplate.resources.limits"
```
3. Watch the Pipelines.
4. Update the Rollout with a new value in the excludedPath. In this case, I updated the limits of the ISBServiceRollout.
5. Note that the Pipeline pauses even though we are updating an excludedPath. Confirm by checking controller logs for `upgrade decision result` log statement.

**Expected behavior**

We expect the Pipeline to not pause and the resulting upgrade strategy to be DirectApply.

---

afugazzotto:
Fixes #330 #407 

### Modifications

Changed USDE logic from using a list of direct-apply fields to a list of data-loss fields.

### Verification

Updated unit tests, executed e2e tests, manual tests.

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -1482,12 +1482,12 @@ metadata:
 apiVersion: v1
 data:
   defaultUpgradeStrategy: pause-and-drain
-  isbServiceSpecExcludedPaths: |
-    - "jetstream.containerTemplate.resources.limits"
-  pipelineSpecExcludedPaths: |
-    - "lifecycle"
-    - "limits"
-    - "watermark"
+  isbServiceSpecDataLossFields: |
+    - path: spec
+      includeSubfields: true
+  pipelineSpecDataLossFields: |
+    - path: spec
+      includeSubfields: true
 kind: ConfigMap
 metadata:
   labels:
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -1483,10 +1483,55 @@ apiVersion: v1
 data:
   defaultUpgradeStrategy: pause-and-drain
   isbServiceSpecDataLossFields: |
-    - path: spec
-      includeSubfields: true
+    - path: spec.jetstream.version
+    - path: spec.redis.native.version
+    - path: spec.redis.external.url
+    - path: spec.redis.external.sentinelUrl
+    - path: spec.redis.external.masterName
   pipelineSpecDataLossFields: |
-    - path: spec
+    - path: spec.edges.from
+    - path: spec.edges.to
+    - path: spec.edges.conditions
+      includeSubfields: true
+    - path: spec.interStepBufferServiceName
+    - path: spec.vertices.source.generator
+    - path: spec.vertices.source.kafka
+    - path: spec.vertices.source.http
+    - path: spec.vertices.source.nats
+    - path: spec.vertices.source.jetstream
+    - path: spec.vertices.source.serving
+    - path: spec.vertices.source.udsource.container.image
+    - path: spec.vertices.source.transformer.container.image
+    - path: spec.vertices.source.transformer.container.command
+    - path: spec.vertices.source.transformer.container.args
+    - path: spec.vertices.source.transformer.container.env
+      includeSubfields: true
+    - path: spec.vertices.source.transformer.container.envfrom
+      includeSubfields: true
+    - path: spec.vertices.source.transformer.builtin
+      includeSubfields: true
+    - path: spec.vertices.sink.log
+    - path: spec.vertices.sink.kafka
+    - path: spec.vertices.sink.blackhole
+    - path: spec.vertices.sink.udsink.container.image
+    - path: spec.vertices.sink.udsink.container.command
+    - path: spec.vertices.sink.udsink.container.args
+    - path: spec.vertices.sink.udsink.container.env
+      includeSubfields: true
+    - path: spec.vertices.sink.udsink.container.envfrom
+      includeSubfields: true
+    - path: spec.vertices.udf.container.image
+    - path: spec.vertices.udf.container.command
+    - path: spec.vertices.udf.container.args
+    - path: spec.vertices.udf.container.env
+      includeSubfields: true
+    - path: spec.vertices.udf.container.envfrom
+      includeSubfields: true
+    - path: spec.vertices.udf.builtin
+      includeSubfields: true
+    - path: spec.sideInputs
+      includeSubfields: true
+    - path: spec.templates
       includeSubfields: true
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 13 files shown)


config/manager/usde-config.yaml
```diff
@@ -6,9 +6,54 @@ metadata:
...
+    - path: spec.sideInputs
+      includeSubfields: true
+    - path: spec.templates
+      includeSubfields: true
+  isbServiceSpecDataLossFields: |
+    - path: spec.jetstream.version
+    - path: spec.redis.native.version
+    - path: spec.redis.external.url
+    - path: spec.redis.external.sentinelUrl
+    - path: spec.redis.external.masterName
```
afugazzotto, line 59: @vigith and @whynowy could you please verify that these lists are correct and sufficient to cover all data loss cases?

juliev0: To clarify, the `includeSubfields` value means "if true, we consider it at risk of data loss if the field or any of its children is being modified; if false, we consider it at risk of data loss if the field is modified but not if just its children are"

tests/e2e/functional_test.go
```diff
@@ -151,7 +151,7 @@ var (
 	ISBServiceSpecExcludedField = numaflowv1.InterStepBufferServiceSpec{
 		Redis: nil,
 		JetStream: &numaflowv1.JetStreamBufferService{
-			Version: "2.9.8",
+			Version: updatedJetstreamVersion,
 			Persistence: &numaflowv1.PersistenceStrategy{
 				VolumeSize: &volSize,
 			},
```
juliev0, line 157: nit: do you mind updating this file to change the use of the word "excluded" to "direct apply" or something like that?

internal/usde/usde_test.go
```diff
@@ -372,14 +385,31 @@ func Test_ResourceNeedsUpdating(t *testing.T) {
...
+			}(),
+			usdeConfig: config.USDEConfig{
+				DefaultUpgradeStrategy:       config.PPNDStrategyID,
+				PipelineSpecDataLossFields:   []config.SpecDataLossField{{Path: "spec.vertices.source.something"}},
+				ISBServiceSpecDataLossFields: []config.SpecDataLossField{{Path: "spec.jetstream.containerTemplate.resources.limits", IncludeSubfields: true}},
+			},
+			namespaceConfig:       &config.NamespaceConfig{UpgradeStrategy: "pause-and-drain"},
+			expectedNeedsUpdating: true,
+			expectedStrategy:      apiv1.UpgradeStrategyPPND,
+		},
```
juliev0, line 412: Do you have a test in here which includes the field specified in the new definition and not specified in the original definition? (or the other way around?) For this test, it looks like it's specified in both the `newDefinition` and `existingDefinition` if I'm not mistaken? By removing it from one or the other, we could test Dillen's bug.

internal/util/util.go
```diff
@@ -203,3 +204,68 @@ func removePath(m map[string]any, pathTokens []string) {
...
 }
+
+// ExtractPath extracts the fields specified by the path (given as slice of strings) from the given data.
+//
+// The function takes two arguments:
+//   - data: The input data, which can be a map, a slice, or a nested combination of both.
+//   - path: A slice of strings representing the path to the desired fields.
+//
+// It returns three values:
+//   - any: The extracted data, which can be a map, a slice, or a single value.
```
juliev0, line 215: I think it can also be `nil` (if not found), right? Which would be common

internal/usde/usde.go
```diff
@@ -67,69 +68,79 @@ func resourceSpecNeedsUpdating(ctx context.Context, newDef *kubernetes.GenericOb
...
+		).Debug("checking data loss field differences")
+
+		areDefFieldsMap := newIsMap && existingIsMap
+
+		// If the current field is not a map or if it is (assumed from the config if IncludeSubfields is true) and the config
+		// says to include comparing the subfields, then compare the fields/maps and, if the fields/maps are different,
+		// a data loss prevention strategy is needed
+		if (dataLossField.IncludeSubfields || !areDefFieldsMap) && !reflect.DeepEqual(newDefField, existingDefField) {
+			return true, upgradeStrategy, nil
+		}
```
juliev0, line 129: I'm confused. Where are we looking for the case of `dataLossField.IncludeSubfields==false` with a map? For example, if we add the field `spec.vertices.source.generator` and it wasn't there before. We want this to return PPND/Progressive, right?

afugazzotto: If we want the behavior you described, then the IncludeSubfields for that path must be set to `true`. This is the assumption.

juliev0: No, we want this to result in "data loss":

before:
```
spec:
  vertices:
     source:
```
after:
```
spec:
  vertices:
     source:
       generator:
```

But we don't want this to result in "data loss":

before:
```
spec:
  vertices:
     source:
       generator:
          key: value1
```

after:
```
spec:
  vertices:
     source:
       generator:
          key: value2
```

That is why `spec.vertices.source.generator` has `IncludeSubfields-=false`

juliev0: spoke offline that the logic is likely correct but hard to read - I can merge this PR and more carefully review it after you modify that

afugazzotto: As we discussed I also added 2 more test cases.

juliev0 (approved):

config/install.yaml
```diff
@@ -1482,12 +1482,57 @@ metadata:
...
+    - path: spec.vertices.udf.container.image
+    - path: spec.vertices.udf.container.command
+    - path: spec.vertices.udf.container.args
+    - path: spec.vertices.udf.container.env
+      includeSubfields: true
+    - path: spec.vertices.udf.container.envfrom
+      includeSubfields: true
+    - path: spec.vertices.udf.builtin
+      includeSubfields: true
+    - path: spec.sideInputs
```
juliev0, line 1532: @whynowy is there a risk of data loss associated with updating `sideInputs`?

config/manager/usde-config.yaml
juliev0: By the way, @vigith and @whynowy - this list is strictly the "data loss" list. We recognize that there are some changes for which PPND will also not be acceptable so we have an issue to later add a separate list which would indicate the fields for which somebody would either have to do Progressive (i.e. "new isbsvc, new pipeline") or otherwise delete and recreate.

PR #412, opened 2024-11-15
chore: update e2e to prove usde changes fixed issue, and add comments to USDE config

juliev0:
### Modifications

Updating the e2e test to prove that [this](https://github.com/numaproj/numaplane/issues/330) issue was fixed.

Also, adding comments to explain the fields in the usde_config.yaml

### DIFF
--- a/config/manager/usde-config.yaml
+++ b/config/manager/usde-config.yaml
@@ -57,6 +57,8 @@ data:
     - path: spec.sideInputs.container.args
     - path: spec.sideInputs.container.env
       includeSubfields: true
+    - path: spec.sideInputs.container.envfrom
+      includeSubfields: true
     - path: spec.templates     # note: this one could be broken down further into just the fields that are at risk
       includeSubfields: true
 
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -1540,6 +1540,8 @@ data:
     - path: spec.sideInputs.container.args
     - path: spec.sideInputs.container.env
       includeSubfields: true
+    - path: spec.sideInputs.container.envfrom
+      includeSubfields: true
     - path: spec.templates     # note: this one could be broken down further into just the fields that are at risk
       includeSubfields: true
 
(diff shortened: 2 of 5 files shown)


config/manager/usde-config.yaml
```diff
@@ -47,11 +52,21 @@ data:
...
       includeSubfields: true
-    - path: spec.sideInputs
+    - path: spec.sideInputs.container.image
+    - path: spec.sideInputs.container.command
+    - path: spec.sideInputs.container.args
+    - path: spec.sideInputs.container.env
       includeSubfields: true
-    - path: spec.templates
+    - path: spec.sideInputs.container.envfrom
+      includeSubfields: true
```
juliev0, line 61: broke all of this out into the individual fields

afugazzotto: Great idea!

config/manager/usde-config.yaml
```diff
@@ -47,11 +52,21 @@ data:
       includeSubfields: true
     - path: spec.vertices.udf.builtin
       includeSubfields: true
-    - path: spec.sideInputs
+    - path: spec.sideInputs.container.image
+    - path: spec.sideInputs.container.command
+    - path: spec.sideInputs.container.args
+    - path: spec.sideInputs.container.env
```
afugazzotto, line 58: Can you also update the e2e test files related to this:
- https://github.com/numaproj/numaplane/blob/main/tests/manifests/special-cases/ppnd/usde-config.yaml
- https://github.com/numaproj/numaplane/blob/main/tests/manifests/special-cases/progressive/usde-config.yaml

juliev0: thanks for catching! I already forgot about that even though we were just discussing it. :) It sounds like a good motivation for moving the `defaultUpgradeStrategy` out!

afugazzotto: Working on that now, should be ready soon

PR #413, opened 2024-11-15
chore: post USDE changes cleanup

afugazzotto:
### Modifications

Removed unused functions, reused/updated unit tests, added more USDE logic comments.

### Verification

Unit tests.


internal/usde/usde.go
```diff
@@ -119,10 +119,14 @@ func resourceSpecNeedsUpdating(ctx context.Context, newDef *kubernetes.GenericOb
...
 
+		// Note: if the data loss field is a map but it is nil in either the new or existing spec,
+		// the related variable ...IsMap will be false even though the underlying field is a
+		// map (since it is nil we do not know if it is truly a map).
+		// Therefore, the areDefFieldsMap will also be false.
 		areDefFieldsMap := newIsMap && existingIsMap
 
-		// If the current field is not a map or if it is (assumed from the config if IncludeSubfields is true) and the config
-		// says to include comparing the subfields, then compare the fields/maps and, if the fields/maps are different,
+		// If the current field is not explicitely a map (or if it is assumed from the config because IncludeSubfields would be true)
```
juliev0, line 128: I know it's a little more verbose, but I think I'd like to change this to read like this pseudocode:

```
if IncludeSubfields:
   is the definition at all different?
else:
   if it's a map, then is one present and the other not?
   else:
      are the keys/values different?
```

If I'm being too annoying with this ask, I am okay to change it in a separate PR :)

juliev0 (approved):

PR #416, opened 2024-11-16
chore: move defaultUpgradeStrategy from USDE config to numaplane-controller-config

afugazzotto:
Fixes #415 

### Modifications

Moved the `defaultUpgradeStrategy` env var from USDE Config to `numaplane-controller-config`.

### Verification

Unit tests and e2e tests.

PR #417, opened 2024-11-16
fix: e2e test to execute both "no-strategy" and "ppnd" again; also improve readability for usde logic

juliev0:
### Modifications

I realized our e2e tests have been testing "ppnd" case for both the case of:
- `PPND=true make start` and
- `PPND=false make start`

That's because the default [controller-config.yaml](https://github.com/numaproj/numaplane/blob/main/config/manager/controller-config.yaml#L11) was changed to set `defaultUpgradeStrategy: pause-and-drain`. Previously, that file was effectively set to "no strategy", and then the ppnd case overrode that file with `defaultUpgradeStrategy: pause-and-drain`.

Note that [this](https://github.com/numaproj/numaplane/pull/416) PR moved that field from the `usde_config.yaml` to the `controller-config.yaml`.

Therefore, since "pause-and-drain" is now the default strategy, I've replaced the "ppnd" manifest override directory with a "no-strategy" manifest override directory, and set `controller-config.yaml` in there.

In addition: I discussed with @afugazzotto in his PR: I was confused about the use of  `!areDefFieldsMap` to mean "are they either not maps, or is one of them nil?" so I wanted to clarify that logic. Since he is on vacation, he asked me to address that, so I've done that here.


Makefile
```diff
@@ -37,14 +37,14 @@ GCFLAGS="all=-N -l"
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.28.0
 
-TEST_MANIFEST_DIR_DEFAULT ?= tests/manifests/default
-TEST_PPND_MANIFEST_DIR ?= tests/manifests/special-cases/ppnd
+TEST_MANIFEST_DIR_PPND ?= tests/manifests/default
+TEST_NOSTRATEGY_MANIFEST_DIR ?= tests/manifests/special-cases/no-strategy
```
juliev0, line 41: now we have a `no-strategy` override directory since "pause-and-drain" is the default strategy

internal/usde/usde.go
```diff
@@ -119,17 +123,23 @@ func resourceSpecNeedsUpdating(ctx context.Context, newDef *kubernetes.GenericOb
...
+			// if it's a map, since we don't care about subfields, we just need to know if it's present in one and not the other
+			if isMap {
+				if !newIsMap || !existingIsMap { // this means that one of them is nil
+					return true, upgradeStrategy, nil
+				}
+			} else {
+				if !reflect.DeepEqual(newDefField, existingDefField) {
+					return true, upgradeStrategy, nil
+				}
+			}
```
juliev0, line 142: discussed with @afugazzotto offline that I would update this logic:  I was confused about the use of `!areDefFieldsMap` to mean "are they either not maps, or is one of them nil?"

tests/manifests/special-cases/no-strategy/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ resources:
   - ../../default
 
 patches:
-  - path: usde-config.yaml
\ No newline at end of file
+  - path: controller-config.yaml
```
juliev0, line 8: no need to override `usde-config.yaml` anymore as it's identical to the default one (this was actually the main motivation for [this](https://github.com/numaproj/numaplane/pull/416) change)

internal/usde/usde.go
juliev0: I do recognize that what I have is more verbose than the original, but I think it's okay because to me, it seems easier to understand.

dpadhiar (approved):

PR #419, opened 2024-11-18
fix: Progressive improvements

juliev0:
### Modifications

This PR updates the existing flow for Progressive. Since we have incorporated Pipeline and MonoVertex into "progressive", it updates these specifically.

The new general logic is this:

- Original child is labeled "promoted". Any child that's upgrading is labeled "upgrading". Any child that's ready to be deleted is marked "recyclable". (all of this stays the same as before)
- if the "upgrading" child is Successful, we label it "promoted" and mark the original one as "recyclable" (this stays the same as before)
- if the "upgrading" child is "Failed", we will leave it up for somebody to debug what went wrong (this is different)
- after the failure has occurred, the user can iterate on their spec. If they submit a modification (even if they restore to the original spec), a new child will be created marked "upgrading" and the previous "upgrading" child will be labeled "recyclable" 

For Pipelines, currently we will enter the Progressive path if either a [data loss field](https://github.com/numaproj/numaplane/blob/progressive-improvements/config/manager/usde-config.yaml#L8) is used, or the Controller Instance annotation is changed. 
For MVX, currently we will enter the Progressive path only if the Controller Instance annotation is changed. 

In the future, we will have a set of fields that will be marked as "progressiveDesired" (basically image changes), and we can perform the progressive path for those as well, for both Pipeline and MVX.

### Verification

For both Pipeline and MonoVertex I ran various sequences of successful, failed children such as "failed"->"successful" and "successful"->"successful" and "successful->failed->failed->successful" and checked both the Rollout Status and the child Label. 

I have updated the unit test for pipelinerollout controller: [Test_processExistingPipeline_Progressive](https://github.com/numaproj/numaplane/blob/progressive-improvements/internal/controller/pipelinerollout/pipelinerollout_controller_test.go#L751) and added one for [MonoVertex](https://github.com/numaproj/numaplane/blob/progressive-improvements/internal/controller/monovertexrollout/monovertexrollout_controller_test.go#L178).

### to-do

We will certainly need to add an e2e test/tests for "progressive", which will better test all of this stuff. 

### Backward compatibility risks?

We want to consider any risks as far as backward compatibility with our existing resources. Our existing resources are using PPND and some clusters are still on "Direct Apply". So, we should consider if there's any risk as far as those strategies. We can look at the general logic that is outside of "Progressive" strategy. I believe as far as the interface to the objects themselves, everything should be the same as before.


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -444,8 +444,7 @@ func (r *ISBServiceRolloutReconciler) isISBServiceUpdating(ctx context.Context,
 
 func (r *ISBServiceRolloutReconciler) GetPipelineList(ctx context.Context, rolloutNamespace string, rolloutName string) (*unstructured.UnstructuredList, error) {
 	gvk := schema.GroupVersionKind{Group: common.NumaflowAPIGroup, Version: common.NumaflowAPIVersion, Kind: common.NumaflowPipelineKind}
-	return kubernetes.ListResources(ctx, r.client, gvk,
-		client.InNamespace(rolloutNamespace),
+	return kubernetes.ListResources(ctx, r.client, gvk, rolloutNamespace,
```
juliev0, line 447: this function now always checks the namespace passed in instead of passing it in as a LabelSelector, which would be easy for a caller to accidentally leave out (also creates more consistency with the other functions in `unstructured_util.go`, which take namespace as parameter). I don't think we have a need currently for listing resources across all namespaces. I guess if we do in the future then we can change this or create separate functions that are more explicit.

internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -584,38 +573,20 @@ func (r *MonoVertexRolloutReconciler) IncrementChildCount(ctx context.Context, r
...
-	// remove lifecycle.desiredPhase field from comparison to test for equality
-	mvWithoutDesiredPhaseA, err := numaflowtypes.WithoutDesiredPhase(from)
-	if err != nil {
-		return false, err
-	}
-	mvWithoutDesiredPhaseB, err := numaflowtypes.WithoutDesiredPhase(to)
-	if err != nil {
-		return false, err
-	}
-	numaLogger.Debugf("comparing specs: mvWithoutDesiredPhaseA=%v, mvWithoutDesiredPhaseB=%v\n", mvWithoutDesiredPhaseA, mvWithoutDesiredPhaseB)
```
juliev0, line 614: I found out from Numaflow team that we don't need to set`lifecycle` field in the code at all - so unlike with "Pipeline", I don't need to explicitly remove it from the comparison. (With "Pipeline", there is a danger that we will falsely see a difference because we set it in the code.)

internal/controller/progressive/progressive.go
```diff
@@ -108,108 +120,230 @@ func makeUpgradingObjectDefinition(ctx context.Context, rolloutObject ctlrcommon
...
+	}
+	return AssessmentResultUnknown, nil
 }
 
 // update the in-memory object with the new Label and patch the object in K8S
 func updateUpgradeState(ctx context.Context, c client.Client, upgradeState common.UpgradeState, childObject *unstructured.Unstructured, rolloutObject ctlrcommon.RolloutObject) error {
-	childObject.SetLabels(map[string]string{common.LabelKeyUpgradeState: string(upgradeState)})
+	labels := childObject.GetLabels()
+	labels[common.LabelKeyUpgradeState] = string(upgradeState)
+	childObject.SetLabels(labels)
```
juliev0, line 346: intention is to add "UpgradeState" to whatever the existing labels are

internal/controller/common/numaflowtypes/monovertex.go
```diff
@@ -25,16 +25,26 @@ import (
...
+func MonoVertexWithoutReplicas(monoVertex *unstructured.Unstructured) (map[string]interface{}, error) {
+	var specAsMap map[string]any
+	if err := util.StructToStruct(monoVertex.Object["spec"], &specAsMap); err != nil {
+		return nil, err
+	}
+	// remove "replicas" field
+	comparisonExcludedPaths := []string{"replicas"}
+	util.RemovePaths(specAsMap, comparisonExcludedPaths, ".")
+	return specAsMap, nil
+}
```
juliev0, line 50: this is the field that Numaflow Controller resets, so we want to ignore it in our comparisons

internal/controller/isbservicerollout/isbservicerollout_controller_test.go
```diff
@@ -106,11 +106,12 @@ func Test_reconcile_isbservicerollout_PPND(t *testing.T) {
...
-			newISBSvcSpec:             createDefaultISBServiceSpec("2.10.3"),
-			existingISBSvcDef:         nil,
-			existingStatefulSetDef:    nil,
-			existingPipelineRollout:   ctlrcommon.CreateTestPipelineRollout(numaflowv1.PipelineSpec{InterStepBufferServiceName: ctlrcommon.DefaultTestISBSvcRolloutName}, map[string]string{}, map[string]string{}),
+			name:                   "new ISBService",
+			newISBSvcSpec:          createDefaultISBServiceSpec("2.10.3"),
+			existingISBSvcDef:      nil,
+			existingStatefulSetDef: nil,
+			existingPipelineRollout: ctlrcommon.CreateTestPipelineRollout(numaflowv1.PipelineSpec{InterStepBufferServiceName: ctlrcommon.DefaultTestISBSvcRolloutName},
+				map[string]string{}, map[string]string{}, map[string]string{}, map[string]string{}),
```
juliev0, line 114: in this function, I just changed the function signature for `CreateTestPipelineRollout()` so nothing really here of interest. (same with numaflowcontrollerrollout test below)

internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -291,17 +291,27 @@ func (r *MonoVertexRolloutReconciler) processExistingMonoVertex(ctx context.Cont
...
 
+		numaLogger.Debug("processing MonoVertex with Progressive")
+		done, err := progressive.ProcessResourceWithProgressive(ctx, monoVertexRollout, existingMonoVertexDef, mvNeedsToUpdate, r, r.client)
+		if err != nil {
+			return err
+		}
+		if done {
+			r.inProgressStrategyMgr.UnsetStrategy(ctx, monoVertexRollout)
+		}
+
```
juliev0, line 314: just like with Pipelines, if we're in the middle of "progressive" strategy, get the latest version of the MVX just to be sure rather than going to cache (note this is getting the one labeled "promoted")

internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -584,38 +574,29 @@ func (r *MonoVertexRolloutReconciler) IncrementChildCount(ctx context.Context, r
...
-}
-
-func (r *MonoVertexRolloutReconciler) Drain(ctx context.Context, monoVertexDef *unstructured.Unstructured) error {
-	patchJson := `{"spec": {"lifecycle": {"desiredPhase": "Paused"}}}`
-	return kubernetes.PatchResource(ctx, r.client, monoVertexDef, patchJson, k8stypes.MergePatchType)
+func (r *MonoVertexRolloutReconciler) Recycle(ctx context.Context,
+	monoVertexDef *unstructured.Unstructured,
+	c client.Client,
+) error {
+	return kubernetes.DeleteResource(ctx, c, monoVertexDef)
```
juliev0, line 581: replacing the original `ChildIsDrained()` and `Drain()` calls which were part of the `progressiveController` interface with `Recycle()` (since draining is not something MVX does, this is more generic)

internal/controller/monovertexrollout/monovertexrollout_controller.go
juliev0: But I did realize that we need to exclude the "replicas" field, which is something that Numaflow Controller may set and we want to ignore it.

internal/controller/monovertexrollout/monovertexrollout_controller_ginkgo_test.go
```diff
@@ -38,31 +38,6 @@ var _ = Describe("MonoVertexRollout Controller", Ordered, func() {
...
-		Sink: &numaflowv1.Sink{
-			AbstractSink: numaflowv1.AbstractSink{
-				UDSink: &numaflowv1.UDSink{
-					Container: &numaflowv1.Container{
-						Image: "quay.io/numaio/numaflow-java/simple-sink:stable",
-					},
-				},
-			},
-		},
-	}
```
juliev0, line 64: moved to `monovertexrollout_controller_test.go` and both files have tests that use it

internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -98,7 +98,11 @@ func (r *PipelineRolloutReconciler) processExistingPipelineWithPPND(ctx context.
 
 	// but if the PipelineRollout says to pause and we're Paused (or won't pause), stop doing PPND in that case too
 	specBasedPause := r.isSpecBasedPause(newPipelineSpec)
-	if specBasedPause && numaflowtypes.IsPipelinePausedOrWontPause(ctx, existingPipelineDef, pipelineRollout) {
+	pausedOrWontPause, err := numaflowtypes.IsPipelinePausedOrWontPause(ctx, existingPipelineDef, pipelineRollout, false)
+	if err != nil {
+		return false, err
+	}
+	if specBasedPause && pausedOrWontPause {
```
juliev0, line 105: now `IsPipelinePausedOrWontPause()` returns an error

internal/controller/progressive/progressive.go
```diff
@@ -108,108 +120,231 @@ func makeUpgradingObjectDefinition(ctx context.Context, rolloutObject ctlrcommon
...
+		WithValues("namespace", existingUpgradingChildDef.GetNamespace(), "name", existingUpgradingChildDef.GetName()).
+		Debugf("Upgrading child is in phase %s", upgradingObjectStatus.Phase)
+
+	if upgradingObjectStatus.Phase == "Running" && isNumaflowChildReady(&upgradingObjectStatus) {
+		return AssessmentResultSuccess, nil
+	}
+	if upgradingObjectStatus.Phase == "Failed" {
+		return AssessmentResultFailure, nil
+	}
+	return AssessmentResultUnknown, nil
```
juliev0, line 341: this function in the future will do an actual assessment of data movement

chandankumar4 (approved):
looks good

PR #420, opened 2024-11-18
test: generate pod logs for resources during e2e tests (WIP)

### LINKED ISSUE #404: generating Pod logs in addition to controller logs

dpadhiar:
Fixes #404 

### Modifications

WORK IN PROGRESS

TESTING CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -178,4 +178,13 @@ jobs:
           path: |
             tests/e2e/output/resources/
           retention-days: 7
+
+      - name: Archive pod logs
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: pod-logs-${{matrix.ppnd}}
+          path: |
+            tests/e2e/output/pods/
+          retention-days: 7
         
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -1,9 +1,11 @@
 package e2e
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -16,6 +18,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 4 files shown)


tests/e2e/common.go
```diff
@@ -130,40 +139,106 @@ func getNumaflowResourceStatus(u *unstructured.Unstructured) (kubernetes.Generic
...
-		stream, err := client.CoreV1().Pods(namespace).GetLogs(pod.Name, podLogOptions).Stream(ctx)
-		if err != nil {
-			fmt.Printf("Error getting pods logs: %v\n", err)
+	for {
+		select {
+		case event := <-watcher.ResultChan():
+			if event.Type == watch.Modified || event.Type == watch.Added {
+				pod := event.Object.(*corev1.Pod)
+				for _, container := range pod.Spec.Containers {
+					streamPodLogs(context.Background(), kubeClient, Namespace, pod.Name, container.Name, stopCh)
```
juliev0, line 155: I see this can be called on `event.Type == watch.Modified` - does that imply it gets called for the same Pod more than once?

tests/e2e/suite_test.go
```diff
@@ -136,27 +148,37 @@ var _ = AfterEach(func() {
...
 })
 
 func setupOutputDir() {
 
-	var dirs = []string{ResourceChangesPipelineOutputPath, ResourceChangesISBServiceOutputPath,
-		ResourceChangesMonoVertexOutputPath, ResourceChangesNumaflowControllerOutputPath}
+	var (
+		dirs = []string{ResourceChangesPipelineOutputPath, ResourceChangesISBServiceOutputPath,
+			ResourceChangesMonoVertexOutputPath, ResourceChangesNumaflowControllerOutputPath}
+		podDirs = []string{PodLogsPipelineOutputPath, PodLogsISBServiceOutputPath,
```
juliev0, line 164: nit: can we rename to `logDirs`?

tests/e2e/common.go
dpadhiar: Hm, I'm considering this might be causing an issue with our open files. Previously for the other watches I am using modified but that would make more sense in those cases as we want to keep updating on the state of the resource YAML. For this, we just want to be informed of new pods. Going to try a CI run with just Added.

.github/workflows/ci.yaml
```diff
@@ -159,23 +159,23 @@ jobs:
...
         env:
           GOPATH: /home/runner/go
-        run: KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true PPND=${{matrix.ppnd}} make test-e2e
+        run: KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true PPND=${{matrix.ppnd}} ENABLE_POD_LOGS=true make test-e2e
 
-      - name: Archive controller logs
+      - name: Archive resource changes
         uses: actions/upload-artifact@v4
-        if: always()
+        if: failure()
```
juliev0, line 166: how about we always do this instead of just on failure?

tests/e2e/suite_test.go
```diff
@@ -124,9 +136,9 @@ var _ = AfterSuite(func() {
 	cancel()
 	By("tearing down test environment")
 	close(stopCh)
-	if disableTestArtifacts != "true" {
-		getPodLogs(kubeClient, Namespace, NumaplaneLabel, "manager", filepath.Join(ControllerOutputPath, "numaplane-controller.log"))
-	}
+	// if disableTestArtifacts != "true" {
+	// 	getPodLogs(kubeClient, Namespace, NumaplaneLabel, "manager", filepath.Join(ControllerOutputPath, "numaplane-controller.log"))
+	// }
```
juliev0, line 141: should we clean this up or are you leaving it there just in case we need it?

dpadhiar: Should be cleaned up now.

juliev0 (approved):

PR #424, opened 2024-11-19
fix: PausingPipelines Condition should get set upon next reconciliation if there's a failure

### LINKED ISSUE #421: e2e test failure: need to ensure idempotent application of Conditions
**Describe the bug**
[This](https://github.com/numaproj/numaplane/actions/runs/11900642612/job/33161911859#step:9:1521) e2e test failure (PPND=true) was caused by the fact that the ISBServiceRollout Condition failed to get set idempotently in the code.

We attempted to set it [here](https://github.com/numaproj/numaplane/blob/main/internal/controller/ppnd/ppnd_pause_requester.go#L124) but got a resource version conflict error, then failed to enter that same code block on the subsequent reconciliation. Need to make sure there's no way that can happen.

---

juliev0:
Fixes #421 

### Modifications

Previously, we had to be in the middle of the "ppnd" in-progress-strategy in order to unset the "PausingPipelines" Conditions. I moved the setting/unsetting of that Condition to the same place where we process Child Health conditions, which is called by the `Reconcile()` function every time (regardless of whether a failure occurred previously). This ensures that we set it.

### Verification

Ran e2e test through CI.


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -554,6 +554,10 @@ func (r *ISBServiceRolloutReconciler) processISBServiceStatus(ctx context.Contex
 		}
 	}
 
+	// check if PPND strategy is requesting Pipelines to pause, and set true/false
+	// (currently, only PPND is accounted for as far as system pausing, not Progressive)
+	_ = r.MarkRolloutPaused(ctx, rollout, ppnd.IsRequestingPause(r, rollout))
+
```
juliev0, line 560: We always call this function, so there's no danger of missing this

internal/controller/isbservicerollout/isbservicerollout_controller_test.go
```diff
@@ -147,7 +147,7 @@ func Test_reconcile_isbservicerollout_PPND(t *testing.T) {
 			initialInProgressStrategy:  apiv1.UpgradeStrategyNoOp,
 			expectedPauseRequest:       &trueValue,
 			expectedRolloutPhase:       apiv1.PhasePending,
-			expectedConditionsSet:      map[apiv1.ConditionType]metav1.ConditionStatus{apiv1.ConditionPausingPipelines: metav1.ConditionTrue},
+			expectedConditionsSet:      map[apiv1.ConditionType]metav1.ConditionStatus{},
```
juliev0, line 150: this is no longer set in `reconcile()` (it's set in `Reconcile()`)

chandankumar4 (approved):

PR #428, opened 2024-11-20
fix: child resource's labels and annotations are combination of Rollout's specified and existing resource

### LINKED ISSUE #427: Labels and Annotations should be a combination of what's specified in the Rollout and what's in the existing resource
**Describe the bug**
When we make an update to the Labels and Annotations for a pipeline, isbsvc, or monovertex, use the existing resource definition, but also add/replace any keys that are identified in the Rollout.

---

dpadhiar:
Fixes #427 

### Modifications

When merging an existing spec with the new spec defined in the parent Rollout, we previously only used the new spec's annotations and labels in the resulting spec. This can cause issues when determining what the appropriate update strategy should be and some labels or annotation can be lost. 

We remedy this by using the `MergeMaps` function to merge the old and new labels and annotations maps without losing any previous keys.

```
func MergeMaps(existing, new map[string]string) map[string]string {
	merged := make(map[string]string)
	if existing != nil {
		merged = existing
	}

	for key, val := range new {
		merged[key] = val
	}

	return merged
}
```

This PR also includes updates to the `ChildNeedsUpdating` functions for the PipelineRollout and MonoVertexRollout reconcilers which also compares the `metadata` of the old and new specs to determine if an update is needed if there is a difference.

### Verification

Created unit test case for MergeMaps function. 

Tested locally with ISBService, Pipeline and MonoVertex rollouts that update from instance 0 to 1 and checking that each child resource has their annotations updated.

### DIFF
--- a/internal/controller/common/merge.go
+++ b/internal/controller/common/merge.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+func MergeMaps(existing, new map[string]string) map[string]string {
+	merged := make(map[string]string)
+	if existing != nil {
+		merged = existing
+	}
+
+	for key, val := range new {
+		merged[key] = val
+	}
+
+	return merged
+}
--- a/internal/controller/pipelinerollout/pipelinerollout_controller_test.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_controller_test.go
@@ -659,7 +659,7 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
 			initialInProgressStrategy:      apiv1.UpgradeStrategyPPND,
 			numaflowControllerPauseRequest: &falseValue,
 			isbServicePauseRequest:         &falseValue,
-			expectedInProgressStrategy:     apiv1.UpgradeStrategyNoOp,
+			expectedInProgressStrategy:     apiv1.UpgradeStrategyPPND,
 			expectedRolloutPhase:           apiv1.PhaseDeployed,
 			expectedPipelineSpecResult: func(spec numaflowv1.PipelineSpec) bool {
 				return reflect.DeepEqual(ctlrcommon.PipelineWithDesiredPhase(pipelineSpecWithTopologyChange, numaflowv1.PipelinePhaseRunning), spec)
--- a/internal/controller/pipelinerollout/pipelinerollout_controller_test.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_controller_test.go
@@ -654,15 +654,15 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
 			existingPipelineDef: *ctlrcommon.CreateTestPipelineOfSpec(
 				ctlrcommon.PipelineWithDesiredPhase(pipelineSpecWithTopologyChange, numaflowv1.PipelinePhasePaused),
 				ctlrcommon.DefaultTestPipelineName, numaflowv1.PipelinePhasePaused, numaflowv1.Status{},
-				false, map[string]string{}),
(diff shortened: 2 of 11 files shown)


internal/controller/common/merge.go
```diff
@@ -0,0 +1,30 @@
...
+	merged := make(map[string]string)
+	if existing != nil {
+		merged = existing
+	}
+
+	for key, val := range new {
+		merged[key] = val
+	}
+
+	return merged
```
juliev0, line 29: Great idea!

internal/controller/pipelinerollout/pipelinerollout_controller_test.go
```diff
@@ -659,10 +659,10 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
 			initialInProgressStrategy:      apiv1.UpgradeStrategyPPND,
 			numaflowControllerPauseRequest: &falseValue,
 			isbServicePauseRequest:         &falseValue,
-			expectedInProgressStrategy:     apiv1.UpgradeStrategyNoOp,
+			expectedInProgressStrategy:     apiv1.UpgradeStrategyPPND,
 			expectedRolloutPhase:           apiv1.PhaseDeployed,
 			expectedPipelineSpecResult: func(spec numaflowv1.PipelineSpec) bool {
-				return reflect.DeepEqual(ctlrcommon.PipelineWithDesiredPhase(pipelineSpecWithTopologyChange, numaflowv1.PipelinePhaseRunning), spec)
+				return reflect.DeepEqual(ctlrcommon.PipelineWithDesiredPhase(pipelineSpecWithTopologyChange, numaflowv1.PipelinePhasePaused), spec)
```
juliev0, line 665: What's happening here? Previously, it seems that if the spec were already applied, then we would unpause the pipeline and take it out of UpgradeInProgress=PPND. So, what's happening differently now?

juliev0: Can you investigate further? I know this unit test is a little tough since sometimes things take multiple reconciliations in certain cases. But we need to understand what changed that is causing this.

dpadhiar: I think the metadata must need updating if the Pipeline is staying paused.

dpadhiar: > Can you investigate further? I know this unit test is a little tough since sometimes things take multiple reconciliations in certain cases. But we need to understand what changed that is causing this.

can do!

dpadhiar: This is being caused by us checking the annotations and labels in `ChildNeedsUpdating` in the PipelineRollout reconciler. When commenting those checks out, this test case works with the expected values it did before.

I also tried reconciling the rollout multiple times and that did not make the test pass.

juliev0:
moved to "Draft" - please reopen when ready, thanks!

internal/controller/pipelinerollout/pipelinerollout_controller_test.go
dpadhiar: Potentially found and fixed the issue - the existingPipelineDef doesn't have the labels set that are added to Pipeline when they are created with PPND.

Tested this locally by creating PipelineRollout and checking labels of initial Pipeline.

dpadhiar: Issue was resolved as we found that using `DeepEqual` for maps will not consider nil maps equal to empty maps and thus we created a helper function that checks this first.

internal/controller/pipelinerollout/pipelinerollout_controller_test.go
```diff
@@ -654,7 +654,7 @@ func Test_processExistingPipeline_PPND(t *testing.T) {
 			existingPipelineDef: *ctlrcommon.CreateTestPipelineOfSpec(
 				ctlrcommon.PipelineWithDesiredPhase(pipelineSpecWithTopologyChange, numaflowv1.PipelinePhasePaused),
 				ctlrcommon.DefaultTestPipelineName, numaflowv1.PipelinePhasePaused, numaflowv1.Status{},
-				false, map[string]string{}),
+				false, map[string]string{"numaplane.numaproj.io/isbsvc-name": "my-isbsvc", "numaplane.numaproj.io/parent-rollout-name": "pipelinerollout-test", "numaplane.numaproj.io/upgrade-state": "promoted"}),
```
juliev0, line 657: Hate to ask this, but I see we're adding these to `existingPipelineDef` here but not in the other test cases. For consistency, I wonder if we could. You know how most of them are using the function `createDefaultTestPipeline()`? What if we set these same labels in that function?

juliev0 (approved):

PR #431, opened 2024-11-25
feat: refactoring watch logic for e2e tests

### LINKED ISSUE #430: consolidating watch logic to reduce redundant code

dpadhiar:
Fixes #430 

### Modifications

Refactors watch logic for e2e tests to minimize code redundancy between every call. This is achieved by making the `writeToFile` function that they all share to write the `kubectl get -o yaml` output to the resulting files. We also can start up every watch before the suite starts as opposed to starting them when the resource is created.

### Verification

Running `make test-e2e` and checking result artifacts.

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -294,16 +294,19 @@ func writeToFile(fileName string, resource Output) error {
 	mutex.Lock()
 	defer mutex.Unlock()
 
-	file, err := os.Create(fileName)
-	if err != nil {
-		fmt.Printf("Failed to open log file: %v\n", err)
-		return err
+	if _, ok := openFiles[fileName]; !ok {
+		file, err := os.Create(fileName)
+		if err != nil {
+			fmt.Printf("Failed to open log file: %v\n", err)
+			return err
+		}
+		openFiles[fileName] = file
 	}
-	openFiles[fileName] = file
 
+	file := openFiles[fileName]
 	bytes, _ := yaml.Marshal(resource)
 	updateLog := fmt.Sprintf("%s\n%v\n\n%s\n", LogSpacer, time.Now().Format(time.RFC3339Nano), string(bytes))
-	_, err = file.WriteString(updateLog)
+	_, err := file.WriteString(updateLog)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


tests/e2e/common.go
```diff
@@ -294,3 +284,23 @@ func watchPods() {
...
 }
+
+// helper func to write `kubectl get -o yaml` output to file
+func writeToFile(fileName string, resource Output) error {
+	file, err := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		fmt.Printf("Failed to open log file: %v\n", err)
+		return err
+	}
+	defer file.Close()
```
juliev0, line 295: this is interesting that the file is being closed with every write. Might be a little less performant to continually open and close, but I guess it's working. The alternative would be to maintain a list of all open files and then close them all at the end. But since it's working, I'm fine as is. Maybe we can just add a "TODO" comment: "for better performance, don't close file after every write"

dpadhiar: That's right, this approach is much more cautious just cause I don't want any files to continue to be open for too long or not close to cause an issue.

Keeping a list of every open file would be good, we could automatically close everything when a test fails as well too.

juliev0: I'm fine if you just want to add a "TODO" for now - up to you

dpadhiar: I think I could add this change to this PR since we're adding the `writeToFile` func here and the other comment to consolidate further could be its own PR.

tests/e2e/common.go
```diff
@@ -294,3 +287,38 @@ func watchPods() {
...
 	}
 }
+
+// helper func to write `kubectl get -o yaml` output to file
+func writeToFile(fileName string, resource Output) error {
+
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	file, err := os.Create(fileName)
```
juliev0, line 297: We should probably avoid creating this again if we already find it in the map, right?

juliev0 (approved):

PR #438, opened 2024-11-30
fix: Ensure NumaflowControllerRollout is not removed until dependent are deleted

### LINKED ISSUE #425: Ensure NumaflowControllerRollout is not removed until dependent resources are
# Summary

ArgoCD will no longer use sync waves, so we need another way to ensure that the Numaflow Controller is up to remove the finalizers of all other resources that are being reconciled by it. A finalizer on the NumaflowControllerRollout can be used which only gets removed once there are no dependent resources remaining.

---

chandankumar4:
Fixes #425 

### Modifications

Hold the deletion of NumaflowControllerRollout util the dependent resource are deleted

### Verification

Verified in local kind cluster.
Tested Scenario:
- Created NumaflowControllerRollout, ISBServiceRollout, PipelineRollout and MonovertexRollout
- Tried Deleting NumaflowControllerRollout - It got stuck in deleting as the dependent resources are still running and Numaplane controller started showing logs as below:
`{"level":"warn","numaflowcontrollerrollout":{"Namespace":"example-namespace","Name":"numaflow-controller"},"logger":"numaplane.controller-manager.numaflowcontrollerrollout-reconciler","caller":"numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go:248","ts":"2024-11-30T16:15:59.713068342Z","msg":"Dependent resources are not deleted yet, requeue after 5 seconds"}`
- Deleted all the dependent resource, then NumaflowControllerRollout get auto deleted.

### DIFF
--- a/internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
+++ b/internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
@@ -245,9 +245,9 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
 		if controllerutil.ContainsFinalizer(controllerRollout, finalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(controllerKey)
 			// Check if dependent resources are deleted, if not then requeue after 5 seconds
-			if !r.isDependentResourceDeleted(ctx, controllerRollout) {
+			if !r.areDependentResourcesDeleted(ctx, controllerRollout) {
 				numaLogger.Warn("Dependent resources are not deleted yet, requeue after 5 seconds")
-				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+				return ctrl.Result{Requeue: true}, nil
 			}
 			controllerutil.RemoveFinalizer(controllerRollout, finalizerName)
 		}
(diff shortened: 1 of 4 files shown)


internal/util/metrics/metrics.go
```diff
@@ -43,8 +43,8 @@ type CustomMetrics struct {
 	MonoVertexROSyncs *prometheus.CounterVec
 	// NumaflowControllersRolloutHealth is the gauge for the health of NumaflowControllerRollouts.
 	NumaflowControllersRolloutHealth *prometheus.GaugeVec
-	// NumaflowControlleRORunning is the gauge for the number of running NumaflowControllerRollouts with a specific version.
-	NumaflowControlleRORunning *prometheus.GaugeVec
+	// NumaflowControllerRORunning is the gauge for the number of running NumaflowControllerRollouts with a specific version.
+	NumaflowControllerRORunning *prometheus.GaugeVec
```
chandankumar4, line 47: This change just fix the Typo in name

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -244,10 +244,15 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
 		r.recorder.Eventf(controllerRollout, corev1.EventTypeNormal, "Deleting", "Deleting NumaflowControllerRollout")
 		if controllerutil.ContainsFinalizer(controllerRollout, finalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(controllerKey)
+			// Check if dependent resources are deleted, if not then requeue after 5 seconds
+			if !r.isDependentResourceDeleted(ctx, controllerRollout) {
+				numaLogger.Warn("Dependent resources are not deleted yet, requeue after 5 seconds")
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
juliev0, line 250: nit: wonder if instead of setting `RequeueAfter`, we could just set `Requeue: true`? Then we should be able to make use of the exponential backoff algorithm

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -244,10 +244,15 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
 		r.recorder.Eventf(controllerRollout, corev1.EventTypeNormal, "Deleting", "Deleting NumaflowControllerRollout")
 		if controllerutil.ContainsFinalizer(controllerRollout, finalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(controllerKey)
+			// Check if dependent resources are deleted, if not then requeue after 5 seconds
+			if !r.isDependentResourceDeleted(ctx, controllerRollout) {
```
juliev0, line 248: I like your idea to be dependent only on the pipelines/isbsvc/monovertices which are children of rollout objects, in the off chance that there were another resource that were not owned by a Numaplane rollout. (Although, there are other places in the code in which we do just assume that any we find are owned by Numaplane.)

juliev0: nit: can we rename the function to `areDependentResourcesDeleted()`?

juliev0 (approved):

Issue #379, opened 2024-10-30
During PPND, if Numaflow Controller is in an error state, the pipelines will remain paused 


juliev0:
---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

juliev0:
The issue is that we are [comparing](https://github.com/numaproj/numaplane/blob/v0.7.1/internal/controller/numaflowcontrollerrollout_controller.go#L357-L376) the current image version tag with the one specified in the `NumaflowControllerRollout`, which are the same, so we continually see this as a difference and a reason for pause. 

See comment [here](https://github.com/numaproj/numaplane/issues/377#issuecomment-2455848938) on one way to handle.

juliev0:
@afugazzotto resolved this, so closing

PR #440, opened 2024-12-03
feat: add previous attempt instanceID and version in the NFC Rollout status to skip unnecessary reconciliations

### LINKED ISSUE #377: Numaplane didn't appear to re-reconcile the new NumaflowControllerRollout after the previous one was broken
**Describe the bug**
I was using my test asset in DevPortal. I tried updating my NumaflowControllerRollout from "1.3.3" to "1.3.3-copy1" based on a definition I'd uploaded. My user namespace config set "pause-and-drain" as the preferred strategy.

The "1.3.3-copy1" version failed due to this error:
```
2024-10-30T17:56:52Z	ERROR	Reconciler error	{"controller": "numaflow-controller-rollout-controller", "object": {"name":"numaflow-controller","namespace":"oss-analytics-testasyncjulie-e2e"}, "namespace": "oss-analytics-testasyncjulie-e2e", "name": "numaflow-controller", "reconcileID": "6e4a8587-6926-489f-8d15-7f90a6eb0d35", "error": "couldn't find image named \"numaflow\" in Deployment &Deployment{ObjectMeta:{numaflow-controller
```

This was expected, since the name of the image in that definition isn't "numaflow".

However, when I tried to re-update the NumaflowControllerRollout back to "1.3.3", nothing got fixed. 

---
[numaplane.log](https://github.com/user-attachments/files/17577121/numaplane.log)
[numaflowcontrollerrollout.yaml.txt](https://github.com/user-attachments/files/17577129/numaflowcontrollerrollout.yaml.txt)

afugazzotto:
Fixes #377 #379 

### Modifications

Added previous attempt instanceID and version to the NumaflowControllerRollout status to be set on success or failure of the previous reconciliation so that if the values match the current version/instanceID the controller will skip the update. This is particularly useful in case of failures to be able to avoid useless retires and to be able to recover by performing a new version change.

### Verification

Unit tests and E2E tests should remain unchanged. Manually tested various scenarios:
- updating the version of a single instance using a valid version
- updating the version of a single instance using an invalid version and trying to recover by changing the version back to the previous one and also to a newer existing version
- performing the same tests as above but on multiple instance at the same time


internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -901,6 +902,7 @@ func (r *NumaflowControllerRolloutReconciler) updateNumaflowControllerRolloutSta
 
 func (r *NumaflowControllerRolloutReconciler) updateNumaflowControllerRolloutStatusToFailed(ctx context.Context, controllerRollout *apiv1.NumaflowControllerRollout, err error) error {
 	controllerRollout.Status.MarkFailed(err.Error())
+	controllerRollout.Status.UpdatePreviousAttemptStatus(controllerRollout)
```
juliev0, line 905: We need to think through this and make sure there's no transient error that would cause us to get here. And then we would never try again.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -377,7 +378,19 @@ func (r *NumaflowControllerRolloutReconciler) isControllerDeploymentUpdating(ctx
...
+		controllerRollout.Status.PreviousAttemptStatus.InstanceID, controllerRollout.Status.PreviousAttemptStatus.Version,
+		controllerRollout.Spec.Controller.InstanceID, controllerRollout.Spec.Controller.Version)
+
+	// If the Rollout status indicates previous attempt instanceID and version same to the current instanceID and version,
+	// it means that the Numaplane Controller already attempted to reconcile this instance/version of the Numaflow Controller.
+	// Therefore, independently of the previous attempt result (either success or failure), the reconciler should not perform
+	// the update (but without pausing the pipelines).
+	controllerVersionNeedsToUpdate := (controllerRollout.Spec.Controller.Version != currentVersion &&
+		(controllerRollout.Status.PreviousAttemptStatus.InstanceID != controllerRollout.Spec.Controller.InstanceID ||
+			controllerRollout.Status.PreviousAttemptStatus.Version != controllerRollout.Spec.Controller.Version))
```
juliev0, line 392: sorry if I mis-read your slack message. I didn't realize you were still comparing to the version in `getControllerDeploymentVersion()`. I was imagining that we were going to remove that altogether, so the reason I was saying to remove the `DefaultNumaflowControllerImageName` code was because what I really meant was that we should remove the entire `getControllerDeploymentVersion()` function, thus no longer comparing to the live image at all, and only comparing to the previously saved version in the Status. 

Do you see any value with also comparing to the version of the image running?

juliev0: Sorry for lack of clarity

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
juliev0: Consider this: what happens if there's a network error hitting the Kubernetes API?

juliev0: This is why I think I'm leaning to the `NumaflowController` approach to be honest unless we've got a better idea. 

In that case, the `NumaflowControllerRollout` reconciler merely determines if it needs to update based on whether its spec is different from the `NumaflowController` spec. If it is, it pauses the pipelines and then updates the `NumaflowController` spec. It only determines that it can unpause the pipelines once it determines that the entire Numaflow Controller Deployment (which the `NumaflowController` reconciler deployed) has been fully deployed. It becomes very similar to the `ISBServiceRollout` reconciler.

juliev0: There is one way I can think of to make the current solution viable - if we store the time of first failure and then say “if it’s been failed for x seconds, then consider it failed and give up”. Note we may replace this code in the future with the NumaflowController code, however

afugazzotto:
Keeping this PR into draft after discussing with @juliev0 to instead move forward with the implementation of the NumaflowController instead of this solution. Creating a NumaflowController CRD should solve this issue in a safer way.

Issue #388, opened 2024-11-04
Configure SyncPeriod


juliev0:
# Summary

Reconciliations normally occur through:

- a change to a Watched resource in the informer cache
- a call we made to RequeueAfter to queue after some period of time

In addition, it will still be good to have a fail safe to ensure we reconcile every resource at some minimum regularity. [This](https://akashjain971.medium.com/part-1-configure-reconcile-frequency-4a000b98cb0#:~:text=Though%20Kubernetes%20is%20designed%20to,after%20the%20object%20is%20processed) article how to configure that. Note the default is 10 hours, and I believe we should make it more frequent than that. At this point, I would say once every 15 minutes.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
I have attempted [this](https://github.com/numaproj/numaplane/pull/432/files) but I'm not sure why it's not working

chandankumar4:
After doing some debugging and research, I landed on this [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/2355) from controller-runtime which is exactly our problem is, as per mentioned in the issue I have tried removing the predicate filter and set SyncPeriod to 2min and it worked as expected, I see in numaplane log which is reconciling every 2min.

PR #444, opened 2024-12-04
feat: created new NumaflowController CRD

### LINKED ISSUE #390: Add a new CRD called "NumaflowController"
# Summary

In the future, if we have a single `NumaflowControllerRollout`, which has 2 children, it would create better modularity to have a `NumaflowController` CRD to represent them.  There would be more parity with other Rollouts and could potentially benefit from some shared code. (I am not sure if in this case we should use a RawExtension or not, though. It's not needed, so probably having strong typing is better. Only question is if we'd benefit from shared code if it were.)

Some benefits of this:
- if we use Progressive delivery with a single `NumaflowControllerRollout`, we could label one `NumaflowController` as "promoted" and the other as "upgrading", just like we do with the other Rollouts (otherwise, we need to label the "Deployment" specifically, which is okay but less symmetrical)
- it's one way to take care of bugs like [this](https://github.com/numaproj/numaplane/issues/379) one and [this](https://github.com/numaproj/numaplane/issues/377) one (see [comment](https://github.com/numaproj/numaplane/issues/377#issuecomment-2455848938))

# Use Cases

When would you use this?

---

afugazzotto:
Fixes #390 

### Modifications

Added the initial scaffolding for the new NumaflowController CRD. We will need to implement the reconciliation logic and add appropriate status, label, spec, etc.


pkg/apis/numaplane/v1alpha1/numaflowcontroller_types.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NumaflowControllerSpec defines the desired state of NumaflowController
+type NumaflowControllerSpec struct {
+	Version string `json:"version"`
```
juliev0, line 28: I think we need `InstanceID` field too, right?

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -0,0 +1,120 @@
...
+	// return ctrl.NewControllerManagedBy(mgr).
+	// 	For(&numaplanenumaprojiov1alpha1.NumaflowController{}).
+	// 	Complete(r)
+
+	controller, err := runtimecontroller.New(ControllerNumaflowController, mgr, runtimecontroller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch ISBServiceRollouts
```
juliev0, line 100: typo here

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -0,0 +1,120 @@
...
+		Kind:    apiv1.NumaflowControllerGroupVersionKind.Kind,
+		Group:   apiv1.NumaflowControllerGroupVersionKind.Group,
+		Version: apiv1.NumaflowControllerGroupVersionKind.Version,
+	})
+	if err := controller.Watch(source.Kind(mgr.GetCache(), numaflowControllerUns,
+		handler.TypedEnqueueRequestForOwner[*unstructured.Unstructured](mgr.GetScheme(), mgr.GetRESTMapper(),
+			&apiv1.NumaflowController{}, handler.OnlyControllerOwner()), predicate.TypedResourceVersionChangedPredicate[*unstructured.Unstructured]{})); err != nil {
+		return fmt.Errorf("failed to watch NumaflowController: %v", err)
+	}
+
```
juliev0, line 118: I guess later you can add the watching of the [child resources](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go#L814-L853)

juliev0: (note that @chandankumar4  is actually going to be touching these watch functions as he addresses the [Sync-Period issue](https://github.com/numaproj/numaplane/issues/388)) so there will be unfortunately be some small merge conflict

afugazzotto: I was planning to move the watch logic later when also changing that logic in the NumaflowController Rollout since now that logic will go here while the NumaflowControllerRollout should only watch the NumaflowController. Correct?
I don't think there will be conflicts now since this is a new file. @chandankumar4 and I can sync up though to address the Sync-Period once we work on our respective watch changes based on how merges those changes first.

juliev0: you're right - I shouldn't have said there will be a merge conflict - just that you're touching the same type of stuff. I think it will be apparent for him what to change

juliev0 (approved):

PR #129, opened 2024-07-22
Fixing security vulnerabilities Present in Quay.io

shubhamdixit863:
Yes

Fixes https://github.com/numaproj/numaplane/issues/82

### Modifications

According to the quay.io scan, the vulnerability was caused by the layer that was copying Helm into the Docker image. I have upgraded the Helm version to the recent v3.15.3 in the Dockerfile, which seems to have fixed the issue.

### Verification

The Quay.io scanner does not show the security vulnerability anymore. There are two unknown security issues that I still need to work on.
Screenshot 2024-07-22 at 5 24 42 PM

PR #449, opened 2024-12-11
fix: Debug event logger and use zerolog instead of fmt

chandankumar4:
Fixes #129 

### Modifications

- Fix debug event logger
- Replace printf statement to zerolog 

### Verification

Verified in local kubernetes cluster, attached the log file for reference
[numaplane.log](https://github.com/user-attachments/files/18093527/numaplane.log)

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 
 	"github.com/fsnotify/fsnotify"
+	"github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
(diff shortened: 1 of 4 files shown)


internal/controller/config/config.go
```diff
@@ -114,7 +115,7 @@ func (cm *NumaflowControllerDefinitionsManager) UpdateNumaflowControllerDefiniti
 	for _, controller := range config.ControllerDefinitions {
 		cm.rolloutConfig[controller.Version] = controller.FullSpec
 
-		fmt.Printf("Added/Updated Controller definition Config, version: %s\n", controller.Version)
+		log.Debug().Msg(fmt.Sprintf("Added/Updated Controller definition Config, version: %s", config)) // due to cyclical dependency, we can't call logger
```
juliev0, line 118: good solution to the problem of the cyclical dependency :) I forgot about these print statements

juliev0 (approved):

juliev0:
Did you basically search the entire codebase for `fmt.Print` (so as to catch both `fmt.Print` and `fmt.Printf`?

chandankumar4:
> fmt.Printf

Yes, It's only left on e2e test cases and 1-2 places in unit test cases

PR #450, opened 2024-12-11
feat: enable new NumaflowController and split logic

### LINKED ISSUE #452: Enable new NumaflowController
# Summary

Split the current NumaflowControllerRollout logic as follows:
- the NumaflowController controller will take care of syncing the Numaflow resources defined in the definitions configmap for the specified version
- the NumaflowControllerRollout controller will take care of handling the upgrade strategy (direct apply, ppnd, progressive) similarly to what ISBSvcRollout does

---

afugazzotto:
Fixes #452 

### Modifications

Split the current NumaflowControllerRollout logic as follows:
- the NumaflowController controller will take care of syncing the Numaflow resources defined in the definitions configmap for the specified version
- the NumaflowControllerRollout controller will take care of handling the upgrade strategy (direct apply, ppnd, progressive) similarly to what ISBSvcRollout does

### Verification

Manual tests, unit tests, e2e tests.


pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -71,6 +71,9 @@ type Status struct {
 
 	// ObservedGeneration stores the generation value observed when setting the current Phase
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// UpgradeInProgress indicates the upgrade strategy currently being used and affecting the resource state or empty if no upgrade is in progress
+	UpgradeInProgress UpgradeStrategy `json:"upgradeInProgress,omitempty"`
```
juliev0, line 76: Good idea. It does seem now that we will use this across all 4 Rollouts

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -76,45 +121,623 @@ func NewNumaflowControllerReconciler(
...
+
+	// Get the target manifests based on the version of the controller and throw an error if the definition not for a version.
+	version := controller.Spec.Version
+	definition := config.GetConfigManagerInstance().GetControllerDefinitionsMgr().GetNumaflowControllerDefinitionsConfig()
+	manifest := definition[version]
+	if len(manifest) == 0 {
+		return gitopsSyncCommon.OperationError, fmt.Errorf("no controller definition found for version %s", version)
+	}
+
+	// Update templated manifest with information from the
```
juliev0, line 405: unfinished comment :)

pkg/apis/numaplane/v1alpha1/status_types.go
afugazzotto: Thanks. I thought to move some functions and fields I saw used by multiple controllers (ex: IsHealthy func). Then a single controller can decide to use a field/func or not.

internal/controller/numaflowcontroller/numaflowcontroller_controller_test.go
```diff
@@ -16,4 +16,376 @@ limitations under the License.
...
+			existingPauseRequest:    nil,
+			expectedPauseRequest:    nil,
+			expectedPhase:           apiv1.PhaseDeployed,
+			expectedConditionsSet: map[apiv1.ConditionType]metav1.ConditionStatus{
+				apiv1.ConditionChildResourceDeployed: metav1.ConditionTrue,
+			},
+			expectedResultControllerVersion: "1.2.0",
+		},
+		{
+			name:                 "new Controller version, pipelines paused",
```
juliev0, line 208: I see the unit test involves pausing - was this brought over from before or did you need to modify it? I'm wondering if it should still go in NumaflowControllerRollout's unit test, no?

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,312 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
-	}
-	reconciliationResult := gitopsSync.Reconcile(targetObjs, liveObjByKey, namespace, infoProvider)
-
-	// Ignore `status` field for all comparison.
-	// TODO: make it configurable
-	overrides := map[string]sync.ResourceOverride{
-		"*/*": {
-			IgnoreDifferences: sync.OverrideIgnoreDiff{JSONPointers: []string{"/status"}}},
-	}
+// TODO: do we need to also check the underlying Deployment?
```
juliev0, line 452: the answer to this "TODO" is "yes". We must essentially verify the 3 items listed in the comments below. The danger is that if we only verify that the NumaflowController has been reconciled, the underlying Deployment or ReplicaSet or Pod is still the old one. We need for everything to be updated before we unpause pipelines.

juliev0: The document I wrote up speaks to this issue. See [here](https://docs.google.com/document/d/1JYdANNPE8GH-vFeJNP-AgjYra4nvZOk8eeV2-xx3oeQ/edit?tab=t.0#bookmark=id.owhl2v3hlkzq).

I know we said we could postpone what's in here, but on the other hand, I don't want commit this as is, if it has this risk. We can consider whether to implement the solution from that document or a temporary solution to address it.

juliev0: One thing to mention is that there is a benefit if we can get all of the information from the `NumaflowController` resource and not have to look at the `Deployment` as well. This is related to [this](https://github.com/numaproj/numaplane/pull/450#discussion_r1882612842) comment. If we look at both resources, and if they both came from cache, then one of them may be out of date - there is not atomic correctness between the two, and we may make a false decision. 

(Note that my document above does get all the info from the `NumaflowController`)

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,312 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
-	deployment, deploymentExists, err := r.getNumaflowControllerDeployment(ctx, controllerRollout)
+	newNumaflowControllerDef, err := generateNewNumaflowControllerDef(nfcRollout)
 	if err != nil {
-		return ctrl.Result{}, err
+		return ctrl.Result{}, fmt.Errorf("error generating NumaflowController: %v", err)
 	}
 
-	// determine the Upgrade Strategy user prefers
-	upgradeStrategy, err := usde.GetUserStrategy(ctx, controllerRollout.Namespace)
+	existingNumaflowControllerDef, err := kubernetes.GetResource(ctx, r.client, newNumaflowControllerDef.GroupVersionKind(),
```
juliev0, line 234: technically we don't need to get this as an `Unstructured` type. We could use the actual type. Did you use `Unstructured` because you were thinking you could take advantage of shared code somewhere? 

You can alternatively use `r.client.Get()` with an actual `NumaflowController`

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -811,124 +541,49 @@ func (r *NumaflowControllerRolloutReconciler) SetupWithManager(mgr ctrl.Manager)
...
+func (r *NumaflowControllerRolloutReconciler) ErrorHandler(nfcRollout *apiv1.NumaflowControllerRollout, err error, reason, msg string) {
+	r.customMetrics.NumaflowControllerRolloutSyncErrors.WithLabelValues().Inc()
+	r.recorder.Eventf(nfcRollout, corev1.EventTypeWarning, reason, msg+" %v", err.Error())
 }
 
-func (r *NumaflowControllerRolloutReconciler) updateNumaflowControllerRolloutStatusToFailed(ctx context.Context, controllerRollout *apiv1.NumaflowControllerRollout, err error) error {
-	controllerRollout.Status.MarkFailed(err.Error())
-	return r.updateNumaflowControllerRolloutStatus(ctx, controllerRollout)
-}
+func generateNewNumaflowControllerDef(nfcRollout *apiv1.NumaflowControllerRollout) (*unstructured.Unstructured, error) {
```
juliev0, line 574: this could also return a `NumaflowController` type

juliev0: note: if we do decide to change these to `NumaflowController` type, I'm okay if it's a follow-up PR

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
juliev0: I see that you are calling the `usde` which takes an Unstructured type - perhaps that's the rationale for using `Unstructured`? Anything else?

juliev0: I think I'm good with it

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
juliev0: I see you needed to use Unstructured in order to use usde, so I think I'm good with it

juliev0:
Wow, you really wrote a significant amount of code in the last week!

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,312 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
+	numaLogger.Debugf("numaflowController status: %+v", numaflowControllerStatus)
 
-	diffOpts := []diff.Option{
-		diff.WithLogr(*numaLogger.LogrLogger),
-		diff.WithServerSideDiff(true),
-		diff.WithServerSideDryRunner(diff.NewK8sServerSideDryRunner(resourceOps)),
-		diff.WithManager(common.SSAManager),
-		diff.WithGVKParser(clusterCache.GetGVKParser()),
-	}
+	numaflowControllerReconciled := numaflowController.GetGeneration() <= numaflowControllerStatus.ObservedGeneration
```
juliev0, line 467: so, I'm guessing this one isn't working either right now...but that can be handled separately I guess

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,312 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
-func (r *NumaflowControllerRolloutReconciler) GetPipelineList(ctx context.Context, rolloutNamespace string, _ string) (*unstructured.UnstructuredList, error) {
-	return kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion, "pipelines", rolloutNamespace, common.LabelKeyParentRollout, "")
+func (r *NumaflowControllerRolloutReconciler) GetPipelineList(ctx context.Context, rolloutNamespace string, rolloutName string) (*unstructured.UnstructuredList, error) {
+	gvk := schema.GroupVersionKind{Group: common.NumaflowAPIGroup, Version: common.NumaflowAPIVersion, Kind: common.NumaflowPipelineKind}
+	// List all the pipelines since they are all managed by the same single NumaplaneController
+	return kubernetes.ListResources(ctx, r.client, gvk, rolloutNamespace)
 }
 
 func (r *NumaflowControllerRolloutReconciler) GetRolloutKey(rolloutNamespace string, rolloutName string) string {
+	// TODO: we may want to also use the rolloutName to identify and store the NumaflowControllers
```
juliev0, line 279: Interesting. Yes, I guess if one day somebody does create 2 NumaflowControllerRollouts with different Instances, that would be the case.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,312 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
-	return "Numaflow Controller"
+	return "numaflowcontroller"
 }
 
-func (r *NumaflowControllerRolloutReconciler) GetPipelineList(ctx context.Context, rolloutNamespace string, _ string) (*unstructured.UnstructuredList, error) {
-	return kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion, "pipelines", rolloutNamespace, common.LabelKeyParentRollout, "")
+func (r *NumaflowControllerRolloutReconciler) GetPipelineList(ctx context.Context, rolloutNamespace string, rolloutName string) (*unstructured.UnstructuredList, error) {
+	gvk := schema.GroupVersionKind{Group: common.NumaflowAPIGroup, Version: common.NumaflowAPIVersion, Kind: common.NumaflowPipelineKind}
+	// List all the pipelines since they are all managed by the same single NumaplaneController
+	return kubernetes.ListResources(ctx, r.client, gvk, rolloutNamespace)
```
juliev0, line 275: (deleted my original comment here in case you saw it - decided it was not an issue)

juliev0:
Have you had a chance to test the phase and Conditions for each resource fairly well? (I'm okay if we have a follow up issue to do so)

internal/controller/numaflowcontroller/numaflowcontroller_controller_test.go
juliev0: If we simply removed this function, I'm guessing we would probably still have the same code coverage due to the other test?

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -17,53 +17,98 @@ limitations under the License.
...
+	client client.Client,
+	scheme *runtime.Scheme,
+	rawConfig *rest.Config,
+	kubectl kubeUtil.Kubectl,
 	customMetrics *metrics.CustomMetrics,
 	recorder record.EventRecorder,
-) *NumaflowControllerReconciler {
+) (*NumaflowControllerReconciler, error) {
+	stateCache := sync.NewLiveStateCache(rawConfig, customMetrics)
+	numaLogger := logger.GetBaseLogger().WithName("state cache").WithValues("numaflowcontroller")
```
juliev0, line 90: Curious about the logger name?

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
juliev0: I think maybe the shortest path for now would be to have a "Progressing" condition in the NumaflowController and use that over here.

internal/controller/numaflowcontroller/numaflowcontroller_controller_test.go
afugazzotto: Sure, I'll remove those unit tests from this file then.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
afugazzotto: That is one of the reasons, I also tried to keep the code consistent between ISBSvcRollout and NumaflowControllerRollout for easier maintainability and to be able to possibly reduce code duplication at some point.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
afugazzotto: Good catch. Once we get in your temporary fix on the unstructured util this should resolve as well.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,312 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
+		} else {
+			if healthyChildCond != nil {
+				nfcRollout.Status.MarkChildResourcesUnhealthy(healthyChildCond.Reason, healthyChildCond.Message, nfcRollout.Generation)
 			} else {
-				// expect first letter to be a "v"
-				if tag[0] != 'v' {
-					return "", fmt.Errorf("expected numaflow-controller tag to start with 'v': %q", tag)
-				}
-				return tag[1:], nil
+				nfcRollout.Status.MarkChildResourcesUnhealthy("Progressing", "Progressing", nfcRollout.Generation)
```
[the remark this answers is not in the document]
afugazzotto: I'll change it to generation <= observedGeneration. Good catch.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
afugazzotto: I'll keep the TODO for now, we can address this in a separate PR if the need arises.

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
afugazzotto: I kept it from this: https://github.com/numaproj/numaplane/blob/9d05eb5a74449dfb08e0077da23bf3230e807937/internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go#L98
I could remove it though, I'm also unsure about it's purpose.

afugazzotto: Found that it is used by the state cache for logging. Moving the name in the state cache directly.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
afugazzotto: As discussed over Zoom, I modified to code to include the "Progressing" condition check and we will make adjustments with a separate PR.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
afugazzotto: Added comment.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
afugazzotto: Same as above.

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,313 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
-// interface that provides functionality to manage kubernetes resources. Returns a
-// cleanup function that must be called to remove the generated kube config for this
-// server.
-func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.ResourceOperations, func(), error) {
-	clusterCache, err := r.stateCache.GetClusterCache()
-	if err != nil {
-		return nil, nil, fmt.Errorf("error getting cluster cache: %w", err)
-	}
+	numaflowControllerReconciled := numaflowController.GetGeneration() <= numaflowControllerStatus.ObservedGeneration &&
+		healthyChildCond.Status == metav1.ConditionTrue && healthyChildCond.Reason != apiv1.ProgressingReasonString
```
juliev0, line 479:
```suggestion
        ncProgressing := healthyChildCond.Reason == apiv1.ProgressingReasonString
	numaflowControllerReconciled := numaflowController.GetGeneration() <= numaflowControllerStatus.ObservedGeneration && !ncProgressing
```

juliev0: Otherwise, we are setting `numaflowControllerReconciled=true` for degraded state as well

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -269,539 +226,313 @@ func (r *NumaflowControllerRolloutReconciler) reconcile(
...
-	}
-	reconciliationResult := gitopsSync.Reconcile(targetObjs, liveObjByKey, namespace, infoProvider)
-
-	// Ignore `status` field for all comparison.
-	// TODO: make it configurable
-	overrides := map[string]sync.ResourceOverride{
-		"*/*": {
-			IgnoreDifferences: sync.OverrideIgnoreDiff{JSONPointers: []string{"/status"}}},
-	}
+// TODO: do we need to also check the underlying Deployment?
```
juliev0, line 455:
```suggestion
```

juliev0: removing comment now that we're addressing it

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
juliev0: Once complete, do you mind testing to make sure the pipeline is remaining paused until the new Controller pod is created? You can run e2e locally and have one terminal running a watch on “kubectl get pipeline” and another terminal running a watch on “kubectl get pod”

juliev0 (approved):

PR #451, opened 2024-12-11
fix: get live info for isbsvc and statefulset during PPND/Progressive

### LINKED ISSUE #448: pipelines are being unpaused too early when performing isbsvc update
**Describe the bug**
Noticed when updating isbsvc that the pipeline is being unpaused before isbsvc pods have fully updated.

---

juliev0:
Fixes #448 

### Modifications

There were 2 issues:
1. probably the main issue: the `Generation` field was of the wrong type and thus was being returned as a 0. Thus, we always passed the test of [isbsvcReconciled](https://github.com/numaproj/numaplane/blob/main/internal/controller/isbservicerollout/isbservicerollout_controller.go#L500), so we never pause the pipeline for very long.
2. The check for [isbsvcReconciled](https://github.com/numaproj/numaplane/blob/main/internal/controller/isbservicerollout/isbservicerollout_controller.go#L500) relies on an atomic state for `isbsvc` and `StatefulSet` in time. Therefore, we can't have the isbsvc be outdated. 

### Verification

At least 5 times I ran and observed the Pipeline only went out of `desiredPhase=Paused` once the StatefulSet Pods had all been replaced. I also confirmed through resource timestamps and logs.


internal/controller/common/numaflowtypes/isbsvc.go
```diff
@@ -30,18 +31,32 @@ import (
...
+			LabelSelector: fmt.Sprintf("%s=%s", numaflowv1.KeyISBSvcName, isbsvc.GetName())})
+		if err != nil {
+			return nil, fmt.Errorf("Error listing live StatefulSets: %v", err)
+		}
+		statefulSetList = *statefulSets
+	} else {
+		//TODO: this is currently making a Live K8S call, not a call to cache as it's supposed to
+		// Option 1: figure out how we can watch a StatefulSet and have its owner's owner reconcile it
+		// Option 2: update InterstepBufferService code in Numaflow to provide all the info we need so we don't need to access StatefulSet directly
+		err = c.List(ctx, &statefulSetList, &client.ListOptions{Namespace: isbsvc.GetNamespace(), LabelSelector: statefulSetSelector})
```
juliev0, line 56: logged [this](https://github.com/numaproj/numaplane/issues/453) issue based on this comment

juliev0:
@afugazzotto would you mind taking a look at this when you get a chance? 

I won't merge it as is in any case - would like to resolve the `generation` issue first. (Plus, I have the lint error in that code and the raw print statements that need to be removed)

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -332,8 +327,28 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
...
+		newISBServiceDef = r.merge(existingISBServiceDef, newISBServiceDef)
+	}
+
 	switch inProgressStrategy {
 	case apiv1.UpgradeStrategyPPND:
+
+		_, isbServiceIsUpdating, err := r.isISBServiceUpdating(ctx, isbServiceRollout, existingISBServiceDef, true)
+		if err != nil {
+			return false, fmt.Errorf("error determining if ISBService is updating: %v", err)
+		}
```
juliev0, line 350: by moving the call down here, we can use our live version of `existingISBServiceDef`

internal/util/kubernetes/unstructured_util.go
```diff
@@ -53,7 +53,7 @@ func GetLiveResource(
 		return nil, err
 	}
 
-	unstruc, err := dynamicClient.Resource(gvr).Namespace(object.GetNamespace()).Get(ctx, object.GetName(), metav1.GetOptions{})
+	unstruc, err := DynamicClient.Resource(gvr).Namespace(object.GetNamespace()).Get(ctx, object.GetName(), metav1.GetOptions{})
```
chandankumar4, line 56: Do we have plan to use "DynamicClient" outside kubernetes package?

juliev0: To me, it seems okay if we were to. And it seems weird to only expose `KubernetesClient` and not `DynamicClient`. I don't really want to be constrained that only the `kubernetes` package can use it. Do you disagree?

afugazzotto (approved):

Issue #456, opened 2024-12-13
Determine if we can remove the `StructToStruct` calls in unstructured_util.go


juliev0:
**Describe the bug**
By calling `StructToStruct` from `GetResource()` and `GetLiveResource()` functions, the `Generation` field is changing from an integer field to a float64 field. Would like to remove `StructToStruct` if it doesn't cause harm. Need to figure out if we can instead update tests.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

chandankumar4:
Observation: Before and after conversion of `unstruc.Object` via `util.StructToStruct()` method, here `rpu` type is changed from `int64` to `float64`

Before:
<img width="522" alt="Screenshot 2024-12-15 at 5 27 49 PM" src="https://github.com/user-attachments/assets/3779b4c1-a667-4b66-84ff-200746d009db" />

After:
<img width="522" alt="Screenshot 2024-12-15 at 5 28 05 PM" src="https://github.com/user-attachments/assets/3221ff6d-616e-4890-bdc9-209bfec2178d" />

juliev0:
Are all numbers float64 after? Maybe json unmarshal just does that?

chandankumar4:
> Are all numbers float64 after? Maybe json unmarshal just does that?

Yes, Json unmarshal does that, similar conversation going [here](https://forum.golangbridge.org/t/type-problem-in-json-conversion/19420), which is closed saying `JSON does only know floats`

chandankumar4:
[This](https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout/pipelinerollout_controller.go#L783) is where problem start, here it converts the `pipelineRollout.Spec.Pipeline.Spec` of type `runtime.RawExtension` to `pipelineSpec` of type `map[string]interface{}` and treat numeric value as float64, so everywhere we have to convert `int64` value to `float64` otherwise `reflect.DeepEqual()` fail the comparison due to type difference.
So there is two possible fix:
1. Fix the Json parser to convert number to int64 instead of float64
2. Fix `reflect.DeepEqual()` method to only compare value not type in case of number

juliev0:
Which call to `reflect.DeepEqual()` are you referring to? In a test do you mean?

juliev0:
> Fix the Json parser to convert number to int64 instead of float64

[...]

PR #459, opened 2024-12-16
fix: Add compare struct method to compare it without num type

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/456

### DIFF
--- a/internal/usde/usde.go
+++ b/internal/usde/usde.go
@@ -53,11 +53,11 @@ func resourceSpecNeedsUpdating(ctx context.Context, newDef, existingDef *unstruc
 
 	// Get data loss fields config based on the spec type (Pipeline, ISBS)
 	dataLossFields := []config.SpecDataLossField{}
-	if util.CompareStructWithoutNumKind(newDef.GroupVersionKind(), numaflowv1.PipelineGroupVersionKind) {
+	if util.CompareStructNumTypeAgnostic(newDef.GroupVersionKind(), numaflowv1.PipelineGroupVersionKind) {
 		dataLossFields = usdeConfig.PipelineSpecDataLossFields
-	} else if util.CompareStructWithoutNumKind(newDef.GroupVersionKind(), numaflowv1.ISBGroupVersionKind) {
+	} else if util.CompareStructNumTypeAgnostic(newDef.GroupVersionKind(), numaflowv1.ISBGroupVersionKind) {
 		dataLossFields = usdeConfig.ISBServiceSpecDataLossFields
-	} else if util.CompareStructWithoutNumKind(newDef.GroupVersionKind(), apiv1.NumaflowControllerGroupVersionKind) {
+	} else if util.CompareStructNumTypeAgnostic(newDef.GroupVersionKind(), apiv1.NumaflowControllerGroupVersionKind) {
 		// TODO: for NumaflowController updates do we need to figure out which strategy to use based on the type of changes similarly done for Pipeline and ISBSvc?
 		// OR should we always return the user's preferred strategy?
 		// For now, make the entire spec a data loss field and include all its subfields
@@ -108,7 +108,7 @@ func resourceSpecNeedsUpdating(ctx context.Context, newDef, existingDef *unstruc
 
 		if dataLossField.IncludeSubfields {
 			// is the definition (fields + children) at all different?
-			if !util.CompareStructWithoutNumKind(newDefField, existingDefField) {
+			if !util.CompareStructNumTypeAgnostic(newDefField, existingDefField) {
 				return true, upgradeStrategy, nil
 			}
 		} else {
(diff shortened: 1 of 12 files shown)


internal/util/util.go
```diff
@@ -7,21 +7,76 @@ import (
 	"strings"
 )
 
-// StructToStruct converts a struct type (src) into another (dst)
-func StructToStruct(src any, dst any) error {
+func JsonUnmarshaler(src any, dst any) error {
```
chandankumar4, line 10: Probably need a better naming for it?

internal/util/util.go
```diff
@@ -7,21 +7,76 @@ import (
...
+		case map[string]any:
+			convertNumbers(t)
+		case []any:
+			convertNumbersArray(t)
+		}
+	}
+}
+
+func convertNumber(value json.Number) any {
+	i64, err := value.Int64()
```
juliev0, line 69: It looks like this will return an Int64 if it can be parsed as an Int64. But would that work in all cases?

Let's say hypothetically that the `Generation` field were actually supposed to be a float64. But in the json it happens to be a "4". Thus, it can be parsed as an Int64. But then later we call `GetGeneration()` and the type is an Int64 instead of a Float64. It will fail.

internal/util/util.go
```diff
@@ -22,6 +24,50 @@ func StructToStruct(src any, dst any) error {
...
+}
+
+func isNumber(value any) bool {
+	v := reflect.TypeOf(value)
+	if v == nil {
+		return false
+	}
+	kind := v.Kind()
+	switch kind {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Float32, reflect.Float64:
```
juliev0, line 50: what about the Uint types?

internal/util/util_test.go
```diff
@@ -119,3 +119,32 @@ func Test_ExtractPath(t *testing.T) {
...
+			name:               "equal structs - same number type",
+			a:                  map[string]interface{}{"number": 1},
+			b:                  map[string]interface{}{"number": 1},
+			expectedComparison: true,
+		},
+		{
+			name:               "equal structs - different number type",
+			a:                  map[string]interface{}{"number": 1},
+			b:                  map[string]interface{}{"number": float64(1)},
+			expectedComparison: true,
```
juliev0, line 140: would you mind adding at least one more case in here, which includes the uint type?

internal/util/util.go
```diff
@@ -22,6 +24,50 @@ func StructToStruct(src any, dst any) error {
 	return nil
 }
 
+func CompareStructWithoutNumKind(src, dst any) bool {
```
juliev0, line 27: nit: could we maybe rename to `CompareStructNumTypeAgnostic()`?

internal/util/util.go
```diff
@@ -22,6 +24,50 @@ func StructToStruct(src any, dst any) error {
...
+		vy, _ := toFloat64(y)
+		return vx == vy
+	})
+
+	// Apply this custom comparison only to pairs of values where both are numbers
+	filterNumber := cmp.FilterValues(func(x, y any) bool {
+		return isNumber(x) && isNumber(y)
+	}, numberComparer)
+
+	equal := cmp.Equal(src, dst, filterNumber)
```
juliev0, line 39: cool find

internal/util/util.go
chandankumar4: Oh, Missed it, Thanks for reminding :)

internal/util/util_test.go
chandankumar4: Added test cases for all types

internal/usde/usde.go
```diff
@@ -54,11 +53,11 @@ func resourceSpecNeedsUpdating(ctx context.Context, newDef, existingDef *unstruc
...
 	// Get data loss fields config based on the spec type (Pipeline, ISBS)
 	dataLossFields := []config.SpecDataLossField{}
-	if reflect.DeepEqual(newDef.GroupVersionKind(), numaflowv1.PipelineGroupVersionKind) {
+	if util.CompareStructNumTypeAgnostic(newDef.GroupVersionKind(), numaflowv1.PipelineGroupVersionKind) {
 		dataLossFields = usdeConfig.PipelineSpecDataLossFields
-	} else if reflect.DeepEqual(newDef.GroupVersionKind(), numaflowv1.ISBGroupVersionKind) {
+	} else if util.CompareStructNumTypeAgnostic(newDef.GroupVersionKind(), numaflowv1.ISBGroupVersionKind) {
 		dataLossFields = usdeConfig.ISBServiceSpecDataLossFields
-	} else if reflect.DeepEqual(newDef.GroupVersionKind(), apiv1.NumaflowControllerGroupVersionKind) {
+	} else if util.CompareStructNumTypeAgnostic(newDef.GroupVersionKind(), apiv1.NumaflowControllerGroupVersionKind) {
```
juliev0, line 60: Sorry, I know it's not hurting anything to have this, but it may create more confusion to read. Maybe we should just put these 3 back to DeepEqual() function?

juliev0: (since there are no numbers involved)

chandankumar4: Sure, NP

juliev0 (approved):

PR #460, opened 2024-12-16
fix: Configure SyncPeriod for reconcile at given interval of time

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/388

### Modifications

Add custom predicate filter which filter all the events other than "ResourceVersion" change and "SyncPeriod" event.

Reference taken from already open issue in controller-runtime: https://github.com/kubernetes-sigs/controller-runtime/issues/2355#issuecomment-2477548322

### Verification

Verified on local Kubernetes cluster by setting the SyncPeriod time as 2 min, Numaplane is reconciling every resource every 2min.

### DIFF
--- a/internal/controller/common/predicate_filters.go
+++ b/internal/controller/common/predicate_filters.go
@@ -39,6 +39,7 @@ type TypedGenerationChangedPredicate[object metav1.Object] struct {
 }
 
 func (p TypedGenerationChangedPredicate[object]) Update(e event.TypedUpdateEvent[object]) bool {
+	// It will effectively check of ObjectOld and ObjectNew does not exist or if the name of the object is empty
 	if e.ObjectOld.GetName() == "" || e.ObjectNew.GetName() == "" {
 		return false
 	}
--- a/internal/controller/common/predicate_filters.go
+++ b/internal/controller/common/predicate_filters.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package common
 
 import (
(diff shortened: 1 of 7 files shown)


internal/controller/common/predicate_filters.go
```diff
@@ -0,0 +1,27 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type TypedGenerationChangedPredicate[object metav1.Object] struct {
+	predicate.TypedFuncs[object]
+}
+
+func (p TypedGenerationChangedPredicate[object]) Update(e event.TypedUpdateEvent[object]) bool {
```
chandankumar4, line 13: Reference: https://github.com/kubernetes-sigs/controller-runtime/issues/2355#issuecomment-2477548322

internal/controller/common/predicate_filters.go
```diff
@@ -0,0 +1,27 @@
+package common
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type TypedGenerationChangedPredicate[object metav1.Object] struct {
```
juliev0, line 9: can we add comments for what this is all about?

internal/controller/common/predicate_filters.go
```diff
@@ -0,0 +1,27 @@
...
+		return false
+	}
+
+	if e.ObjectNew.GetGeneration() != e.ObjectOld.GetGeneration() {
+		return true
+	}
+
+	if e.ObjectNew.GetResourceVersion() == e.ObjectOld.GetResourceVersion() {
+		return true
+	}
```
juliev0, line 24: maybe we can take the original [comments](type MyKindCustomPredicate struct {
	// Returns true by default for all events; see overrides below.
	predicate.Funcs
}

func (p MyKindCustomPredicate) Update(e event.UpdateEvent) bool {
	if e.ObjectOld == nil || e.ObjectNew == nil {
		return false
	}
	
	// Process the event if the generation changed (which happens e.g. if the spec
	// was updated, but not if the status or metadata fields were updated).
	if e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() {
		return true
	}
	
	// If the generation is unchanged, we generally don't want to process the event,
	// with the exception of events triggered by periodic resyncs (which are, for some
	// reason, categorized as updates). We identify such events by checking if the old
	// and new resource versions are equal.
	return e.ObjectOld.GetResourceVersion() == e.ObjectNew.GetResourceVersion()
}) and add them here as well?

internal/controller/common/predicate_filters.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+type TypedGenerationChangedPredicate[object metav1.Object] struct {
+	predicate.TypedFuncs[object]
+}
+
+func (p TypedGenerationChangedPredicate[object]) Update(e event.TypedUpdateEvent[object]) bool {
+	if e.ObjectOld.GetName() == "" || e.ObjectNew.GetName() == "" {
+		return false
+	}
```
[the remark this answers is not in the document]
chandankumar4: Yes, Build was failing with below error
```bash
1.420 internal/controller/common/predicate_filters.go:14:20: invalid operation: e.ObjectOld == nil (mismatched types object and untyped nil)
1.420 internal/controller/common/predicate_filters.go:14:42: invalid operation: e.ObjectNew == nil (mismatched types object and untyped nil)

```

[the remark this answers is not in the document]
juliev0: Okay, can you add a comment that this check is effectively testing if either e.ObjectOld or e.ObjectNew doesn't exist?

juliev0 (approved):

PR #463, opened 2024-12-17
feat: update e2e test to check for promoted label on child resources

### LINKED ISSUE #393: e2e test for Progressive - update functional_test.go to check "promoted" pipeline, monovertex (and isbsvc when ready)
# Summary

We will need to start to test the Progressive logic. 

I'm thinking that we should be able to add "Progressive" as one variety of the current `functional_test.go`. Just like `ppnd`, we'd have some logic in there that's conditionally performed only in the case of `progressive`.

In order to make this work, we need to change how we do things. Instead of looking for a pipeline, monovertex, isbsvc by name, we will look for the one whose upgrade state Label is "promoted" (note the "isbsvc" code isn't in there yet but I am in process with it).

For "nostrategy" and "pause-and-drain", there is only one "promoted" child so it will effectively do the same thing as now except that the test doesn't specify the child resource by name. 

For "progressive", ultimately, the "upgrading" pipeline/monovertex/isbsvc will become the new "promoted" one. 

Please let me know if you see any reason why this won't work.

---

dpadhiar:
Fixes #393 

### Modifications

Modifies `functional_test.go` to instead use a helper function `getPipelineName` or `getMonoVertexName` to get the child resource's name instead of outright using the constant we have previously used before. This is due to the Progressive strategy of updating rollouts eventually changing the instance of each child resource ie. `my-pipeline-0` will be `my-pipeline-1` after a successful update. 

We can achieve this currently by listing the resources with the label `numaplane.numaproj.io/upgrade-state=promoted` which there should only be 1 of with the correct label `numaplane.numaproj.io/parent-rollout-name`.

### Verification

Running `make test-e2e` locally and passing multiple times. Double checking with CI now.

No compatibility errors as this is for end to end testing.


tests/e2e/functional_test.go
```diff
@@ -353,7 +354,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 		document("Verifying Pipeline got updated")
 
 		// get Pipeline to check that spec has been updated to correct spec
-		verifyPipelineSpec(Namespace, pipelineName, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
+		verifyPipelineSpec(Namespace, getPipelineName(Namespace, pipelineRolloutName), func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
```
juliev0, line 357: Actually, what I'm thinking we may want to do is query for the "promoted" child inside the `Eventually` calls in the individual functions like `verifyPipelineSpec()`. Ideally, we should be able to call something like `verifyPipelineSpec()` and within the call the "promoted" pipeline should be able to change from `pipeline-0` to `pipeline-1`. `pipeline-0` does not have the spec we want, but `pipeline-1` does.

juliev0: Since this would be a major change, maybe create a second PR and just leave this one up in Draft?

dpadhiar: Got it, so we just want to remove the pipelineName parameter in the individual functions. I can open a new PR to work on that approach.

PR #464, opened 2024-12-18
feat: NumaflowController update delete and recreate child resources

### LINKED ISSUE #455: NumaflowController to delete/recreate rather than replace
# Summary

```
``` 

---

afugazzotto:
Fixes #455 

### Modifications

During NumaflowController reconciliation, if a Deployment already exists for the NumaflowController and the new version differs from the current Deployment version, delete all the NumaflowController child resources and recreate them at the next reconciliation.

### Verification

Executed unit tests and e2e tests.
Manually tested the following cases:
- going from a valid version to an invalid version and back to the previous valid version
- going from a valid version to an invalid version, then to a new valid version
- going from a valid version to a new valid version
- going from a version without instance labels to a version with instance labels

There should be no backward incompatibilities in these changes.

### DIFF
--- a/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
+++ b/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
@@ -257,12 +257,20 @@ func (r *NumaflowControllerReconciler) reconcile(
 		return ctrl.Result{}, err
 	}
 
+	newVersion := controller.Spec.Version
+	targetObjs, err := determineTargetObjects(controller, newVersion)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to determine the target objects for the new version %s: %w", newVersion, err)
+	}
+
+	numaLogger.Debugf("found %d target objects associated with NumaflowController version %s", len(targetObjs), newVersion)
+
 	// If a Deployment already exists for the NumaflowController and the new version differs from the current Deployment version,
 	// delete all the NumaflowController child resources and recreate them at the next reconciliation
 	if deploymentExists {
 		currentVersion, err := getControllerDeploymentVersion(existingDeployment)
 		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("error while getting the version of the existing NumaflowController Deployment: %v", err)
+			return ctrl.Result{}, fmt.Errorf("error while getting the version of the existing NumaflowController Deployment: %w", err)
 		}
 
 		childResourcesNeedToBeDeleted := controller.Spec.Version != currentVersion
@@ -271,7 +279,7 @@ func (r *NumaflowControllerReconciler) reconcile(
 
 			err := r.deleteNumaflowControllerChildren(controller, currentVersion, namespace)
 			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("error deleting NumaflowController child resources: %v", err)
+				return ctrl.Result{}, fmt.Errorf("error deleting NumaflowController child resources: %w", err)
 			}
 
 			return ctrl.Result{}, nil
@@ -283,7 +291,7 @@ func (r *NumaflowControllerReconciler) reconcile(
 	// - new Controller
 	// - auto healing
 	// - somebody changed the manifest associated with the Controller version (shouldn't happen but could)
-	phase, err := r.sync(controller, namespace, numaLogger)
+	phase, err := r.sync(controller, namespace, numaLogger, targetObjs)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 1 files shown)


internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -388,41 +417,60 @@ func resolveManifestTemplate(manifest string, controller *apiv1.NumaflowControll
...
-) (gitopsSyncCommon.OperationPhase, error) {
+	version string,
+) ([]*unstructured.Unstructured, error) {
 
-	// Get the target manifests based on the version of the controller and throw an error if the definition not for a version.
-	version := controller.Spec.Version
+	// Get the target manifests based on the given version and throw an error if the definition does not have that version
 	definition := config.GetConfigManagerInstance().GetControllerDefinitionsMgr().GetNumaflowControllerDefinitionsConfig()
 	manifest := definition[version]
 	if len(manifest) == 0 {
```
juliev0, line 440: So, `len()` works to check existence in a map? That's interesting. I just see this method when I look up checking existence in a map:
```
manifest, found := definition[version]
if !found {
```

juliev0: (just want to make sure we're not doing some non-standard thing that happens to work)

juliev0:
>going from a valid version to an invalid version and back to the previous valid version
going from a valid version to an invalid version, then to a new valid version
going from a valid version to a new valid version
going from a version without instance labels to a version with instance labels

Thanks for spelling all of these out. Did you verify the pipeline paused and unpaused for these? (I think you showed me that but want to verify)

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -741,3 +789,84 @@ func (r *NumaflowControllerReconciler) areDependentResourcesDeleted(ctx context.
...
+
+// deleteNumaflowControllerChildren deletes child resources associated with a NumaflowController
+// for a specified version and namespace. It determines the target objects to delete and attempts
+// to remove them using the Kubernetes client. If a resource is not found, it continues with the
+// next one. Returns an error if unable to determine target objects or delete a resource.
+func (r *NumaflowControllerReconciler) deleteNumaflowControllerChildren(controller *apiv1.NumaflowController, currentVersion, namespace string) error {
+
+	targetObjs, err := determineTargetObjects(controller, currentVersion)
+	if err != nil {
+		return fmt.Errorf("unable to determine the target objects for the current version %s: %w", currentVersion, err)
```
juliev0, line 851: How about if this were to happen, we handle it gracefully? Let's say that the current version's definition were for some reason no longer there

juliev0: Ideally, as a back up, determine the target objects from the new version?

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -252,17 +252,46 @@ func (r *NumaflowControllerReconciler) reconcile(
...
+		childResourcesNeedToBeDeleted := controller.Spec.Version != currentVersion
+		if childResourcesNeedToBeDeleted {
+			numaLogger.Debugf("current Deployment image tag=%q differs from desired %q", currentVersion, controller.Spec.Version)
+
+			err := r.deleteNumaflowControllerChildren(controller, currentVersion, namespace)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("error deleting NumaflowController child resources: %w", err)
+			}
+
+			return ctrl.Result{}, nil
```
juliev0, line 285: oh, did you decide not to explicitly requeue because you are figuring that we'll automatically re-reconcile due to the change in child resources that we're watching?

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
afugazzotto: I did not notice that `len()` was being used to check version existence. I also prefer the other way to check if something exists on a map. I'll make that change.

afugazzotto:
> > going from a valid version to an invalid version and back to the previous valid version
> > going from a valid version to an invalid version, then to a new valid version
> > going from a valid version to a new valid version
> > going from a version without instance labels to a version with instance labels
> 
> Thanks for spelling all of these out. Did you verify the pipeline paused and unpaused for these? (I think you showed me that but want to verify)

Yes, I made sure the pipeline pauses and unpauses as expected for all those tests.

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
afugazzotto: I assumed this would not be a scenario that could happen. I'll check the new version as backup though.

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
afugazzotto: That's correct, I didn't think requeuing was necessary unless you can think of a case in which reconciliation would not re-execute. If we want to be 100% sure, I can explicitely requeue.

afugazzotto: On a second thought, would it be better to explicitly requeue to start the next reconciliation faster?

juliev0: I think we should be okay as is. Can we just add a comment at that return statement about the assumption we're making?

Worst case, Chandan's new [PR ](https://github.com/numaproj/numaplane/pull/460) will ensure we re-reconcile every 15 minutes.

juliev0 (approved):

PR #465, opened 2024-12-18
feat: update e2e tests to accommodate for progressive strategy

### LINKED ISSUE #393: e2e test for Progressive - update functional_test.go to check "promoted" pipeline, monovertex (and isbsvc when ready)

dpadhiar:
Fixes #393 

### Modifications

Modifies `functional_test.go` to instead use a helper function `getPipelineName` or `getMonoVertexName` to get the child resource's name instead of outright using the constant we have previously used before. This is due to the Progressive strategy of updating rollouts eventually changing the instance of each child resource ie. `my-pipeline-0` will be `my-pipeline-1` after a successful update. 

We can achieve this currently by listing the resources with the label `numaplane.numaproj.io/upgrade-state=promoted` which there should only be 1 of with the correct label `numaplane.numaproj.io/parent-rollout-name`.

### Verification

Running `make test-e2e` locally and passing multiple times. Double checking with CI now.

No compatibility errors as this is for end to end testing.

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -19,6 +19,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 
@@ -361,4 +362,20 @@ func closeAllFiles() error {
 	}
 
 	return nil
+}
+
+func getChildResource(gvr schema.GroupVersionResource, namespace, rolloutName string) (*unstructured.Unstructured, error) {
+
+	label := fmt.Sprintf("%s,%s=%s", UpgradeStateLabelSelector, ParentRolloutLabel, rolloutName)
+
+	unstructList, err := dynamicClient.Resource(gvr).Namespace(namespace).List(ctx, metav1.ListOptions{LabelSelector: label})
+	if err != nil {
+		return nil, err
+	}
+	if len(unstructList.Items) == 0 {
+		return nil, fmt.Errorf("list is empty")
+	}
+
+	return &unstructList.Items[0], nil
+
 }
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -81,6 +81,9 @@ const (
 	NumaplaneLabel = "app.kubernetes.io/part-of=numaplane"
 	NumaflowLabel  = "app.kubernetes.io/part-of=numaflow"
 
+	ParentRolloutLabel        = "numaplane.numaproj.io/parent-rollout-name"
+	UpgradeStateLabelSelector = "numaplane.numaproj.io/upgrade-state=promoted"
+
 	LogSpacer = "================================"
 )
 
--- a/tests/e2e/functional_test.go
+++ b/tests/e2e/functional_test.go
@@ -402,8 +402,9 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
 
-		pipelineName := getPipelineName(Namespace, pipelineRolloutName)
-		verifyPodsRunning(Namespace, 0, getVertexLabelSelector(pipelineName))
+		pipeline, err := getPipeline(Namespace, pipelineRolloutName)
+		Expect(err).ShouldNot(HaveOccurred())
+		verifyPodsRunning(Namespace, 0, getVertexLabelSelector(pipeline.GetName()))
 	})
 
 	time.Sleep(2 * time.Second)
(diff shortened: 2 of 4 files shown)


tests/e2e/monovertex.go
```diff
@@ -19,6 +20,27 @@ import (
...
+		unstructList, err := dynamicClient.Resource(getGVRForMonoVertex()).Namespace(namespace).List(ctx, metav1.ListOptions{LabelSelector: label})
+		if err != nil {
+			return false
+		}
+		if len(unstructList.Items) == 0 {
+			return false
+		}
+		monoVertexName = unstructList.Items[0].GetName()
+		return true
+	}, 60*time.Second, testPollingInterval).Should(BeTrue())
```
juliev0, line 38: Maybe we don't need `Eventually` here anymore? Are all the calls to here within their own Eventually or Consistently block? Are all of them followed by a `Get()`? If so, could this just return the MonoVertex itself, and then the subsequent `Get()` doesn't need to be called?

juliev0: (same comment for Pipeline)

dpadhiar: Within `monovertex.go` we are either passing the name to another function or calling Get. In `functional_test.go`, we do need to get the Name to check if the MonoVertex is deleted or not.

juliev0: How about in the cases in which we are subsequently calling `Get()`, don't call `Get()` anymore and instead call `getMonoVertex()` which does not contain an Eventually call and just returns the `*Unstructured`? (and do let me know if I'm missing something)

dpadhiar: Got it! Should be addressed in my latest commit now.

tests/e2e/functional_test.go
```diff
@@ -37,11 +37,11 @@ import (
...
 
 const (
-	isbServiceRolloutName            = "test-isbservice-rollout"
-	pipelineRolloutName              = "test-pipeline-rollout"
-	pipelineName                     = "test-pipeline-rollout-0"
-	monoVertexRolloutName            = "test-monovertex-rollout"
-	monoVertexName                   = "test-monovertex-rollout-0"
+	isbServiceRolloutName = "test-isbservice-rollout"
+	pipelineRolloutName   = "test-pipeline-rollout"
+	// pipelineName                     = "test-pipeline-rollout-0"
```
dpadhiar, line 42: Need to remove this and monoVertexName.

tests/e2e/pipeline.go
```diff
@@ -166,12 +169,12 @@ func updatePipelineRolloutInK8S(namespace string, name string, f func(apiv1.Pipe
 	Expect(err).ShouldNot(HaveOccurred())
 }
 
-func getPipelineFromK8S(namespace string, pipelineName string) (*unstructured.Unstructured, numaflowv1.PipelineSpec, numaflowv1.PipelineStatus, error) {
+func getPipelineFromK8S(namespace string, pipelineRolloutName string) (*unstructured.Unstructured, numaflowv1.PipelineSpec, numaflowv1.PipelineStatus, error) {
```
juliev0, line 172: nit: just thinking about this function name and the `getPipeline()` name - maybe we can call this one `getPipelineSpecAndStatus()` or something like that>

tests/e2e/monovertex.go
```diff
@@ -88,13 +92,14 @@ func verifyMonoVertexReady(namespace, monoVertexName string) {
 
 }
 
-func verifyMonoVertexStatus(namespace, monoVertexName string, f func(numaflowv1.MonoVertexSpec, kubernetes.GenericStatus) bool) {
+func verifyMonoVertexStatus(namespace, monoVertexRolloutName string, f func(numaflowv1.MonoVertexSpec, kubernetes.GenericStatus) bool) string {
```
juliev0, line 95: is this name that it's returning being used anywhere? If not, maybe we can take it out?

juliev0: oh, I see you're using it above

tests/e2e/common.go
```diff
@@ -359,3 +363,19 @@ func closeAllFiles() error {
...
+	unstructList, err := dynamicClient.Resource(gvr).Namespace(namespace).List(ctx, metav1.ListOptions{LabelSelector: label})
+	if err != nil {
+		return nil, err
+	}
+	if len(unstructList.Items) == 0 {
+		return nil, fmt.Errorf("list is empty")
+	}
+
+	return &unstructList.Items[0], nil
+
```
juliev0, line 380: good idea to add this one

juliev0 (approved):

PR #467, opened 2024-12-21
fix: set the NumaflowControllerRollout ChildResourceHealthy condition Unhealthy

afugazzotto:
… to Unhealthy when the NumaflowController is not healthy (phase failed) and the ChildResourceHealthy is not present in the NumaflowController

### Modifications

Set the NumaflowControllerRollout ChildResourceHealthy condition to Unhealthy when the NumaflowController is not healthy (phase failed) and the ChildResourceHealthy is not present in the NumaflowController.

### Verification

WIP

This change should not be backward incompatible.

### DIFF
--- a/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
+++ b/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
@@ -317,7 +317,6 @@ func (r *NumaflowControllerReconciler) reconcile(
 		return ctrl.Result{}, err
 	}
 
-	// phase = gitopsSyncCommon.OperationFailed // TTODO: for testing special case, remove afterwards
 	if phase != gitopsSyncCommon.OperationSucceeded {
 		return ctrl.Result{}, fmt.Errorf("sync operation is not successful")
 	}
--- a/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
+++ b/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
@@ -252,39 +252,9 @@ func (r *NumaflowControllerReconciler) reconcile(
 		controllerutil.AddFinalizer(controller, common.FinalizerName)
 	}
 
-	existingDeployment, deploymentExists, err := r.getNumaflowControllerDeployment(ctx, controller)
+	_, deploymentExists, err := r.getNumaflowControllerDeployment(ctx, controller)
 	if err != nil {
 		return ctrl.Result{}, err
-	}
-
-	newVersion := controller.Spec.Version
-	newVersionTargetObjs, err := determineTargetObjects(controller, newVersion)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to determine the target objects for the new version %s: %w", newVersion, err)
-	}
-
-	numaLogger.Debugf("found %d target objects associated with NumaflowController version %s", len(newVersionTargetObjs), newVersion)
-
-	// If a Deployment already exists for the NumaflowController and the new version differs from the current Deployment version,
-	// delete all the NumaflowController child resources and recreate them at the next reconciliation
-	if deploymentExists {
-		currentVersion, err := getControllerDeploymentVersion(existingDeployment)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("error while getting the version of the existing NumaflowController Deployment: %w", err)
-		}
-
-		childResourcesNeedToBeDeleted := controller.Spec.Version != currentVersion
-		if childResourcesNeedToBeDeleted {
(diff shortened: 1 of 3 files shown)


juliev0:
Although it's failing the test, it looks pretty good to me in general.

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -470,14 +441,35 @@ func (r *NumaflowControllerReconciler) sync(
...
+
+	numaLogger.Debugf("found %d target objects associated with NumaflowController version %s", len(newVersionTargetObjs), newVersion)
+
+	reconciliationResult, diffResults, liveObjectsMap, err := r.compareState(controller, namespace, newVersionTargetObjs, numaLogger)
 	if err != nil {
 		return gitopsSyncCommon.OperationError, err
 	}
 
+	// Delete current resources if any of the specs differ
+	childResourcesNeedToBeDeleted := diffResults.Modified
```
juliev0, line 460: Actually, I'm kind of wondering if this puts us in an infinite loop? If we delete the resources below and requeue then do we come back through here, compare the desired objects to the deleted objects, decide they are different and then attempt to delete again?

juliev0: I do like the idea of each time checking if everything has been deleted. 

If my theory above is true about what's happening, one thought I have is that `childResourcesNeedToBeDeleted` could be based on whether there are in fact resources that need to be deleted. If `diffResults.Modified == true but all of the differences are deletions, then `childResourcesNeedToBeDeleted` could be set to false.

afugazzotto: As discussed over Slack, this solution did not work since liveObjectsMap is always empty (further investigation needed). Changed strategy by comparing the specs of the current version and the new version.

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -258,33 +259,52 @@ func (r *NumaflowControllerReconciler) reconcile(
...
 	if deploymentExists {
 		currentVersion, err := getControllerDeploymentVersion(existingDeployment)
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("error while getting the version of the existing NumaflowController Deployment: %w", err)
 		}
 
-		childResourcesNeedToBeDeleted := controller.Spec.Version != currentVersion
-		if childResourcesNeedToBeDeleted {
-			numaLogger.Debugf("current Deployment image tag=%q differs from desired %q", currentVersion, controller.Spec.Version)
+		currentTargetObjsMap, _, err := determineTargetObjects(controller, currentVersion)
```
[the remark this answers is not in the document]
juliev0: (for this reason, I think just executing the "else" block would be fine, since ultimately we want to make sure that the resources we're attempting to create are deleted first)

juliev0:
I'm realizing one aspect of this PR is that even if nothing has changed in the desired manifest, if we simply perform a `Reconcile()`, we will delete and recreate the manifests. That's what I liked about what you were trying to do this past weekend - we were asking gitops-engine what was changed. Was [StateDiffs()](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L563) not returning a value for `Modified` that was what we wanted? (Spec differences and not Status differences)

juliev0:
I've created a new branch based off of your branch from last Friday and want to see if I can get it to work.

I think there is a bug which causes the [liveObjByKey](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L536) to always return empty. I think this bug has been there all along. If I go back to release-0.7 branch (before your refactor), it's always empty there too. I noticed that the gitops-engine GetManagedLiveObjs() call that we make expects the resources to not have OwnerReferences as far as I can tell, but ours do, so I think this may be why it doesn't work. Maybe if you use gitops-engine you're not supposed to add your own OwnerReferences? I need to investigate further...

juliev0:
FYI, I've been making a bunch of changes [here](https://github.com/numaproj/numaplane/pull/471). This is based off of your PR as of last Friday. I removed OwnerReferences to get the gitops-engine code to work. Not having OwnerReferences caused me to have to change a couple of things that were dependent on them:

- deletion of NumaflowController needs to delete its children
- the `Watch()` functions were using an OwnerReference predicate, so I had to change that to find the "owner" through its label instead

My manual tests are in general working so far.  Looks like e2e tests aren't working yet, so I'll need to investigate that.

PR #470, opened 2024-12-23
feat: update e2e test to incorporate isbsvc look up with "promoted" label

### LINKED ISSUE #469: Update e2e to incorporate isbsvc look up through "promoted" label and "parent-rollout-name" label
# Summary

Even though full blue/green has not been done for the isbsvc, it now uses both "promoted" label and "parent-rollout-name" label. Therefore, we should be able to update the e2e to look up by those now. 

With this change, theoretically, we should be able to run the e2e test like this:
```
PROGRESSIVE=true make start
PPND=false make test-e2e    // this could later become `PROGRESSIVE=true make test-e2e` once we have progressive-specific stuff that we're testing
```

---

dpadhiar:
Fixes #469 

### Modifications

Following a previous PR for pipelines and monovertices, we are updating the e2e test to accommodate for testing `progressive` upgrade strategy with isbservices. This is done by querying the isbservices with the promoted and parent rollout labels instead of getting them with a constant name.

### Verification

Running `make test-e2e` after running `PROGRESSIVE=true make start` a few times.

Testing that CI is working properly.

No changes will impact backwards compatibility as this is a testing update.


juliev0:
this looks good. Do you want it as "Draft" still?

PR #471, opened 2024-12-28
fix: gitops-engine fixes, plus enable Numaflow Controller deleted and then recreated

### LINKED ISSUE #455: NumaflowController to delete/recreate rather than replace

juliev0:
Fixes #455 

### Modifications

This is a follow-up to [this](https://github.com/numaproj/numaplane/pull/464) previous PR. That PR made it so that if the version number changes, then the previous Numaflow Controller manifests are deleted and then recreated. However, it doesn't take care of the cases in which the version number doesn't change but there's some other change to the manifest of a given version. (In fact, we saw this issue when deploying.)

When the [NumaflowController](https://github.com/numaproj/numaplane/blob/main/config/samples/numaplane.numaproj.io_v1alpha1_numaflowcontroller.yaml) reconciler is invoked either through version change or something else, this PR makes it so that we follow this process:
- `NumaflowController` reconciler sees the change and finds the manifests associated with the version
**Scenario 1: It finds the version:**
- `NumaflowController` reconciler deletes the previous manifests and requeues
- If any of those manifests still need to be deleted, it attempts to delete them and requeues. Otherwise, it applies the new manifests
**Scenario 2: it doesn't find the version:**
- `NumaflowController` is marked "Failed"

#### gitops-engine code wasn't really working

While working on this PR, I realized that the code we had around the gitops-engine (for applying Numaflow Controller manifests) wasn't truly working the way it was supposed to. The way it's supposed to work is that we first determine if we need to do an "apply" based on [whether there's a difference](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L477) between what we want to apply and what is live, and then only do an "apply" if there's a difference. However, the code was never able to detect what was [live](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L536), so it was always just doing an "apply". 

I realized that the reason it wasn't able to detect what was live is because gitops-engine assumes that manifests don't have "Owner references". I am guessing since it needs to manage the lifecycle of the resources, it doesn't want something else to have applied an OwnerRef to it? In any case, if I remove the owner references that we were applying previously, it is able to detect the live resources and perform the comparison. 

Because I removed the Owner reference from the manifests, I had to do the following:
1. ensure that we delete the child manifests when the `NumaflowController` is deleted (using [Finalizer](https://github.com/numaproj/numaplane/blob/ncdeletion/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L244))
2. update the [Watch](https://github.com/numaproj/numaplane/blob/ncdeletion/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L682) calls which were previously using the Owner Reference to identify the NumaflowController to Reconcile, to instead use the existing Label to do so

### Verification

Performed the following:
1. Started from release-0.9 branch and updated the 1.3.3 manifest to remove "InstanceID" field from all of the labels
2. Created NumaflowControllerRollout at 1.3.3, ISBServiceRollout, and PipelineRollout
3. Then switched to this branch and started Numaplane
4. Observed everything still running healthily - `NumaflowController` CR was created since that's new
5. Updated NumaflowControllerRollout to 1.4.0 manifest which contains the "InstanceID" label. Observed Pipeline pause, manifest update to 1.4.0, pipelines unpause. Observed NumaflowControllerRollout "Deployed" with child condition healthy
6. Updated NumaflowControllerRollout to 1.4.9, which doesn't exist. Observed Pipeline pause, manifest can't be found, so `NumaflowController` marked "Failed" and `NumaflowControllerRollout` HealthyChildCondition marked False ("degraded"). Numaflow Controller pod still running 1.4.0 and pipeline unpaused.
7. Updated NumaflowControllerRollout back to 1.4.0 and observed `NumaflowController` went back to "Deployed" and `NumaflowControllerRollout` HealthyChildCondition back to True.


internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -239,6 +241,10 @@ func (r *NumaflowControllerReconciler) reconcile(
 				numaLogger.Warn("Dependent resources are not deleted yet, requeue after 5 seconds")
 				return ctrl.Result{Requeue: true}, nil
 			}
+			err := r.deleteChildren(ctx, controller.Namespace, controller.Name)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
```
juliev0, line 247: Because the manifests don't have an OwnerReference anymore, the deletion of the children is no longer automatic when NumaflowController is deleted, so we explicitly do it.

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -248,55 +254,30 @@ func (r *NumaflowControllerReconciler) reconcile(
...
-
-			err := r.deleteNumaflowControllerChildren(ctx, controller, currentVersion, namespace, newVersionTargetObjs)
-			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("error deleting NumaflowController child resources: %w", err)
-			}
-
-			// No need to explicitly requeue since a new reconciliation will occur due to changes in watched child resources
-			return ctrl.Result{}, nil
-		}
+		return ctrl.Result{}, err
```
juliev0, line 266: previous code looked specifically for "version change" to determine if we need to delete/recreate - now do it any time the manifest has a difference, regardless of whether it's due to "version change"

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -454,29 +358,54 @@ func determineTargetObjects(
 	if err != nil {
 		return nil, fmt.Errorf("can not parse file data, err: %w", err)
 	}
-	manifestsWithOwnership, err := applyOwnershipToManifests(manifests, controller)
-	if err != nil {
-		return nil, fmt.Errorf("failed to apply ownership reference, %w", err)
-	}
```
juliev0, line 460: we no longer apply OwnerReference since gitops-engine assumes there's no OwnerReferences

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -454,29 +358,54 @@ func determineTargetObjects(
...
-
-	targetObjs, err := toUnstructuredAndApplyLabel(manifestsWithOwnership, controller.Name)
+	targetObjs, err := toLabeledUnstructured(manifests, controller.Name)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse the manifest, %w", err)
 	}
+	for _, obj := range targetObjs {
+		obj.SetNamespace(namespace)
+	}
```
juliev0, line 367: In order to compare the manifests effectively, the target manifests need to have "namespace" applied

PR #472, opened 2024-12-29
fix:  Numaflow Controller deletion - use gitops-engine settings to delete/recreate instead of doing it ourselves

### LINKED ISSUE #455: NumaflowController to delete/recreate rather than replace

juliev0:
Fixes #455 #377 

### Modifications

This is a follow-up to [this](https://github.com/numaproj/numaplane/pull/464) previous PR. That PR made it so that if the version number changes, then the previous Numaflow Controller manifests are deleted and then recreated. However, it doesn't take care of the cases in which the version number doesn't change but there's some other change to the manifest of a given version. (In fact, we saw this issue when deploying.)

I noticed that gitops-engine can actually be configured to do a "Replace" with force rather than an "Apply", which effectively handles the problem of backward compatibility rather than our needing to explicitly delete the manifest first. 

Note that it is also able to detect whether there is a difference before doing anything: in the log I can see that when there is no difference in a given resource, it just says "Skipping as resource was not modified".

This PR also makes a few other minor fixes which are described inline.

#### Updated e2e test

The e2e test now changes the Numaflow Controller manifest from one version to another in a way which previously would've failed, due to an incompatible change.
In addition, the e2e test verifies the change of the NumaflowControllerRolllout to an invalid version and back again.

#### Manual testing

Performed the following test (multiple times to ensure consistency) to test the path we plan to take in Production going from Numaplane 0.9.1:

1. Started from 0.9.1 tag and updated the 1.3.3 manifest to remove "InstanceID" field from all of the labels
2. Created NumaflowControllerRollout at 1.3.3, ISBServiceRollout, and PipelineRollout
3. Then switched to this branch and started Numaplane
4. Observed everything still running healthily - NumaflowController CR was created since that's new
5. Updated NumaflowControllerRollout to 1.4.0 manifest which contains the "InstanceID" label. Observed Pipeline pause, manifest update to 1.4.0, pipelines unpause. Observed NumaflowControllerRollout "Deployed" with child condition healthy
6. Updated NumaflowControllerRollout to 99.99.99, which doesn't exist. Observed Pipeline pause, manifest can't be found, so NumaflowController marked "Failed" and NumaflowControllerRollout HealthyChildCondition marked False ("degraded"). Numaflow Controller pod still running 1.4.0 and pipeline unpaused.
7. Updated NumaflowControllerRollout back to 1.4.0 and observed NumaflowController went back to "Deployed" and NumaflowControllerRollout HealthyChildCondition back to True.

I have explicitly tested the path from 0.9.1 as described above. I don't think there are any issues.


internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -253,41 +253,19 @@ func (r *NumaflowControllerReconciler) reconcile(
...
-			err := r.deleteNumaflowControllerChildren(ctx, controller, currentVersion, namespace, newVersionTargetObjs)
-			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("error deleting NumaflowController child resources: %w", err)
-			}
-
-			// No need to explicitly requeue since a new reconciliation will occur due to changes in watched child resources
-			return ctrl.Result{}, nil
-		}
-	}
-
```
juliev0, line 290: gitops-engine can delete/recreate so we don't have to

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -481,13 +464,11 @@ func (r *NumaflowControllerReconciler) sync(
 
 	opts := []gitopsSync.SyncOpt{
 		gitopsSync.WithLogr(*numaLogger.LogrLogger),
-		gitopsSync.WithOperationSettings(false, true, false, false),
+		gitopsSync.WithOperationSettings(false, true, true, false),
```
juliev0, line 467: this is "force", so with this and "replace" setting below, we can force replace

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -464,6 +443,10 @@ func determineTargetObjects(
 		return nil, fmt.Errorf("failed to parse the manifest, %w", err)
 	}
 
+	for _, obj := range targetObjs {
+		obj.SetNamespace(namespace)
+	}
```
juliev0, line 448: these should have "namespace" set so they can effectively be compared to the live ones (although, note, that I will follow up later with a full effective comparison as part of [this](https://github.com/numaproj/numaplane/pull/471) PR)

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -546,16 +527,9 @@ func (r *NumaflowControllerReconciler) compareState(
...
-	if err != nil {
-		return gitopsSync.ReconciliationResult{}, nil, err
-	}
-	defer cleanup()
-
 	diffOpts := []diff.Option{
 		diff.WithLogr(*numaLogger.LogrLogger),
-		diff.WithServerSideDiff(true),
-		diff.WithServerSideDryRunner(diff.NewK8sServerSideDryRunner(resourceOps)),
+		diff.WithServerSideDiff(false),
```
juliev0, line 532: we don't need this (it just creates more opportunity for gitops-engine to error out)

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -633,17 +590,17 @@ func processDeploymentHealth(deployment *appsv1.Deployment) (bool, string, strin
...
-			msg := fmt.Sprintf("Waiting for Deployment  to finish: %d old replicas are pending termination...", deploymentStatus.Replicas-deploymentStatus.UpdatedReplicas)
+			msg := fmt.Sprintf("Waiting for Deployment to finish: %d old replicas are pending termination...", deploymentStatus.Replicas-deploymentStatus.UpdatedReplicas)
 			return false, apiv1.ProgressingReasonString, msg
 		} else if deploymentStatus.AvailableReplicas < deploymentStatus.UpdatedReplicas {
-			msg := fmt.Sprintf("Waiting for Deployment  to finish: %d of %d updated replicas are available...", deploymentStatus.AvailableReplicas, deploymentStatus.UpdatedReplicas)
+			msg := fmt.Sprintf("Waiting for Deployment to finish: %d of %d updated replicas are available...", deploymentStatus.AvailableReplicas, deploymentStatus.UpdatedReplicas)
 			return false, apiv1.ProgressingReasonString, msg
 		}
 	} else {
-		msg := "Waiting for Deployment  to finish: observed deployment generation less than desired generation"
```
juliev0, line 646: just removing empty space over here

internal/controller/numaflowcontrollerrollout/numaflowcontrollerrollout_controller.go
```diff
@@ -501,20 +501,26 @@ func (r *NumaflowControllerRolloutReconciler) processNumaflowControllerStatus(
...
+				if healthyChildCond != nil {
+					nfcRollout.Status.MarkChildResourcesUnhealthy(healthyChildCond.Reason, healthyChildCond.Message, nfcRollout.Generation)
+				} else {
+					nfcRollout.Status.MarkChildResourcesUnhealthy("Unhealthy", "Unhealthy", nfcRollout.Generation)
+				}
 			}
+		} else {
+			nfcRollout.Status.MarkChildResourcesUnhealthy("Failed", "Failed", nfcRollout.Generation)
 		}
+
```
juliev0, line 523: Updated this function a little bit. I prefer to check the `NumaflowController.Status.ObservedGeneration` as opposed to the `ObservedGeneration` of the Condition to know whether NumaflowController has been reconciled

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -563,10 +563,16 @@ func (r *PipelineRolloutReconciler) processPipelineStatus(ctx context.Context, p
...
+		pipelineName, err := progressive.GetChildName(ctx, pipelineRollout, r, common.LabelValueUpgradePromoted, r.client, true)
 		if err != nil {
-			return err
+			return fmt.Errorf("Unable to process pipeline status: err=%s", err)
 		}
+		pipelineDef := &unstructured.Unstructured{}
+		pipelineDef.SetGroupVersionKind(numaflowv1.PipelineGroupVersionKind)
+		pipelineDef.SetNamespace(pipelineRollout.Namespace)
+		pipelineDef.SetName(pipelineName)
+
```
juliev0, line 575: This is admittedly not related to the rest of the PR, but is another bug I found, and I wanted to be able to do comprehensive testing of everything together, which is why I kept it in one PR. The previous call to `makePromotedPipelineDefinition()` doesn't work in the case of the ISBServiceRollout having already been deleted. So, I didn't want to have to depend on that function as part of processing Pipeline status.

tests/e2e/functional_test.go
```diff
@@ -211,7 +212,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		verifyNumaflowControllerRolloutReady()
 
-		verifyNumaflowControllerReady(Namespace)
+		verifyNumaflowControllerExists(Namespace)
```
juliev0, line 215: rename for what the function is doing

tests/e2e/functional_test.go
```diff
@@ -395,9 +396,9 @@ var _ = Describe("Functional e2e", Serial, func() {
 			if err != nil {
 				return false
 			}
-			return getRolloutCondition(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused) == metav1.ConditionTrue &&
+			return getRolloutConditionStatus(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused) == metav1.ConditionTrue &&
 				(retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePaused || retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePausing)
-		}, 1*time.Minute, testPollingInterval).Should(BeTrue())
+		}, 15*time.Second, testPollingInterval).Should(BeTrue())
```
juliev0, line 401: previous 1 minute wait was creating unnecessary delay in the e2e

tests/e2e/functional_test.go
```diff
@@ -493,47 +494,19 @@ var _ = Describe("Functional e2e", Serial, func() {
 	time.Sleep(2 * time.Second)
 
 	It("Should update the child NumaflowController if the NumaflowControllerRollout is updated", func() {
+		updateNumaflowControllerRolloutVersion(initialNumaflowControllerVersion, updatedNumaflowControllerVersion, true)
```
juliev0, line 497: consolidated shared functionality

tests/manifests/default/controller_def_1.3.3.yaml
```diff
@@ -14,29 +14,25 @@ data:
           kind: ServiceAccount
           metadata:
             labels:
-              app.kubernetes.io/instance: '{{ .InstanceID }}'
```
juliev0, line 17: Removed these "instance" labels from the 1.3.3 manifest - this effectively enables the e2e test to duplicate the issue we had before in the clusters - moving from one version which did not have the labels to a version which did

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
juliev0: I believe Server Side Diff and Server Side Apply are used in the case that multiple parties may be modifying the same resource, which we don't need to deal with.

afugazzotto (approved):

PR #477, opened 2025-01-06
feat: Blue/Green for isbsvc 

### LINKED ISSUE #486: Blue/Green for isbsvc
# Summary

What change needs making?

# Use Cases

When would you use this?

---

juliev0:
Fixes #486 

### Modifications

The isbsvc now executes the blue/green path similar to Pipeline and Monovertex. The main differences are:

- when we create an "upgrading" isbsvc, there needs to be an "upgrading" pipeline as well which uses it. This is done by the PipelineRollout reconciler - if it observes an "upgrading" isbsvc, it will make sure to create an "upgrading" pipeline using it
- the ISBServiceRollout's health assessment is based on the health assessment of the underlying pipelines (will need to expand this to also include isbsvc's health itself). The ISBServiceRollout reconciler gets this information from the new `ProgressiveStatus.UpgradingChildStatus` information in the PipelineRollout's Status.
- the ISBServiceRollout's "recycle" logic only performs deletion once all pipelines using it have been deleting it

### Remaining Items

- There are some TODOs called out in the code.
- Would like to add a unit test for isbsvc progressive case
- Want to test end to end in ArgoCD and see if I can break this at all

### Verification

* e2e test now tests the progressive case - this is the happy path for isbservice updates (passes every time)
* manual tests:
    * isbsvc and pipeline updated at the same time, specifically {{isbsvc: data loss field, pipeline: data loss field}, {isbsvc: data loss field, pipeline: regular field}, {isbsvc: regular field, pipeline: regular field}   
    * isbsvc updated with 2 pipeline updates at the same time; one pipeline fails and the other succeeds; user fixes pipeline

I didn't see anything that would be a problem but decided to make sure by testing. I installed 0.12.1 locally and ran it in ppnd mode (since that's what we're using on the clusters). Then I created an ISBServiceRollout and PipelineRollout. Then I updated to this branch. The only thing that happened was the existing pipeline got the label: `numaplane.numaproj.io/isbsvc-child-name`, which was expected.


config/install.yaml
```diff
@@ -1868,26 +1949,6 @@ spec:
...
-          isbsvc\n  for isbsvc in $interstepbuffersvcs; do\n    echo \"Patching isbsvc:
-          $isbsvc in namespace: $namespace\"\n    \n    # Patch the isbsvc to add
-          the labels\n    kubectl patch isbsvc $isbsvc -n $namespace --type='merge'
-          -p \"{\\\"metadata\\\": {\\\"labels\\\": {\\\"numaplane.numaproj.io/upgrade-state\\\":
-          \\\"promoted\\\"}}}\"\n    kubectl patch isbsvc $isbsvc -n $namespace --type='merge'
-          -p \"{\\\"metadata\\\": {\\\"labels\\\": {\\\"numaplane.numaproj.io/parent-rollout-name\\\":
-          \\\"$isbsvc\\\"}}}\"\n  done\ndone\n\necho \"All interstepbufferservices
-          have been patched.\"\n"
-        image: bitnami/kubectl
-        name: label-isbsvc
```
juliev0, line 1890: This initContainer was added to numaplane 0.12.0, which is being rolled out now to clusters. It helped the transition to the use of new labels "parent-rollout-name" and "upgrade-state" for isbsvc in the clusters without backward compatibility issues. However, before we actually have blue/green operational, we need to remove this Initcontainer since it will create the wrong label for "parent-rollout-name".

juliev0: Therefore, we must release numaplane 0.12.0 (or otherwise some patch 0.12.x) to all clusters which adds the labels prior to rolling this PR (which removes the initcontainer) out to clusters.

tests/e2e/functional_test.go
```diff
@@ -85,6 +85,7 @@ var (
 			},
 		},
 	}
+	numPipelineVertices = 2
```
juliev0, line 88: using a variable instead of a "magic number" in the code

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	"github.com/numaproj/numaplane/internal/controller/progressive"
+	"github.com/numaproj/numaplane/internal/util/kubernetes"
+	"github.com/numaproj/numaplane/internal/util/logger"
+	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// TODO: can move other progressive interface functions here as well? (same with PipelineRollout, ISBServiceRollout)
```
juliev0, line 13: I'll do this in a follow up PR - didn't want to make this harder to read by doing it in this one

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -0,0 +1,35 @@
...
+		Debugf("Upgrading child is in phase %s", upgradingObjectStatus.Phase)
+
+	if upgradingObjectStatus.Phase == "Running" && progressive.IsNumaflowChildReady(&upgradingObjectStatus) {
+		return apiv1.AssessmentResultSuccess, nil
+	}
+	if upgradingObjectStatus.Phase == "Failed" {
+		return apiv1.AssessmentResultFailure, nil
+	}
+	return apiv1.AssessmentResultUnknown, nil
+}
```
juliev0, line 35: This used to be a single function in progressive.go, but it's now part of the progressiveController interface, and all 3 can have different implementations (although, if it makes sense they can use some shared function for common functionality).

internal/controller/progressive/progressive.go
```diff
@@ -246,110 +253,85 @@ func GetChildName(ctx context.Context, rolloutObject ctlrcommon.RolloutObject, c
...
+		return false, false, err
 	}
 	numaLogger.WithValues("name", existingUpgradingChildDef.GetName()).Debugf("assessment returned: %v", assessment)
 
 	switch assessment {
-	case AssessmentResultFailure:
+	case apiv1.AssessmentResultFailure:
 
 		rolloutObject.GetRolloutStatus().MarkProgressiveUpgradeFailed(fmt.Sprintf("New Child Object %s/%s Failed", existingUpgradingChildDef.GetNamespace(), existingUpgradingChildDef.GetName()), rolloutObject.GetRolloutObjectMeta().Generation)
+		rolloutObject.GetRolloutStatus().ProgressiveStatus.UpgradingChildStatus = &apiv1.ChildStatus{Name: existingUpgradingChildDef.GetName(), AssessmentResult: apiv1.AssessmentResultFailure}
```
juliev0, line 279: this information from the PipelineRollout is what informs the ISBServiceRollout reconciler as to what happened.

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -363,9 +364,30 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 			// requeue if done with PPND is false
 			return true, nil
 		}
-	// TODO: Progressive strategy should ideally be creating a second parallel isbsvc, and all Pipelines should be on it;
-	// for now we just create a 2nd ISBServiceRollout, so we need the Apply path to work
-	case apiv1.UpgradeStrategyApply, apiv1.UpgradeStrategyProgressive:
+	case apiv1.UpgradeStrategyProgressive:
+		numaLogger.Debug("processing InterstepBufferService with Progressive")
+		done, _, err := progressive.ProcessResourceWithProgressive(ctx, isbServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
```
afugazzotto, line 369: In the func name `ProcessResourceWithProgressive` the `WithProgressive` part seems redundant since the package is already named `progressive`.
Also, instead of passing `r` and `r.client`, could you just pass `r` and get the client within the func?

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -377,6 +399,11 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 	default:
 		return false, fmt.Errorf("%v strategy not recognized", inProgressStrategy)
 	}
+	// clean up recyclable interstepbufferservices
+	err = progressive.GarbageCollectChildren(ctx, isbServiceRollout, r, r.client)
```
afugazzotto, line 403: Similar comment here as above about `r` and `r.client`

afugazzotto (commented):
Some minor comments, rest looks good.

internal/controller/isbservicerollout/isbservicerollout_controller.go
juliev0: yeah, I can change the name and remove the redundancy. 

> Also, instead of passing r and r.client, could you just pass r and get the client within the func?

In that case, `Client()` needs to be made part of the `progressiveController` interface - do you think that's better?

afugazzotto: nvm, I don't like that idea either. Leave it as a func arg.

afugazzotto (approved):

internal/controller/isbservicerollout/isbservicerollout_controller.go
juliev0: okay, I'll fix the name and then merge this, thanks

Issue #402, opened 2024-11-12
USDE for Progressive, Delete/Recreate


juliev0:
# Summary

For Progressive, there are 3 things to consider:
1. "recreate" fields: Some fields are only safe to change with Progressive and not PPND - this means that even if the user prefers PPND, it should be done with Progressive (or there will be completely breakage otherwise) - the alternative is to delete and re-create (basically, they require a recreation in either case). There can be a "recreateStrategy" for this, which can either be "progressive" or "delete-recreate". And actually if the user's "upgradeStrategy" is "ppnd", I think "delete-recreate" should actually mean: "pause, and then delete/recreate"
2. "data-loss" fields: We still care about all of the data loss fields, which would indicate that we should do "Progressive" rather than "Direct Apply". This is still determined by "upgradeStrategy". 
3. "progressive-desired" fields: if the Image changes, that's a good time to do Progressive, since this is when we can most benefit from the Blue/Green aspect of Progressive. This would be used if `upgradeStrategy==progressive`. It is only really meaningful for MonoVertex, since for Pipeline we will already do Progressive, since it's a "data loss" field. 

**Examples of Case 1:**

***Pipeline:***
- InterstepBufferServiceName (we should move this from the data loss fields to here)

***isbsvc:***

- provider change: jetstream <-> redis
- replicas
- PVC
- Any field change that is not allowed in a statefulset (not sure what this includes)
 

## Restructuring of the USDE:
I'm thinking something like this...but open to improvements on naming...
```
dataLoss:                      # case 2: these are the fields we have now
   pipeline:
   interstepbufferservice:
recreate:  # case 1
   pipeline:
   interstepbufferservice:
progressive:    # case 3
   pipeline:
   monovertex:
```

**Configurability in numaplane-controller-config.yaml/user strategy config:**
In addition to `defaultUpgradeStrategy`, we can also have `defaultRecreateStrategy` for case 1 above (either "delete-recreate" or "progressive").

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
I'm thinking we would in our [usde_config.yaml](https://github.com/numaproj/numaplane/blob/main/config/manager/usde-config.yaml) file have 2 lists instead of just 1. 

Currently, we have one list for "PPND" - or we could think of it as a "data loss" list. 

We would need to have 2 lists instead. I am thinking either of these options:
[...]

PR #479, opened 2025-01-07
feat: changed USDE config to include more than just data-loss fields

afugazzotto:
### Modifications

Changed USDE config to include more than just data-loss fields (recreate and progressive fields).
This PR only includes the changes to modify the USDE config structure. Future PR will include changes to the logic to utilize those new configuration fields and to properly setup the configuration spec fields.

### Verification

Manually tested that the USDE config structure is loaded correctly. Updated unit tests to reflect the new structure.

At this point, this PR should not cause backward incompatibility issues unless the USDE ConfigMap cannot be updated during upgrade.

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -1737,72 +1737,89 @@ metadata:
 ---
 apiVersion: v1
 data:
-  isbServiceSpecDataLossFields: |
-    # Principal: Update to StatefulSet may not be backward compatible to understand the data on the PVC
-    - path: spec.jetstream.version
-    - path: spec.redis.native.version
-    - path: spec.redis.external.url
-    - path: spec.redis.external.sentinelUrl
-    - path: spec.redis.external.masterName
-  pipelineSpecDataLossFields: |
-    # Principal: Pipeline topology change could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed
-    - path: spec.edges.from
-    - path: spec.edges.to
-    - path: spec.edges.conditions
-      includeSubfields: true
+  isbService: |
+    dataLoss:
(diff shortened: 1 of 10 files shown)


afugazzotto:
@juliev0 I decided to change the USDE config structure slightly from [your proposal](https://github.com/numaproj/numaplane/issues/402). Instead of having `dataLoss`, `recreate`, and `progressive` on the top level, I decided to have the resources at the top level. In my mind this makes more sense to reuse the path/includeSubfields struct under each of the type of strategy we want to define.

config/install.yaml
```diff
@@ -1737,72 +1737,89 @@ metadata:
...
+      # Principal: Pipeline topology change could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed
+      - path: spec.edges.from
+      - path: spec.edges.to
+      - path: spec.edges.conditions
+        includeSubfields: true
 
-    # Principal: Changing the InterstepBufferService could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed
-    - path: spec.interStepBufferServiceName
+      # Principal: Changing the InterstepBufferService could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed
+      - path: spec.interStepBufferServiceName
```
juliev0, line 1768: note this one will get moved to being a `recreate` field - until that part's ready, it's fine to keep it here

afugazzotto: I was thinking to move spec fields and add new ones in another PR to keep USDE config structure changes separated from logic changes.

juliev0 (approved):

PR #481, opened 2025-01-07
feat: add progressive to e2e test matrix

### LINKED ISSUE #474: Add Progressive to our e2e test matrix
# Summary

Let's have 3 tests instead of 2. We can replace `PPND=true` and `PPND=false` with something like `STRATEGY={progressive, pause-and-drain, no-strategy}`

---

dpadhiar:
Fixes #474 

### Modifications

Incorporates `progressive` strategy to e2e test matrix and redefines matrix to be 3 different runs: `no-strategy`, `pause-and-drain` and `progressive`

### Verification

Running `make test-e2e` with `STRATEGY` variable set locally to verify change is working.
Checking that the CI's matrix is updated correctly to reflect new matrix.

None.

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -390,7 +390,7 @@ func getChildResource(gvr schema.GroupVersionResource, namespace, rolloutName st
 
 func getUpgradeStrategy() string {
 
-	var validStrategies = []string{"ppnd", "progressive", "no-strategy", ""}
+	var validStrategies = []string{"pause-and-drain", "progressive", "no-strategy", ""}
 
 	strategy := strings.ToLower(os.Getenv("STRATEGY"))
 	for _, validStrategy := range validStrategies {
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -53,7 +53,7 @@ var (
 	mutex  sync.RWMutex
 	stopCh chan struct{}
 
-	ppnd                 string
+	upgradeStrategy      string
 	disableTestArtifacts string
 	enablePodLogs        string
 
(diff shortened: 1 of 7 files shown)


tests/e2e/common.go
```diff
@@ -387,3 +387,17 @@ func getChildResource(gvr schema.GroupVersionResource, namespace, rolloutName st
 	return &unstructList.Items[0], nil
 
 }
+
+func getUpgradeStrategy() string {
+
+	var validStrategies = []string{"pause-and-drain", "progressive", "no-strategy", ""}
```
afugazzotto, line 393: Could you use [these constants](https://github.com/numaproj/numaplane/blob/main/internal/controller/config/user_config.go#L28) instead?

afugazzotto: Consider using them also in the other .go files

tests/e2e/functional_test.go
```diff
@@ -579,8 +579,14 @@ var _ = Describe("Functional e2e", Serial, func() {
...
-			if isbRollout.Status.UpgradeInProgress != apiv1.UpgradeStrategyNoOp || plRollout.Status.UpgradeInProgress != apiv1.UpgradeStrategyNoOp {
-				return false
+			if upgradeStrategy == "progressive" {
+				if isbRollout.Status.UpgradeInProgress != apiv1.UpgradeStrategyProgressive || plRollout.Status.UpgradeInProgress != apiv1.UpgradeStrategyNoOp {
+					return false
+				}
+			} else {
+				if isbRollout.Status.UpgradeInProgress != apiv1.UpgradeStrategyNoOp || plRollout.Status.UpgradeInProgress != apiv1.UpgradeStrategyNoOp {
+					return false
+				}
```
juliev0, line 589: maybe it could improve readability if we remove this entire block of code that you've changed here, and instead make it so we only do this entire "Verify that dependent Pipeline is not paused when an update to ISBService not requiring pause is made" section if we're doing PPND? 

I think this test doesn't make sense if we're using another strategy.

tests/e2e/common.go
juliev0: agree

juliev0 (approved):

PR #487, opened 2025-01-09
feat: modified USDE logic to include USDE Config changes with the new progressive and recreate lists

### LINKED ISSUE #402: USDE for Progressive, Delete/Recreate

afugazzotto:
Fixes #402 

### Modifications

Modified the USDE logic to include USDE Config changes with the new `progressive` and `recreate` lists.

### Verification

Unit tests and e2e tests.

This PR should not cause backward incompatibility issues unless the USDE ConfigMap cannot be updated during upgrade.

### DIFF
--- a/config/manager/usde-config.yaml
+++ b/config/manager/usde-config.yaml
@@ -6,6 +6,10 @@ metadata:
     numaplane.numaproj.io/config: usde-config
 data:
   pipeline: |
+    recreate:
+      # Principal: Changing the InterstepBufferService could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed.
+      # Furthermore, this change needs to recreate the pipeline so that ISBSvc buckets and buffers are also recreated.
+      - path: spec.interStepBufferServiceName
     dataLoss:
       # Principal: Pipeline topology change could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed
       - path: spec.edges.from
@@ -62,18 +66,7 @@ data:
       # Principal: TBD
       - path: spec.sideInputs.container.envfrom
         includeSubfields: true
-    recreate:
-      # Principal: Changing the InterstepBufferService could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed.
-      # Furthermore, this change needs to recreate the pipeline so that ISBSvc buckets and buffers are also recreated.
-      - path: spec.interStepBufferServiceName
   interstepbufferservice: |
-    dataLoss:
-      # Principal: Update to StatefulSet may not be backward compatible to understand the data on the PVC
-      - path: spec.jetstream.version
-      - path: spec.redis.native.version
-      - path: spec.redis.external.url
-      - path: spec.redis.external.sentinelUrl
-      - path: spec.redis.external.masterName
     recreate:
       # Changing persistence settings and number of replicas needs to recreate both the ISBSvc and also the pipelines associated to it to 
       # recreate the buckets and buffers which are only recreated when a pipeline is started.
@@ -87,6 +80,13 @@ data:
       # This checks if the spec changed from jetstream to redis and viceversa. In this case, the ISBSvc and associated pipelines will need to be recreated.
       - path: spec.jetstream
       - path: spec.redis
+    dataLoss:
+      # Principal: Update to StatefulSet may not be backward compatible to understand the data on the PVC
+      - path: spec.jetstream.version
+      - path: spec.redis.native.version
+      - path: spec.redis.external.url
+      - path: spec.redis.external.sentinelUrl
+      - path: spec.redis.external.masterName
   monovertex: |
     progressive:
       - path: spec.initContainers.image
--- a/config/manager/usde-config.yaml
+++ b/config/manager/usde-config.yaml
@@ -115,4 +115,4 @@ data:
       - path: spec.source.udsource.container.env
   numaflowcontroller: |
     dataLoss:
-      - path: spec.controller.version
+     - path: spec.version
--- a/config/manager/usde-config.yaml
+++ b/config/manager/usde-config.yaml
@@ -83,6 +83,9 @@ data:
         includeSubfields: true
       - path: spec.redis.native.persistence
         includeSubfields: true
+      
+      # This checks if the spec changed from jetstream to redis and viceversa. In this case, the ISBSvc and associated pipelines will need to be recreated.
+      - path: spec
   monovertex: |
     progressive:
       - path: spec.initContainers.image
@@ -110,7 +113,7 @@ data:
       - path: spec.source.udsource.container.args
       - path: spec.source.udsource.container.env
   numaflowcontroller: |
-    # Idependently of the USDE strategy, the NumaflowController will be deleted and recreated even if the USDE logic will not specify to recreate.
+    # Independently of the USDE strategy, the NumaflowController will be deleted and recreated even if the USDE logic will not specify to recreate.
     # The NumaflowController controller will always process a USDE response to update the spec by recreating the NumaflowController.
     recreate:
       - path: spec
--- a/config/manager/usde-config.yaml
+++ b/config/manager/usde-config.yaml
@@ -13,9 +13,6 @@ data:
       - path: spec.edges.conditions
         includeSubfields: true
 
-      # Principal: Changing the InterstepBufferService could cause data loss since unconsumed data may be left on the InterstepBufferService which won't get processed
-      - path: spec.interStepBufferServiceName
-
       # Principal: Changing an Image definition could cause data loss if the new Image is not able to process the old data from the InterstepBufferService
       - path: spec.vertices.source.generator
       - path: spec.vertices.source.kafka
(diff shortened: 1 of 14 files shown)


config/manager/usde-config.yaml
```diff
@@ -78,13 +75,43 @@ data:
...
+      - path: spec.source.transformer.container.image
+      - path: spec.source.transformer.container.command
+      - path: spec.source.transformer.container.args
+      - path: spec.source.transformer.container.env
+      - path: spec.source.udsource.container.image
+      - path: spec.source.udsource.container.command
+      - path: spec.source.udsource.container.args
+      - path: spec.source.udsource.container.env
+  numaflowcontroller: |
+    # Idependently of the USDE strategy, the NumaflowController will be deleted and recreated even if the USDE logic will not specify to recreate.
```
juliev0, line 113:
```suggestion
    # Independently of the USDE strategy, the NumaflowController will be deleted and recreated even if the USDE logic will not specify to recreate.
```

internal/controller/config/usde_config.go
```diff
@@ -28,16 +28,12 @@ type SpecField struct {
 }
 
 type USDEResourceConfig struct {
-	DataLoss    []SpecField `json:"dataLoss" yaml:"dataLoss"`
-	Recreate    []SpecField `json:"recreate" yaml:"recreate"`
-	Progressive []SpecField `json:"progressive" yaml:"progressive"`
+	DataLoss    []SpecField `json:"dataLoss,omitempty" yaml:"dataLoss,omitempty"`
+	Recreate    []SpecField `json:"recreate,omitempty" yaml:"recreate,omitempty"`
+	Progressive []SpecField `json:"progressive,omitempty" yaml:"progressive,omitempty"`
```
juliev0, line 33: We should add comments here for these and what they mean to do, including the order of priority.

config/manager/usde-config.yaml
```diff
@@ -78,13 +75,43 @@ data:
...
-    progressive:
-      # TODO: add the progressive fields
+      # Changing persistence settings and number of replicas needs to recreate both the ISBSvc and also the pipelines associated to it to 
+      # recreate the buckets and buffers which are only recreated when a pipeline is started.
+      - path: spec.jetstream.replicas
+      - path: spec.redis.native.replicas
+      - path: spec.jetstream.persistence
+        includeSubfields: true
+      - path: spec.redis.native.persistence
+        includeSubfields: true
```
juliev0, line 85: nit: if we're going to include redis fields in here, we may as well also include the changing from redis to jetstream and jetstream to redis (even though Derek said it's "not allowed", that was only because redis doesn't have all of the features)

config/manager/usde-config.yaml
```diff
@@ -78,13 +75,43 @@ data:
...
+      - path: spec.source.udsource.container.image
+      - path: spec.source.udsource.container.command
+      - path: spec.source.udsource.container.args
+      - path: spec.source.udsource.container.env
+  numaflowcontroller: |
+    # Idependently of the USDE strategy, the NumaflowController will be deleted and recreated even if the USDE logic will not specify to recreate.
+    # The NumaflowController controller will always process a USDE response to update the spec by recreating the NumaflowController.
+    recreate:
+      - path: spec
+        includeSubfields: true
```
juliev0, line 117: Hmm, need to think about if we want this. I can see the motivation for it, as it has better symmetry with everything else. But on the other hand, if we were try to actually make use of this, we would need to enable it in the code. This means that the `NumaflowControllerRollout` would need to set something in the `NumaflowController` for this, and the NumaflowController would need to have logic to either delete/recreate or not. This seems like a bunch of code that would probably not even be used.

internal/usde/usde.go
```diff
@@ -47,107 +49,163 @@ func ResourceNeedsUpdating(ctx context.Context, newDef, existingDef *unstructure
...
+		}
+
+		// Use the progressive fields list from config and return direct-apply strategy if the spec needs to update
+		specNeedsUpdating, err = checkFieldsList(ctx, progressiveFields, newDef, existingDef)
+		if err != nil {
+			return false, apiv1.UpgradeStrategyError, false, fmt.Errorf("error while checking spec changes using 'progressive' USDE Config (strategy '%s'): %w", upgradeStrategy, err)
+		}
+		if specNeedsUpdating {
+			return specNeedsUpdating, apiv1.UpgradeStrategyApply, false, nil
+		}
```
juliev0, line 134: nit: I'm thinking we should be able to remove this block? It will be handled in the check for "all fields" below, right?

internal/usde/usde.go
```diff
@@ -47,107 +49,163 @@ func ResourceNeedsUpdating(ctx context.Context, newDef, existingDef *unstructure
...
+		if specNeedsUpdating {
+			// Also return "recreate" true to communicate to the controller to recreate the appropriate resources
+			return specNeedsUpdating, upgradeStrategy, true, nil
+		}
+	}
+
+	numaLogger.Debug("no USDE Config field changes detected, comparing specs for any DirectApply-type of changes")
+
+	// If there were no changes in the data loss fields, there could be changes in other fields of the specs.
```
juliev0, line 148: nit: update comment to account for the other types of fields, not just "data loss"

juliev0 (commented):
looks good overall!

config/manager/usde-config.yaml
juliev0: Also, it’s not like we’re deleting and recreating the NumaflowController. So, it’s different.

afugazzotto: I decided to implement it this way to avoid a special exception in the code just for the NumaflowController and to maintain the consistency with the other resources. If we want to eliminate this config for NumaflowController we can do one of the following:
1. create an exception in code
2. create a fourth list just for the NumaflowController case (similar to the Monovertex case)

If you go with option 1, I don't see why we wouldn't do the same thing (creating an exception in code) also for the Monovertex.

internal/usde/usde.go
afugazzotto: Yes, this is related to [this comment](https://github.com/numaproj/numaplane/pull/487#discussion_r1910779105).

config/manager/usde-config.yaml
juliev0: Okay, here's what I'm thinking - the "dataLoss" fields do apply to NumaflowControllerRollout, but "recreate" fields do not. We never recreate the NumaflowController.

juliev0: `Version` can be a data loss field. 

I'm thinking about `InstanceID`. If you change it, then all of the ISBServiceRollouts, PipelineRollouts, and MonoVertexRollouts really ought to be updated at the same time with the new InstanceID label. Say somebody does actually change those at the same time in github. What would happen if they did PPND? They'd be pausing the pipeline during both the Pipeline change and during the Numaflow Controller change (for which the times may or may not overlap depending on timing). Really they only need to pause the pipeline during the Pipeline update, however. It doesn't need to be paused during the Numaflow Controller change - since the version isn't changing, then the "numa" container will remain as-is, and there will be no data loss issue.

Therefore, I think we only need to list `version` as a data loss field and that's it.

internal/controller/config/usde_config.go
juliev0: sorry for the nit, but do you mind reordering these in order of their priority? I think it would help readability for this

afugazzotto: Reordered also in other files for consistency.

juliev0 (approved):

PR #488, opened 2025-01-09
prevent panic in e2e test

juliev0:
### Modifications

I realized that if the e2e test panics, it does it silently. It doesn't print the panic. It just fails the test. The thing I noticed is that it doesn't perform the `Eventually()` statement for the full 2 minute timeout like it's supposed to, which gave me a clue that this was happening.


chandankumar4 (approved):
looks good

PR #499, opened 2025-01-14
feat: delay child resource status assessment during progressive rollout to allow some time before promoting the upgraded Pipeline/MonoVertex

### LINKED ISSUE #445: Current check for Pipeline/Monovertex health assumes Conditions will remain healthy if they're initially healthy
**Describe the bug**
Current check for Pipeline/Monovertex health assumes that once all Conditions are "true" plus Pipeline/monovertex is "Running", that it's healthy. However, it can momentarily be healthy and then become unhealthy after processing some data. 

Therefore, we should check that the Pipeline/Monovertex enters the healthy condition by some time x and then remains healthy by some time x + y. 

(Note, currently the only Failed Pipelines/Monovertices that have been tested have had invalid specs, so they were marked "Failed". So, we haven't tested any with failing Conditions. We need to test as many types of failed child health conditions as we can.)

---

afugazzotto:
Fixes #445 

### Modifications

Added a new field called `NextAssessmentTime` to the `Status.ProgressiveStatus.UpgradingChildStatus` to store a future time when to perform the child resource status assessment to determine its healthiness.
During reconciliation in both the PipelineRollout and MonoVertexRollout controllers, we'll get the live rollout status (to avoid having stale info from the cache) and check if the `NextAssessmentTime` has already been set, meaning that the "timer" has been started. If the `NextAssessmentTime` has not been set, we use the delay from ConfigMap to calculate the time and set it on the rollout.
Then, we check if the `NextAssessmentTime` has passed to determine if it is time to assess the child status for healthiness.

### Verification

Unit tests, E2E tests, and manual tests.

This change should not cause backward incompatibility issues.


pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -227,6 +229,25 @@ func (status *Status) ClearUpgradeInProgress() {
...
+// non-nil and NextAssessmentTime is not the zero value of time.Time.
+func (status *Status) NextAssessmentTimeHasBeenSet() bool {
+	return status != nil && status.ProgressiveStatus.UpgradingChildStatus != nil && status.ProgressiveStatus.UpgradingChildStatus.NextAssessmentTime != metav1.NewTime(time.Time{})
+}
+
+// CanAssessChildStatus determines if the child status can be assessed.
+// It checks if the Status object is not nil, the UpgradingChildStatus is not nil,
+// the NextAssessmentTime is in the future, and the AssessmentResult is unknown.
+// If the AssessmentResult is not unknown and the NextAssessmentTime is in the future
+// we should not check again since the assessment has already been performed.
```
afugazzotto, line 244: @juliev0 do you think this assumption is valid? Is it a good idea to also check the AssessmentResult and if it is not unknown we should not have to return true since the status has been assessed after the delay?
If we do not want to check also the AssessmentResult, I think it's ok since the rest of the code should handle the returned value correctly anyways.

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -526,6 +526,25 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
 	case apiv1.UpgradeStrategyProgressive:
 		numaLogger.Debug("processing pipeline with Progressive")
 
+		// Get the PipelineRollout live resource
```
afugazzotto, line 529: Decided to move this code inside the Progressive case since it is only relevant to Progressive strategy.

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -526,6 +526,25 @@ func (r *PipelineRolloutReconciler) processExistingPipeline(ctx context.Context,
...
+		if !livePipelineRollout.Status.NextAssessmentTimeHasBeenSet() {
+			// Get the delay from Numaplane ConfigMap
+			globalConfig, err := config.GetConfigManagerInstance().GetConfig()
+			if err != nil {
+				return false, fmt.Errorf("error getting the global config for assessment processing: %w", err)
+			}
+			delay := time.Duration(globalConfig.ChildStatusAssessmentDelaySeconds) * time.Second
+
+			// Add to the current time the delay and set the NextAssessmentTime in the PipelineRollout
+			pipelineRollout.Status.ProgressiveStatus.UpgradingChildStatus.NextAssessmentTime = metav1.NewTime(time.Now().Add(delay))
```
afugazzotto, line 545: @juliev0 do you think we should replace/merge the entire `pipelineRollout` object with the `livePipelineRollout` or could just update the `NextAssessmentTime` be enough?

juliev0: I'd like to update the entire `pipelineRollout`

juliev0: I was actually kind of hoping all of the assessment-related code could be collocated inside of `AssessUpgradingChild()` unless there's a reason that it shouldn't be (sorry, should've said that if it was in my head) - what do you think?

juliev0: BTW, if you saw my original comment, I deleted that because I see that you're trying to include as much in the shared progressive.go code as you can, so that's why you didn't put the code directly in `AssessUpgradingChild()`. That seems good.

pkg/apis/numaplane/v1alpha1/status_types.go
juliev0: thanks for asking. Yes, I am thinking we should not check `status.ProgressiveStatus.UpgradingChildStatus.AssessmentResult == AssessmentResultUnknown`. It is fine to continue to assess after the delay has occurred.

juliev0: BTW, one thing to be aware of (which you may already be) - ISBServiceRollout is also using all of this shared code. And its assessment logic is based on the assessment result of the underlying pipelines.

juliev0: I am kind of thinking maybe we should do this check:
```
status.ProgressiveStatus.UpgradingChildStatus.AssessmentResult != AssessmentResultSuccess
```
This is because:
1. I think it's fine to mark it as Failed and then later mark it as Successful
2. It's not a good idea to mark it as Successful and then later mark it as Failed, since once it's marked Successful, it's already started the process of promoting that pipeline, and we don't want to interfere with that

internal/controller/pipelinerollout/pipelinerollout_controller.go
juliev0: > I'd like to replace the entire pipelineRollout

Changing my mind after you made the astute realization that we may have changed the Status value locally in the code and we don't want to overwrite anything. Okay, I think I'm good with this approach.

config/manager/controller-config.yaml
```diff
@@ -9,3 +9,4 @@ data:
     group=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=Service;\
     group=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role"
     defaultUpgradeStrategy: "pause-and-drain"
+    childStatusAssessmentDelaySeconds: 60
```
afugazzotto, line 12: @juliev0 the delay of 30 seconds is not enough for ISBSvc (it is enough for Pipeline and MonoVertex). I decided to increase the delay for all to 60s. However, I think we could go a step further and have 2 or more delay settings for each of the resource types (ex: 30s for pipeline, 20s for monovertex, etc.). We would need to figure out for each resource type what is the ideal delay and maybe add a margin of seconds to account for system processing delays. What do you think?

juliev0: What's the result of the delay not being long enough? It should still work, right? If the delay has arrived and the result is still Unknown, then that should be fine and it should still do further reconciliations.

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -227,6 +229,37 @@ func (status *Status) ClearUpgradeInProgress() {
...
+		status.ProgressiveStatus.UpgradingChildStatus.NextAssessmentTime = nextAssessmentTime
+	}
+}
+
+// NextAssessmentTimeHasBeenSet checks if the NextAssessmentTime field in the
+// UpgradingChildStatus of the ProgressiveStatus is set to a non-zero value.
+// Returns true if the Status, ProgressiveStatus, and UpgradingChildStatus are
+// non-nil and NextAssessmentTime is not the zero value of time.Time.
+func (status *Status) NextAssessmentTimeHasBeenSet() bool {
+	return status != nil && status.ProgressiveStatus.UpgradingChildStatus != nil && status.ProgressiveStatus.UpgradingChildStatus.NextAssessmentTime != metav1.NewTime(time.Time{})
```
juliev0, line 249: does this work for going from pipeline-0 to pipeline-1 to pipeline-2? I don't think there's anything that clears out `status.ProgressiveStatus.UpgradingChildStatus` or `status.ProgressiveStatus.UpgradingChildStatus.NextAssessmentTime` when there's no longer an upgrading child, is there? So, my guess would be that this would always return true if we had an Upgrading child in the past. Correct me if I'm wrong.

afugazzotto: That's correct. I have not completed everything yet. I was thinking to set `UpgradingChildStatus = nil` either [after garbage collecting the children](https://github.com/numaproj/numaplane/blob/main/internal/controller/progressive/progressive.go#L377) or whenever the `UpgradeInProgress` gets cleared `rollout.Status.UpgradeInProgress = apiv1.UpgradeStrategyNoOp`. Any preference?

config/manager/controller-config.yaml
afugazzotto: I'm looking into this further. Strangely increasing the time passed the E2E tests but I found also some other issues and changed it back to see if it works again at 30s.

juliev0: In answer to your other question, it is reasonable to potentially have different delay settings for each resource type, if we want to do that.

pkg/apis/numaplane/v1alpha1/status_types.go
juliev0: Yeah, maybe when we are about to clear the UpgradeInProgress [here](https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout/pipelinerollout_controller.go#L534)?

internal/controller/progressive/progressive.go
```diff
@@ -253,19 +262,73 @@ func GetChildName(ctx context.Context, rolloutObject ctlrcommon.RolloutObject, c
...
+	// Create a new childStatus object if not present in the live rollout object or
+	// if it is that of a previous progressive upgrade.
+	if childStatus == nil || childStatus.Name != existingUpgradingChildDef.GetName() {
+		childStatus = &apiv1.ChildStatus{
+			Name:             existingUpgradingChildDef.GetName(),
+			AssessmentResult: apiv1.AssessmentResultUnknown,
+		}
+		numaLogger.WithValues("childStatus", childStatus).Debug("live upgrading child not yet set")
+	} else {
+		numaLogger.WithValues("childStatus", *childStatus).Debug("live upgrading child previously set")
```
juliev0, line 304: did you mean for the first one to also be `numaLogger.WithValues("childStatus", *childStatus)` (with asterisk)?

internal/controller/progressive/progressive.go
```diff
@@ -253,19 +262,73 @@ func GetChildName(ctx context.Context, rolloutObject ctlrcommon.RolloutObject, c
...
+		}
+		delay := time.Duration(globalConfig.ChildStatusAssessmentDelaySeconds) * time.Second
+
+		// Add to the current time the delay and set the NextAssessmentTime in the Rollout object
+		childStatus.NextAssessmentTime = metav1.NewTime(time.Now().Add(delay))
+		numaLogger.WithValues("childStatus", *childStatus).Debug("set upgrading child nextAssessmentTime")
+	}
+
+	// Use the NextAssessmentTime to check if it's time to assess the child resource status.
+	// Only assess the child if if the NextAssessmentTime is after the current time plus the delay
```
juliev0, line 322:
```suggestion
	// Only assess the child if the NextAssessmentTime is after the current time plus the delay
```

internal/controller/progressive/progressive.go
afugazzotto: good catch, thanks

juliev0 (approved):

PR #507, opened 2025-01-15
fix: handling in pipelinerollout reconciler of the "won't pause" case for pause and drain

### LINKED ISSUE #505: Pipeline in failed state "can't resume"
**Describe the bug**

This is related to when there's a Failed Pipeline. We've decided before that a Failed Pipeline does not require pausing. 

(Note that a Failed pipeline can occur when an isbsvc is upgrading, which is what we observed.)

I think there may be 2 issues which are in some ways canceling each other out:
1. The Failed Pipeline is supposed to be okay to resume, but there's a check to prevent resuming if isbsvc or numaflow controller is requesting pause, which there shouldn't be.
2. After we try and fail to resume, we leave the PPND "in progress strategy", which we probably shouldn't. Then we direct apply the change which removes the "desiredPhase" field. 

---

juliev0:
### The Problem

Fixes #505 

This fixes an issue seen when updating the ISBServiceRollout. After updating an isbsvc, the Pipeline is temporarily in a "Failed" state. 

Based on a previous agreement between myself and @whynowy , when the Pipeline is "Failed", we [don't try to pause the pipeline](https://github.com/numaproj/numaplane/blob/v0.12.1/internal/controller/pipelinerollout/pipelinerollout_ppnd.go#L146-L148). 

There are 2 problems in this case with the logic for when the pipeline needs to resume (thankfully, the net result was that the pipeline did resume, however):
1. Problem 1: After determining that pipeline "won't pause" because it's Failed, when we try to resume the Pipeline, we [can't because the ISBServiceRollout reconciler is still requesting the pause](https://github.com/numaproj/numaplane/blob/v0.12.1/internal/controller/ppnd/ppnd_pause.go#L120).  Basically, the code is unaware that we already we decided we "won't pause".
2. Problem 2: After getting this error, we actually exit the "PPND In Progress Strategy", even though we didn't effectively manage to resume the pipeline like we intended.
3. In a subsequent reconciliation, we see that the `desiredPhase: Paused` is a mismatch from the Rollout Pipeline spec, which has no `desiredPhase` field in it, so we do a DirectApply (equivalent to "Running"), which is why everything was still okay in the end on the Numaplane side.  

### Modifications

1. To address problem 1:
enable the "won't pause" case to [prevent doing the check](https://github.com/numaproj/numaplane/blob/wont-pause-fix/internal/controller/ppnd/ppnd_pause.go#L110) for whether numaflow controller or isbsvc is requesting pause prior to running the pipeline
2. To address problem 2: 
[return an error](https://github.com/numaproj/numaplane/blob/wont-pause-fix/internal/controller/ppnd/ppnd_pause.go#L117-L126) if this rare race condition actually does happen, which would keep the pipelinerollout in  "PPND In Progress Strategy", so we'd come back through the same logic to try resuming again in the next reconciliation. 

### Verification

Updated ISBServiceRollout jetstream version, and observed that the `desiredPhase` went to "Running" as soon as "wontPause" was true. 

Since the logic that was modified involves the "won't pause" case, which can be because either "pipeline failed" or "data loss annotation" is on the PipelineRollout, I thought I'd add an e2e test for the latter, and an [issue](https://github.com/numaproj/numaplane/issues/509) to add an e2e test which includes the former later.

This only affects the path taken when an isbsvc is updating, so I don't think there should be any issues.


internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -55,16 +55,23 @@ func (r *PipelineRolloutReconciler) processExistingPipelineWithPPND(ctx context.
...
-	if err := r.setPipelineLifecycle(ctx, shouldBePaused, pipelineRollout, existingPipelineDef); err != nil {
-		return false, err
+	if shouldBePaused {
+		if err := r.setPipelineLifecyclePaused(ctx, existingPipelineDef); err != nil {
+			return false, err
+		}
+	} else {
+		if err := r.setPipelineLifecycleRunning(ctx, pipelineRollout, existingPipelineDef, forceResume); err != nil {
+			return false, err
+		}
```
juliev0, line 74: `forceResume` is true if we decided we won't pause. This means we won't check "if it's safe to resume" when we call `r.setPipelineLifecycleRunning()`

internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -254,14 +277,7 @@ func (r *PipelineRolloutReconciler) setPipelineLifecycle(ctx context.Context, pa
...
-		if err != nil {
-			return err
-		}
-		if !run {
-			numaLogger.Infof("new pause request, can't resume pipeline at this time, will try again later")
-			r.recorder.Eventf(existingPipelineDef, "Normal", "PipelineResumeFailed", "new pause request, can't resume pipeline at this time, will try again later")
-		}
+		return ppnd.GetPauseModule().RunPipeline(ctx, r.client, existingPipelineDef, isbsvcRollout.Name, force)
 	}
 	return nil
```
juliev0, line 282: I separated this into 2 functions since it now has different parameters.

internal/controller/ppnd/ppnd_pause.go
```diff
@@ -103,28 +103,30 @@ func (pm *PauseModule) PausePipeline(ctx context.Context, c client.Client, pipel
...
+		}
+		if controllerPauseRequest != nil && *controllerPauseRequest {
+			// controller is requesting to pause - can't run
+			return fmt.Errorf("new pause request from NumaflowController, can't resume pipeline at this time; will try again later")
+		}
+
+		isbsvcPauseRequest := pm.PauseRequests[pm.GetISBServiceKey(pipeline.GetNamespace(), isbsvcName)]
+		if isbsvcPauseRequest != nil && *isbsvcPauseRequest {
+			// isbsvc is requesting to pause - can't run
+			return fmt.Errorf("new pause request from InterstepBufferService, can't resume pipeline at this time; will try again later")
```
juliev0, line 125: returning an error instead of a boolean; this will enable us to stay in the PPND "in progress strategy" so we can come back here and try again

tests/e2e/functional_test.go
```diff
@@ -249,7 +250,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 		document("Verifying that the Pipeline was created")
 		verifyPipelineSpec(Namespace, pipelineRolloutName, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
-			return len(pipelineSpec.Vertices) == numPipelineVertices // TODO: make less kludgey
+			return len(initialPipelineSpec.Vertices) == len(retrievedPipelineSpec.Vertices) // TODO: make less kludgey
```
juliev0, line 253: the previous check wasn't right - wasn't checking the `retrievedPipelineSpec`

tests/e2e/functional_test.go
```diff
@@ -372,6 +373,87 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 	time.Sleep(2 * time.Second)
 
+	// TODO: maybe move this into a ppnd-specific test suite?
+	// Can include tests for allowing data loss which involve all 3 pause requesters (isbsvc, numaflow controller, and pipeline itself)
```
juliev0, line 377: added [issue](https://github.com/numaproj/numaplane/issues/509) to move this to a PPND-specific test

afugazzotto (approved):

PR #513, opened 2025-01-21
fix: requeue if we still have children to delete

juliev0:
### Modifications

For Blue/Green, when we successful roll out the new Pipeline, we need to clean up the old one. We drain it first before we delete it. So, we need to do multiple reconciliations to do that. Previously, we were not requeueing for that, so we were catching it whenever the `SyncPeriod` caused us to do another reconciliation. It's better to proactively requeue if we need to.

ISBService is similar. We need to do multiple reconciliations because we need to make sure all Pipelines have been deleted first. 

For both cases we requeue if we're still in the middle of deleting.  

### Verification

Created a slow pipeline. Updated pipeline. Observed that it got deleted after it had successfully paused and drained.
Created a slow pipeline. Updated both pipeline and isbsvc. Observed both got deleted after the pipeline was successfully paused and drained.

no issues


afugazzotto (approved):

afugazzotto (approved):

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -13,12 +13,12 @@ spec:
...
+            image: quay.io/numaio/numaflow-go/source-simple-source:stable
         # transformer is an optional container to do any transformation to the incoming data before passing to the sink
         transformer:
           container:
             image: quay.io/numaio/numaflow-rs/source-transformer-now:stable
       sink:
         udsink:
           container:
-            image: quay.io/numaio/numaflow-java/simple-sink:stable
+            image: quay.io/numaio/numaflow-go/simple-sink:stable
```
juliev0, line 24: these images work on both amd64 and arm64 unlike the java ones

PR #515, opened 2025-01-22
fix: update monovertexrollout sample and e2e to use a monovertex with images that are multi-platform

juliev0:
### Modifications

The images that our sample MonoVertexRollout and our e2e test uses were not built to be multi-platform, so they result in Crash loop if run on arm64 architecture. (That may be fixed later.) Therefore, updating these to use images which are multi-platform.

### Verification

Doing `kubectl apply` for the sample MonoVertexRollout and seeing it run successfully. 
Running the e2e test locally and seeing the MonoVertexRollout running successfully both at the beginning and end of the test.

N/A

### DIFF
--- a/config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
+++ b/config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
@@ -14,11 +14,7 @@ spec:
         udsource:
           container:
             image: quay.io/numaio/numaflow-go/source-simple-source:stable
-        # transformer is an optional container to do any transformation to the incoming data before passing to the sink
-        transformer:
-          container:
-            image: quay.io/numaio/numaflow-rs/source-transformer-now:stable
       sink:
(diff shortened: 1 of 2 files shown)


config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -14,11 +14,7 @@ spec:
...
           container:
             image: quay.io/numaio/numaflow-go/source-simple-source:stable
-        # transformer is an optional container to do any transformation to the incoming data before passing to the sink
-        transformer:
-          container:
-            image: quay.io/numaio/numaflow-rs/source-transformer-now:stable
       sink:
         udsink:
           container:
-            image: quay.io/numaio/numaflow-go/simple-sink:stable
```
juliev0, line 24: neither of these images were built for multi-architecture

dpadhiar (approved):
LGTM

Issue #429, opened 2024-11-21
Capability for user namespace-specific Controller definitions


juliev0:
# Summary

If user needs to have a Controller definition that's specific to their namespace (say either their definition of 1.5.0 is different from the standard one for some reason, or their namespace includes a 1.5.0 definition but their is no cluster-wide definition), then they could.

Motivation: a request from the team doing performance testing is that they want to be able to deploy a Controller definition to test with (and iterate on) without waiting for the AddOn process to roll it out. If they were to be able to deploy that Controller definition through their Asset's Deployment repo, they could be more nimble. In this case, the Numaflow Controller definition would be made available for their namespace.

I think the flow should be that we first check for the namespace-specific configuration and use that if there is one. Otherwise, look for the cluster-wide one.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
Hey @RohanAshar - I think we can try to get to this this quarter

juliev0:
Hey @chandankumar4 - assigned this to you when you're done with the other items this sprint (if you have time).

PR #518, opened 2025-01-22
feat: Use controller definition configmap from user namespace

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/429

### Modifications

- Update `rolloutConfig` map for storing [namespace/version] as key with fullSpec as value
- While reading the data from `rolloutConfig` firstly check user specific namespace configmap if not found then use from "numaplane-system" namespace

### Verification

- Verified on local kubernetes cluster by creating configmap in user namesapce

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -46,6 +46,7 @@ type ConfigManager struct {
 }
 
 type NumaflowControllerDefinitionsManager struct {
+	// rolloutConfig is a map of controller version to its full spec where key is namespace/version
 	rolloutConfig map[string]string
 	lock          *sync.RWMutex
 }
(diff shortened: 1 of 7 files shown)


chandankumar4:
Unit test is pending to add!!

internal/util/kubernetes/config_watcher.go
```diff
@@ -115,9 +110,9 @@ func handleNumaflowControllerDefinitionsConfigMapEvent(ctx context.Context, conf
 
 		// controller config definition is immutable, so no need to update the existing config
```
juliev0, line 111: I'm kind of thinking we should revisit this idea that the definition needs to be immutable - what if we were to make this `if event.Type == watch.Added || event.Type == watch.Modified`? (as part of a separate PR would be fine)

It would be nice if the user could make a change to their Numaflow Controller definition if they need to. 

Do you know why we chose to make it immutable? I can see that if user is already running Numaflow 1.3.4, for example, and then we update the definition for 1.3.4, we aren't going to retroactively update their Numaflow Controller, so I'm not sure if that's the main reason. But that doesn't seem like a strong reason, because in that case even if the user decided to then delete the Numaflow Controller 1.3.4 definition and then re-add it, it still wouldn't solve that problem.

internal/controller/config/config.go
```diff
@@ -109,30 +111,50 @@ func (cm *ConfigManager) GetConfig() (GlobalConfig, error) {
...
+	manifest, manifestExists := definition[key]
+	if !manifestExists {
+		key = fmt.Sprintf("%s/%s", common.NumaplaneSystemNamespace, version)
+		manifest, manifestExists = definition[key]
+		if !manifestExists {
+			return "", fmt.Errorf("no controller definition found for namespace/version %s/%s", namespace, version)
+		}
+	}
+
+	return manifest, nil
```
juliev0, line 128: nice way of handling

juliev0 (approved):

internal/util/kubernetes/config_watcher.go
chandankumar4: Created an issue for it: https://github.com/numaproj/numaplane/issues/542

chandankumar4: I don't remember exactly but I believe it was the same reason as you said "If user is already running Numaflow 1.3.4, for example, and then we update the definition for 1.3.4, we aren't going to retroactively update their Numaflow Controller"

PR #521, opened 2025-01-25
feat: perform child healthiness assessment over a period of time

### LINKED ISSUE #520: Assess child resource healthiness at regular intervals after a delay and withing a range of time
# Summary

During a progressive upgrade, assess the child resource healthiness after a long delay (120s) and then reassess periodically (every 10s) for a certain period of time (60s).

# Use Cases

In case of bad image changes of a pipeline vertex.

---

afugazzotto:
Fixes #520 

### Modifications

Added a config parameter to describe the assessment schedule via 3 numerical values: a time delay when to start the assessment, for how long to perform assessments, and how often to assess within the assessment window.
Modified the assessment logic and requeuing logic to reflect the 3 values above based on assessment result. The assessment will stop in case of failure unless a new upgrade version of the rollout is being reconciled.

### Verification

Manual tests, E2e tests, and unit tests.

No issue.


internal/controller/progressive/progressive.go
```diff
@@ -288,9 +290,19 @@ func processUpgradingChild(
...
-) (bool, bool, error) {
+) (bool, bool, time.Duration, error) {
 	numaLogger := logger.FromContext(ctx)
 
+	globalConfig, err := config.GetConfigManagerInstance().GetConfig()
+	if err != nil {
+		return false, false, 0, fmt.Errorf("error getting the global config for assessment processing: %w", err)
+	}
+
+	delay, assessFor, assessEvery, err := globalConfig.GetChildStatusAssessmentSchedule()
```
juliev0, line 301: nit: what do you think about calling `assessFor` -> `assessmentPeriod` and `assessEvery` -> `assessmentInterval`?

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -229,21 +231,39 @@ func (status *Status) ClearUpgradeInProgress() {
...
-// It checks if the Status object is not nil, the UpgradingChildStatus is not nil,
-// the NextAssessmentTime is in the future, and the AssessmentResult is not successful.
-// If the AssessmentResult is successful and the NextAssessmentTime is in the future
-// we should not check again since the assessment has already been performed.
+// If the AssessmentResult has not been deemed failed already
+// and the current time is after NextAssessmentTime and before AssessUntil,
+// the assessment can be performed.
+// NOTE: the assessment can also be performed if the current time is past the
+// end time of the window but the previous assessment was a failure. This allows us to
+// fix a failed upgrade by using a new rollout version with the fix.
```
juliev0, line 261: we're not doing that anymore, right?

internal/controller/progressive/progressive.go
afugazzotto: Then I would also change `delay` -> `assessmentDelay`. What do you think?

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -229,21 +231,39 @@ func (status *Status) ClearUpgradeInProgress() {
...
 func (cs *ChildStatus) CanAssess() bool {
-	return cs != nil && cs.NextAssessmentTime.Time.After(time.Time{}) && time.Now().After(cs.NextAssessmentTime.Time) && cs.AssessmentResult != AssessmentResultSuccess
+	return cs != nil &&
+		cs.NextAssessmentTime.After(time.Time{}) && time.Now().After(cs.NextAssessmentTime.Time) &&
+		((cs.AssessUntil.After(time.Time{}) && time.Now().Before(cs.AssessUntil.Time) && cs.AssessmentResult != AssessmentResultFailure) ||
+			cs.AssessmentResult == AssessmentResultFailure)
```
juliev0, line 266: I'm confused about this last part of the if statement `cs.AssessmentResult == AssessmentResultFailure` - I don't think we should be checking for failure after `cs.AssessUntil.Time`. Didn't we agree on this plan?:
"After 120 seconds, we start to check. As soon as we reach "not unknown", we check for an additional 60 seconds."

internal/controller/progressive/progressive.go
juliev0: yes, thanks!

pkg/apis/numaplane/v1alpha1/status_types.go
afugazzotto: That's correct. That last part (`cs.AssessmentResult == AssessmentResultFailure`) is there to stop assessing in case of failures occurring after the 120 seconds and before the 60 seconds have passed (after the initial assessment).

juliev0: I can see this is related to your comment above:
```
// NOTE: the assessment can also be performed if the current time is past the
// end time of the window but the previous assessment was a failure. This allows us to
// fix a failed upgrade by using a new rollout version with the fix.
```

When we have a new rollout version with the fix, that should clear the ChildStatus altogether.

juliev0: > That's correct. That last part (`cs.AssessmentResult == AssessmentResultFailure`) is there to stop assessing in case of failures occurring after the 120 seconds and before the 60 seconds have passed (after the initial assessment).

Am I misreading? It looks like that last part doesn't take into account `AssessUntil` at all

afugazzotto: @juliev0 updated the logic as per discussion

internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -292,7 +292,7 @@ func (r *MonoVertexRolloutReconciler) processExistingMonoVertex(ctx context.Cont
 		}
 	}
 
-	requeue := false
+	requeueDelay := time.Duration(0)
```
juliev0, line 295:
```suggestion
	requeueDelay := time.Duration(0)   // 0 means "no requeue"
```

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -276,7 +276,7 @@ func (r *ISBServiceRolloutReconciler) GetChildTypeString() string {
 // - true if needs a requeue
```
juliev0, line 276: fix comment related to parameter type

internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -255,7 +255,7 @@ func (r *MonoVertexRolloutReconciler) reconcile(ctx context.Context, monoVertexR
 
 // return whether we should requeue, and return error if any (if returning an error, we will requeue anyway)
```
juliev0, line 256: fix comment related to parameter type

internal/controller/progressive/progressive.go
```diff
@@ -299,41 +311,40 @@ func processUpgradingChild(
...
-	numaLogger.WithValues("childStatus", *childStatus).Debug("performing upgrading child assessment")
+		numaLogger.WithValues("name", existingUpgradingChildDef.GetName(), "childStatus", *childStatus).Debugf("performing upgrading child assessment, assessment returned: %v", assessment)
+	} else {
+		numaLogger.WithValues("name", existingUpgradingChildDef.GetName(), "childStatus", *childStatus).Debug("skipping upgrading child assessment but assessing previous child status")
+	}
 
-	assessment, err := controller.AssessUpgradingChild(ctx, existingUpgradingChildDef)
-	if err != nil {
-		return false, false, err
+	// Once a "not unknown" assessment is reached, set the assessments end time (if not set yet)
```
juliev0, line 342:
```suggestion
	// Once a "not unknown" assessment is reached, set the assessment's end time (if not set yet)
```

juliev0 (approved):

PR #523, opened 2025-01-26
fix: if Progressive is set, do Apply strategy for Numaflow Controller updates 

juliev0:
### Modifications

The problem: When updating NumaflowControllerRollout, if Progressive strategy is set as preferred, we are doing Apply strategy, but we are still setting the `UpgradeInProgress` field to "Progressive" in the `NumaflowControllerRollout` and we are leaving it set there. This changes it so we don't set that at all.

### Verification

Updated NumaflowControllerRollout. This applied the Numaflow Controller and did not set the `UpgradeInProgress` field.

none


juliev0:
@afugazzotto can you review this one?

afugazzotto (approved):

PR #530, opened 2025-01-28
fix: slow-to-pause pipeline in e2e test needs "pauseGracePeriodSeconds" configured

### LINKED ISSUE #526: flakey e2e test - "Verifying that Pipeline keeps trying to pause"
**Describe the bug**
e2e test which tests the "allow-data-loss" annotation didn't keep trying to pause. 
Link: https://github.com/numaproj/numaplane/actions/runs/13004082300/job/36267856544?pr=525

---

juliev0:
Fixes #526 

### Modifications

There is an e2e test which tries to pause a pipeline which is very slow to pause. However, I realized that since I wasn't setting `pauseGracePeriodSeconds` in the Pipeline, it was by default getting set to "30 seconds", which meant that it would automatically stop pausing after 30 seconds by Numaflow Controller anyway. Reset to 3600 seconds.

### Verification

e2e

N/A


tests/e2e/functional_test.go
```diff
@@ -378,8 +378,10 @@ var _ = Describe("Functional e2e", Serial, func() {
 
 			document("Creating a slow pipeline")
 			slowPipelineSpec := updatedPipelineSpec.DeepCopy()
-			highRPU := int64(10000)
+			highRPU := int64(10000000)
```
juliev0, line 381: at first I was trying to just increase this before I realized the real issue was the pauseGracePeriodSeconds. But I just left this value here since it doesn't harm anything.

dpadhiar (approved):
LGTM!

Issue #528, opened 2025-01-28
Groundwork to enable creation of more e2e test suites.


juliev0:
# Summary

First, let's consolidate our suite_test.go and functional_test.go into a single file. In the future, when we add a new test suite, it will just be a single file. 

Second, create reusable e2e functions. Once we have reusable e2e functions, we can start to build new e2e tests easily which use them. 

Here are some ideas:
1. Create a NumaflowControllerRollout of a given version and make sure it's running
2. Create an ISBServiceRollout of a given version/name and make sure it's running
3. Create a PipelineRollout of a given spec/name and make sure it's running
4. Create a MonoVertexRollout of a given spec/name and make sure it's running
5. Delete any of those and verify it's deleted (one function each)
6. Perform an update to any of those and verify that it did the right thing (maybe pass in a variable for the strategy being used)

After this, we can try it out by creating [this new test](https://github.com/numaproj/numaplane/issues/509).

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #533, opened 2025-01-29
feat: consolidate suite_test file into common and functional_test files

dpadhiar:
### Modifications

This is the first step for the issue #528 to allow for the creation of more test suites.

Consolidates the `suite_test.go` file into the `functional_test.go` and `common.go` files which will allow us to create more test suites that use the same before and after actions. This will also allow to work on the next step of creating reusable functions to be used across multiple test suites.

### Verification

Running `make test-e2e` locally and passing on CI.

None.


tests/e2e/common.go
```diff
@@ -397,3 +402,163 @@ func getUpgradeStrategy() config.USDEUserStrategy {
...
+	if disableTestArtifacts != "true" {
+		setupOutputDir()
+	}
+
+	openFiles = make(map[string]*os.File)
+
+	stopCh = make(chan struct{})
+
+	upgradeStrategy = getUpgradeStrategy()
+	Expect(upgradeStrategy.IsValid()).To(BeTrue())
```
juliev0, line 423: I guess we should consider if every test requires this environment variable to be specified. Some tests will only be for one strategy, so it will be nice if user doesn't necessarily need to specify it.

juliev0: Perhaps we can just have a reusable function which each suite's `BeforeSuite()` calls which contains everything but that?

dpadhiar: After we have created the appropriate test suites for each strategy, I don't believe it will be necessary but for now it is.

tests/e2e/functional_test.go
```diff
@@ -185,6 +186,12 @@ var (
 	monovertexgvr schema.GroupVersionResource
 )
 
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "E2E Suite")
+}
```
juliev0, line 193: I wonder if we'll want to call this `TestFunctionalE2E()` or `FunctionalTestE2E()`? and `RunSpecs(t, "Functional E2E Suite")`?

juliev0 (approved):

juliev0:
Cool, we can see if any further adjustments need to be made once you create the next test

Issue #436, opened 2024-11-27
e2e test for Concurrent updates


juliev0:
# Summary

It would be great to have a test which tests concurrent requests for Numaflow Controller update, ISBService update, and Pipeline update, MonoVertex update, and maybe just make sure that everything is eventually updated and nothing left pausing.

Maybe this can be a test that is run for each strategy, just like the functional test.

We want to test the updating of each of them with "data loss" field, "recreate" field, "direct apply" field (whichever are [applicable](https://github.com/numaproj/numaplane/blob/main/config/manager/usde-config.yaml) to that Rollout).

How about these combinations:
- PipelineRollout data loss, ISBService data loss, Numaflow Controller data loss (version), MonoVertex image
- PipelineRollout data loss, ISBService recreate, Numaflow Controller data loss (version), MonoVertex image
- PipelineRollout data loss, ISBService direct apply, Numaflow Controller data loss (version), MonoVertex image
- PipelineRollout direct apply, ISBService recreate, Numaflow Controller data loss (version), MonoVertex image
- PipelineRollout direct apply, ISBService data loss, Numaflow Controller data loss (version), MonoVertex image

In our functional test, we sleep 2 seconds between tests. We could do the same here. 

What would need to be verified?:
Primary thing is that everything gets updated (spec matches desired spec) and everything is "Running". All Rollouts should be "Deployed" and none should have their "in progress strategy" set.
Each Rollout should only have one child at the end. 

I'm thinking we should probably do this in multiple PRs, since it's a lot. 

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #534, opened 2025-01-29
feat: Handle "delete/recreate" fields for isbsvc

### LINKED ISSUE #502: Perform delete/recreate for the "recreate" fields (isbsvc)
# Summary

Based on [this](https://github.com/numaproj/numaplane/issues/402) issue. 

Need to delete/recreate in accordance with the [Upgrade Strategy Matrix](https://docs.google.com/spreadsheets/d/1peDmaLS-xvPjFrO_zstUhBwR9cVm8t10W81FjhGJa9c/edit?pli=1&gid=0#gid=0) (with or without PPND).

---

juliev0:
Fixes #502 

Sorry, I know this PR is hard to review, as there's a lot that's changed. But note that I have done some fairly extensive testing, described in "Verification" section below.

### Modifications

This handles the updating of the "[recreate](https://github.com/numaproj/numaplane/blob/main/config/manager/usde-config.yaml#L70)" fields in the isbsvc.

If user-preferred strategy is "no strategy", it does the following:
- mark isbsvc "recyclable"
- pipelinerollout reconciler sees this and marks corresponding pipelines "recyclable"
- pipelinerollout reconciler deletes pipelines
- isbsvcrollout reconciler deletes isbsvc once all pipelines are deleted
- isbsvcrollout reconciler creates new isbsvc once the old one is deleted
- pipelinerollout reconciler creates new pipeline once new "promoted" isbsvc exists

If user-preferred strategy is "ppnd", the same logic occurs except that Pipeline is paused through the sequence shown above.

If user preferred strategy is "progressive", we do standard blue/green.

### Verification

I added a new e2e test to test the updating of a recreate field. The test verifies all of the usual data-loss related logic, plus that a new isbsvc and a new pipeline were created.

In addition, I did various manual tests for updating both pipeline and isbsvc at the same time using our 3 upgrade strategies (we will [add this to e2e](https://github.com/numaproj/numaplane/issues/436) in the future):
1. updating isbsvc recreate field at the same time as pipeline dataloss field
2. updating isbsvc dataloss field at the same time as pipeline dataloss field (just making sure I didn't break anything)

There shouldn't be any. A new label called ["upgrade-state-reason"](https://github.com/numaproj/numaplane/blob/delete-recreate-isbsvc/internal/common/common.go#L92) is being added to the children, but it's optional. (It's really only being used in the code for the "recyclable" upgrade-state anyway)


internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -229,28 +235,50 @@ func (r *MonoVertexRolloutReconciler) reconcile(ctx context.Context, monoVertexR
...
+			}
+			requeueDelay, err = r.processExistingMonoVertex(ctx, monoVertexRollout, existingMonoVertexDef, newMonoVertexDef, syncStartTime)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("error processing existing MonoVertex: %v", err)
+			}
+			if requeueDelay > 0 {
+				return ctrl.Result{RequeueAfter: requeueDelay}, nil
+			}
+			// process status
+			r.processMonoVertexStatus(ctx, existingMonoVertexDef, monoVertexRollout)
```
juliev0, line 258: I ended up deciding to update this code flow, just for symmetry with the other Rollouts, although technically it wasn't necessary for this one.

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -360,46 +360,100 @@ func (r *PipelineRolloutReconciler) reconcile(
...
+					return 0, nil, err
+				}
+				if needsPaused != nil && *needsPaused {
+					err = numaflowtypes.PipelineWithDesiredPhase(newPipelineDef, "Paused")
+					if err != nil {
+						return 0, nil, err
+					}
+				}
+
+			}
```
juliev0, line 404: Now it's possible to be in the middle of PPND up where we're creating the Pipeline (PPND user strategy with delete/recreate operation)

tests/e2e/common.go
```diff
@@ -369,7 +369,11 @@ func writeToFile(resource Output) error {
 
 func closeAllFiles() error {
 	for _, file := range openFiles {
-		err := file.Close()
+		err := file.Sync()
+		if err != nil { // flush file to disk
+			return err
+		}
```
juliev0, line 375: I'm actually not sure if it's not flushed or not when calling `file.Close()` but this is just in case

tests/e2e/functional_test.go
```diff
@@ -368,7 +384,7 @@ var _ = Describe("Functional e2e", Serial, func() {
 			}
 			return getRolloutConditionStatus(rollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused) == metav1.ConditionTrue &&
 				(retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePaused || retrievedPipelineStatus.Phase == numaflowv1.PipelinePhasePausing)
-		}, 15*time.Second, testPollingInterval).Should(BeTrue())
+		}, 10*time.Second, testPollingInterval).Should(BeTrue())
```
juliev0, line 387: reduced some of the time spent waiting in our `Consistently` tests, just to reduce the overall time of the e2e

internal/controller/common/rollout_controller.go
```diff
@@ -128,10 +145,12 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
 			}
 		}
 		// recycle the previous children
+		// TODO: do we want this? consider the cases that would actually use it
 		for _, recyclableChild := range recycleList {
 			numaLogger.Debugf("found multiple children of Rollout %s/%s of upgrade state=%q, marking recyclable: %s",
 				rolloutObject.GetRolloutObjectMeta().Namespace, rolloutObject.GetRolloutObjectMeta().Name, upgradeState, recyclableChild.GetName())
-			_ = UpdateUpgradeState(ctx, c, common.LabelValueUpgradeRecyclable, recyclableChild)
+			purgeOld := common.LabelValuePurgeOld
+			_ = UpdateUpgradeState(ctx, c, common.LabelValueUpgradeRecyclable, &purgeOld, recyclableChild)
```
afugazzotto, line 153: Why are you skipping error checking here?

internal/controller/common/rollout_controller.go
```diff
@@ -142,14 +161,47 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
 }
 
 // update the in-memory object with the new Label and patch the object in K8S
-func UpdateUpgradeState(ctx context.Context, c client.Client, upgradeState common.UpgradeState, childObject *unstructured.Unstructured) error {
+func UpdateUpgradeState(ctx context.Context, c client.Client, upgradeState common.UpgradeState, upgradeStateReason *common.UpgradeStateReason, childObject *unstructured.Unstructured) error {
+	numaLogger := logger.FromContext(ctx)
+
+	numaLogger.WithValues("upgradeState", upgradeState, "upgradeStateReason", util.OptionalString(upgradeStateReason)).Debug("patching upgradeState and upgradeStateReason to %s:%s%s",
```
afugazzotto, line 167:
```suggestion
	numaLogger.WithValues("upgradeState", upgradeState, "upgradeStateReason", util.OptionalString(upgradeStateReason)).Debug("patching upgradeState and upgradeStateReason to %s:%s/%s",
```

internal/controller/common/rollout_controller.go
```diff
@@ -142,14 +161,47 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
...
-	patchJson := `{"metadata":{"labels":{"` + common.LabelKeyUpgradeState + `":"` + string(upgradeState) + `"}}}`
 	return kubernetes.PatchResource(ctx, c, childObject, patchJson, k8stypes.MergePatchType)
 }
 
+func GetUpgradeState(ctx context.Context, c client.Client, childObject *unstructured.Unstructured) (*common.UpgradeState, *common.UpgradeStateReason) {
+	numaLogger := logger.FromContext(ctx)
+
+	upgradeStateStr, found := childObject.GetLabels()[common.LabelKeyUpgradeState]
+	if !found {
+		numaLogger.Debug("upgradeState unset for %s:%s%s", childObject.GetKind(), childObject.GetNamespace(), childObject.GetName())
```
afugazzotto, line 187:
```suggestion
		numaLogger.Debug("upgradeState unset for %s:%s/%s", childObject.GetKind(), childObject.GetNamespace(), childObject.GetName())
```

internal/controller/common/rollout_controller.go
```diff
@@ -142,14 +161,47 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
...
+		return nil, nil
+	} else {
+		reasonStr, found := childObject.GetLabels()[common.LabelKeyUpgradeStateReason]
+		numaLogger.WithValues("upgradeState", upgradeStateStr, "upgradeStateReason", util.OptionalString(reasonStr)).Debug("reading upgradeState and upgradeStateReason for %s:%s%s",
```
afugazzotto, line 191:
```suggestion
		numaLogger.WithValues("upgradeState", upgradeStateStr, "upgradeStateReason", util.OptionalString(reasonStr)).Debug("reading upgradeState and upgradeStateReason for %s:%s/%s",
```

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -224,45 +226,94 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error deleting recyclable interstepbufferservices: %s", err.Error())
+	}
+
+	// if we still have interstepbufferservices that need deleting, or if we're in the middle of an upgrade strategy, then requeue
+	if !allDeleted || inProgressStrategy != apiv1.UpgradeStrategyNoOp {
+		if requeueDelay == 0 {
+			requeueDelay = common.DefaultRequeueDelay
+		} else {
+			requeueDelay = min(requeueDelay, common.DefaultRequeueDelay)
```
afugazzotto, line 310: Is this getting the min between the 2 requeue values? If so, why?

internal/util/util.go
```diff
@@ -185,3 +185,16 @@ func ExtractPath(data any, path []string) (any, bool, error) {
 		return nil, false, fmt.Errorf("invalid type encountered: %s", v.Kind().String())
 	}
 }
+
+// take any pointer passed in and print "nil" if it's nil, and otherwise the value
+// (convenience function for logging)
+func OptionalString(ptr any) string {
```
afugazzotto, line 191: Nice and useful

tests/e2e/functional_test.go
afugazzotto: I agree, we may want to consider different timeouts and intervals based on what is being tested. The same values don't always make sense everywhere. We should revisit for all other cases (separate issue/PR).

internal/controller/common/rollout_controller.go
juliev0: good catch

internal/controller/isbservicerollout/isbservicerollout_controller.go
juliev0: I think if 2 pieces of code think we should requeue, then we should take the minimum amount of time and use that

internal/controller/common/rollout_controller.go
juliev0: You're right - we should at least log it. My original thought was not to error out just because we couldn't patch one of these old children as Recyclable if we somehow got into this state, since this operation doesn't seem essential. I'm inclined to just log it and not return the error unless you disagree. (Keep in mind we re-reconcile every 15 minutes anyway.)

tests/e2e/functional_test.go
juliev0: Sure.

internal/controller/common/rollout_controller.go
afugazzotto: I'm ok with that. Maybe a warning log.

juliev0: I actually added an "Error" level log even though I'm not returning an error. I'm kind of thinking "Error" is if there's a system error of some sort (including network) and "Warning" if there seems to be something wrong with user spec.

juliev0:
addressed your comments @afugazzotto

afugazzotto (approved):

PR #536, opened 2025-01-31
chore: introduce new RolloutController interface

juliev0:
Note: this PR makes no logic changes.

### Modifications

Previously, the whole idea of having an "upgrade-state" and also the idea of using that to garbage collect children was considered to be a "progressive" thing. Now that we are introducing the delete/recreate capability for child objects, we can use the same "recyclable" label value for performing that deletion and the same logic for performing the delete. 

This means we need to move the code related to updating that value and for updating the label altogether out of the progressive-specific code. 

Therefore, I am introducing a new interface called `RolloutController`, where all 4 of our Rollouts either implement them now or will in the future (`NumaflowControllerRollout` would be future for performing blue/green; the others already do). 

### Verification

no logic was changed, so standard CI is sufficient

N/A

### DIFF
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -407,7 +407,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 		return 0, fmt.Errorf("%v strategy not recognized", inProgressStrategy)
 	}
 	// clean up recyclable interstepbufferservices
-	allDeleted, err := progressive.GarbageCollectChildren(ctx, isbServiceRollout, r, r.client)
+	allDeleted, err := ctlrcommon.GarbageCollectChildren(ctx, isbServiceRollout, r, r.client)
 	if err != nil {
 		return 0, fmt.Errorf("error deleting recyclable interstepbufferservices: %s", err.Error())
 	}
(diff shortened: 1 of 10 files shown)


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -790,3 +791,62 @@ func (r *ISBServiceRolloutReconciler) ChildNeedsUpdating(ctx context.Context, fr
...
 }
+
+func (r *ISBServiceRolloutReconciler) getCurrentChildCount(rolloutObject ctlrcommon.RolloutObject) (int32, bool) {
+	isbsvcRollout := rolloutObject.(*apiv1.ISBServiceRollout)
+	if isbsvcRollout.Status.NameCount == nil {
+		return int32(0), false
+	} else {
+		return *isbsvcRollout.Status.NameCount, true
+	}
+}
```
juliev0, line 802: These functions moved from the `isbservicerollout_progressive.go` file (same for the other rollouts)

internal/controller/progressive/progressive_test.go
```diff
@@ -20,169 +18,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// This tests the FindMostCurrentChildOfUpgradeState() function
-// This function both finds the most current child of the upgrade-state, but also recycles any others that are found
-// So, it should only be called if there in fact should be only one child of that upgrade-state in existence
-func TestFindMostCurrentChildOfUpgradeState(t *testing.T) {
```
juliev0, line 26: moved to  rollout_controller_test.go

afugazzotto (approved):

PR #538, opened 2025-02-03
feat: create reusable e2e functions

### LINKED ISSUE #528: Groundwork to enable creation of more e2e test suites.

dpadhiar:
Fixes #528 

### Modifications

This PR largely breaks down the current functional test suite and moves the create/update/deletion functions for each respective rollout to their own file. These functions are now able to be used in different test suites we can build out for different upgrade strategies or any other cases that shouldn't require us to run the whole functional test.

We can further improve on this by introducing new shared functions but these should be sufficient to create new test suites now.

### Verification

Running `make test-e2e` multiple times locally with no strategy and PPND. Using CI for final verification.

None.

### DIFF
--- a/tests/e2e/isbservice.go
+++ b/tests/e2e/isbservice.go
@@ -322,7 +322,7 @@ func updateISBServiceRollout(isbServiceRolloutName, pipelineRolloutName string, 
 		return rollout, nil
 	})
 
-	if upgradeStrategy == config.PPNDStrategyID && dataLoss == true {
+	if upgradeStrategy == config.PPNDStrategyID && dataLoss {
 
 		document("Verify that in-progress-strategy gets set to PPND")
 		verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyPPND)
@@ -343,7 +343,7 @@ func updateISBServiceRollout(isbServiceRolloutName, pipelineRolloutName string, 
 
 	}
 
-	if upgradeStrategy == config.PPNDStrategyID && dataLoss == false {
+	if upgradeStrategy == config.PPNDStrategyID && !dataLoss {
 		document("Verify that dependent Pipeline is not paused when an update to ISBService not requiring pause is made")
 		verifyNotPausing := func() bool {
 			_, _, retrievedPipelineStatus, err := getPipelineSpecAndStatus(Namespace, pipelineRolloutName)
--- a/tests/e2e/isbservice.go
+++ b/tests/e2e/isbservice.go
@@ -2,13 +2,16 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


tests/e2e/isbservice.go
```diff
@@ -229,3 +232,145 @@ func startISBServiceRolloutWatches() {
...
+			return false
+		}
+		if len(list.Items) == 0 {
+			return true
+		}
+		return false
+	}).WithTimeout(testTimeout).Should(BeTrue(), "The ISBService should have been deleted but it was found.")
+}
+
+func updateISBServiceRollout(isbServiceRolloutName, pipelineRolloutName string, newSpec numaflowv1.InterStepBufferServiceSpec, f func(numaflowv1.InterStepBufferServiceSpec) bool, dataLoss bool) {
```
juliev0, line 315: nit: can you change the name `f` to something like `verifySpecFunc`?

juliev0: Actually, maybe it's fine. In that case, maybe just comment the function to explain what the parameters mean?

juliev0: Can you add a TODO to make `pipelineRolloutName` an array?

tests/e2e/isbservice.go
```diff
@@ -229,3 +232,145 @@ func startISBServiceRolloutWatches() {
...
+
+	rawSpec, err := json.Marshal(newSpec)
+	Expect(err).ShouldNot(HaveOccurred())
+
+	updateISBServiceRolloutInK8S(isbServiceRolloutName, func(rollout apiv1.ISBServiceRollout) (apiv1.ISBServiceRollout, error) {
+		rollout.Spec.InterStepBufferService.Spec.Raw = rawSpec
+		return rollout, nil
+	})
+
+	if upgradeStrategy == config.PPNDStrategyID && dataLoss {
```
juliev0, line 325: So, I guess `upgradeStrategy` is a global variable that the user will need to set for any test? That seems good I think.

tests/e2e/numaflowcontroller.go
```diff
@@ -117,3 +120,133 @@ func watchNumaflowControllerRollout() {
...
+			if !errors.IsNotFound(err) {
+				Fail("An unexpected error occurred when fetching the deployment: " + err.Error())
+			}
+			return false
+		}
+		return true
+	}).WithTimeout(testTimeout).Should(BeFalse(), "The deployment should have been deleted but it was found.")
+}
+
+func updateNumaflowControllerRollout(originalVersion, newVersion, pipelineRolloutName string, valid bool) {
```
juliev0, line 196: let's add a todo for this one as well regarding making it an array

tests/e2e/isbservice.go
dpadhiar: I believe it should work for any valid strategy or if it is empty since we're using the `IsValid()` function to verify that it is valid.

tests/e2e/numaflowcontroller.go
juliev0: also let's add a comment for parameters

tests/e2e/pipeline.go
```diff
@@ -333,3 +338,140 @@ func startPipelineRolloutWatches() {
...
+		}
+		if len(list.Items) == 0 {
+			return true
+		}
+		return false
+	}).WithTimeout(testTimeout).Should(BeTrue(), "The Pipeline should have been deleted but it was found.")
+}
+
+// update PipelineRollout and verify correct process
+func updatePipelineRollout(name string, newSpec numaflowv1.PipelineSpec, expectedPhase numaflowv1.PipelinePhase, dataLoss bool) {
```
juliev0, line 431: Can we either name `expectedPhase` -> `expectedFinalPhase` or clarify in a comment what it represents? I don't want someone to get confused thinking that if it's a PPND strategy, that represents the temporary "Pausing" state.

tests/e2e/isbservice.go
juliev0: But it looks like [IsValid(](https://github.com/numaproj/numaplane/blob/main/tests/e2e/functional_test.go#L194)) is only being called for the functional test. Perhaps we should call it from the common before suite logic?

dpadhiar: Sounds good, will move that over so that it should run for all suites.

juliev0 (approved):

PR #540, opened 2025-02-04
feat: scale down pods during progressive upgrade to allow for Kafka rebalancing (in case of Kafka sources)

### LINKED ISSUE #447: Ensure new Pipeline is receiving data
# Summary

This is to solve the following problems:

- eliminate the possibility that the new Pipeline is not receiving data due to too few Kafka partitions (no Kafka partitions available)
- only transfer data to the new Pipeline after it is showing its in a healthy state (phase=Running, healthy conditions for a period of time). (note [this ](https://github.com/numaproj/numaplane/issues/445)issue should be done first or in parallel with this one)

As recommended by Derek:

1. Update "promoted" pipeline such that its scale.max=half of the pods that it's currently running (or really the `floor(half(runningpods))` such that if 1 pod is running, then it would be 0. Given that users are not supposed to have a scale.max which exceeds the number of Kafka partitions and given that Kafka wants to keep things balanced, this should distribute many if not all partitions to the new pipeline/monovertex.  
2. If new Pipeline/mv fails assessment, then Kafka will redistribute the load back to the original one (may not happen immediately) since the new one will stop acking Kafka once there's enough backpressure due to the failure.

Future possible optimization: (I can write a separate issue for this)

The new pipeline could be created with source.scale.min and max=0 and only once we are confident that it's Running with healthy conditions, then we scale the source back up.

---

afugazzotto:
Fixes #447 

### Modifications

Scale down the source vertices pods of a pipeline/monovertex before proceeding with the progressive upgrade.

### Verification

Lots of manual testing, unit tests, E2E tests.

No issues.


internal/controller/progressive/progressive.go
```diff
@@ -68,12 +74,55 @@ func ProcessResource(
...
+		}
+		rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.Name = existingPromotedChild.GetName()
+		rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.ScaleValues = scaleValuesMap
+		rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.MarkAllSourceVerticesScaledDown()
+
+		return false, false, common.DefaultRequeueDelay, nil
+	}
+
+	// In case of failure and if not scaled back yet, scale the promoted rollout child back to its desired scale min and max values stored in the rollout progressive status
+	if liveRolloutObject.GetRolloutStatus().ProgressiveStatus.UpgradingChildStatus.IsFailed() &&
```
afugazzotto, line 112: In case of failure, we restore the desired min and max on the promoted pipeline. After setting to scale back up, should we continue to check for a period of time if the pods were actually restored as desired?

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -164,3 +167,213 @@ func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, ex
...
+			vertexName, found, err := unstructured.NestedString(vertexAsMap, "name")
+			if err != nil {
+				return err
+			}
+			if !found {
+				return errors.New("a vertex must have a name")
+			}
+
+			if _, exists := promotedChildStatus.ScaleValues[vertexName]; !exists {
+				return fmt.Errorf("the scale values for vertex '%s' are not present in the rollout promotedChildStatus", vertexName)
```
afugazzotto, line 358: Is it too drastic to stop the entire process because one vertex status scale information could be missing from the rollout status? If the rollout status is missing such information the process would be completely stuck and would require manual intervention.

internal/controller/progressive/progressive.go
```diff
@@ -46,6 +46,12 @@ type progressiveController interface {
 
 	// AssessUpgradingChild determines if upgrading child is determined to be healthy, unhealthy, or unknown
 	AssessUpgradingChild(ctx context.Context, existingUpgradingChildDef *unstructured.Unstructured) (apiv1.AssessmentResult, error)
+
+	// ScaleDownPromotedChildSourceVertices scales down by half the promoted child source vertices pods
+	ScaleDownPromotedChildSourceVertices(ctx context.Context, rolloutObject ctlrcommon.RolloutObject, promotedChildDef *unstructured.Unstructured, c client.Client) (map[string]apiv1.ScaleValues, bool, error)
+
+	// ScalePromotedChildSourceVerticesToDesiredValues scales back up to the desired scale values the promoted child source vertices pods
+	ScalePromotedChildSourceVerticesToDesiredValues(ctx context.Context, rolloutObject ctlrcommon.RolloutObject, promotedChildDef *unstructured.Unstructured, c client.Client) error
```
juliev0, line 54: We could probably include this in the first PR, but I'd love to not have these in the `progressiveController` interface, since it's not a hard requirement that scaling be part of the Rollouts, or that the progressiveController even have "vertices". I'll think of different options for how we might do that.

juliev0: One option:
Change these to be more generic:
`PreparePromotedChildForUpgrading()` and 
`PreparePromotedChildForUpgradingCompletion()`

We can still have your `scaleDownPromotedChild()` and `scaleUpPromotedChild()` functions, but just have them called from these interface functions for Pipeline and MV.

internal/controller/progressive/progressive.go
```diff
@@ -296,3 +329,124 @@ func IsNumaflowChildReady(upgradingObjectStatus *kubernetes.GenericStatus) bool
...
+		// ScaleDownPromotedChildSourceVertices either updates the existingPromotedChild to scale down the source vertices pods or
+		// retrieves the currently running pods to update the scaleValuesMap used on the rollout status.
+		// This serves to make sure that the pods for each vertex have been really scaled down before proceeding with the progressive update.
+		scaleValuesMap, promotedChildNeedsUpdate, err := controller.ScaleDownPromotedChildSourceVertices(ctx, liveRolloutObject, existingPromotedChild, c)
+		if err != nil {
+			return false, fmt.Errorf("error updating scaling properties to the existing promoted child definition: %w", err)
+		}
+
+		if promotedChildNeedsUpdate {
+			if err = kubernetes.UpdateResource(ctx, c, existingPromotedChild); err != nil {
```
juliev0, line 377: this can be a future PR, but we should consider making this a patch instead of an update. Then we won't have as much change of resource version conflict. We ended up changing the pause logic to do that FYI: https://github.com/numaproj/numaplane/blob/main/internal/controller/ppnd/ppnd_pause.go#L132-L136

internal/controller/progressive/progressive.go
```diff
@@ -296,3 +329,124 @@ func IsNumaflowChildReady(upgradingObjectStatus *kubernetes.GenericStatus) bool
...
+		// Set ScaleValuesRestoredToDesired to false in case previously set to true and now scaling back down to recover from a previous failure
+		rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.ScaleValuesRestoredToDesired = false
+
+		// Return that scaling down was performed
+		return true, nil
+	}
+
+	// Return that scaling down was NOT performed
+	return false, nil
+}
```
juliev0, line 397: I do like the elegance of having the `scaleDownPromotedChild()` and `scaleUpPromotedChild()` functions by the way

internal/controller/progressive/progressive.go
```diff
@@ -79,6 +86,15 @@ func ProcessResource(
...
 		if currentUpgradingChildDef == nil {
+			// Scale down the promoted child first
+			scaledDown, err := scaleDownPromotedChild(ctx, rolloutObject, liveRolloutObject, existingPromotedChild, skipPromotedChildSourceVerticesScaling, controller, c)
+			if err != nil {
+				return false, false, 0, err
+			}
+			if scaledDown {
+				return false, false, common.DefaultRequeueDelay, nil
+			}
+
```
juliev0, line 97: This can be an optimization for later, but I'm thinking it may be better to successfully create the new one before we scale down.

internal/controller/progressive/progressive.go
```diff
@@ -296,3 +329,121 @@ func IsNumaflowChildReady(upgradingObjectStatus *kubernetes.GenericStatus) bool
...
+	}
+
+	if err := kubernetes.UpdateResource(ctx, c, existingPromotedChild); err != nil {
+		return false, fmt.Errorf("error scaling back to desired min and max values the existing promoted child: %w", err)
+	}
+
+	rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.ScaleValuesRestoredToDesired = true
+
+	rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.AllSourceVerticesScaledDown = false
+	rolloutObject.GetRolloutStatus().ProgressiveStatus.PromotedChildStatus.ScaleValues = nil
```
juliev0, line 445: This seems nicely idempotent

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -117,9 +117,37 @@ type ChildStatus struct {
 	AssessUntil *metav1.Time `json:"assessUntil,omitempty"`
 }
 
+// ScaleValues stores the desired min and max, scaleTo, and actual scale values of a pipeline or monovertext vertex
```
juliev0, line 120:
```suggestion
// ScaleValues stores the desired min and max, scaleTo, and actual scale values of a pipeline or monovertex vertex
```

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -117,9 +117,37 @@ type ChildStatus struct {
...
+type ScaleValues struct {
+	// DesiredMin is the min scale value of the original child spec
+	DesiredMin int64 `json:"desiredMin"`
+	// DesiredMax is the max scale value of the original child spec
+	DesiredMax int64 `json:"desiredMax"`
+	// ScaleTo indicates how many pods to scale down to and it is calculated by getting the floor of the desiredMax value divided by 2
```
juliev0, line 126: nit: even though the comment is helpful, I may prefer that we remove the part about how it's calculated from here, since conceivably it could be a different implementation for each rollout using this

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -117,9 +117,37 @@ type ChildStatus struct {
...
+	// Name of the promoted child
+	Name string `json:"name"`
+	// ScaleValues is a map where the keys are the promoted child source vertices names
+	// and the values are the scale values of the source vertices
+	ScaleValues map[string]ScaleValues `json:"scaleValues,omitempty"`
+	// AllSourceVerticesScaledDown indicates if ALL the promoted child source vertices have been scaled down
+	AllSourceVerticesScaledDown bool `json:"allSourceVerticesScaledDown,omitempty"`
+	// ScaleValuesRestoredToDesired indicates if ALL the promoted child source vertices have been set back to the original min and max scale values (desiredMin and desiredMax).
+	// This field being set to `true` invalidates the value(s) in the scaleValues.Actual field.
+	ScaleValuesRestoredToDesired bool `json:"scaleValuesRestoredToDesired,omitempty"`
```
juliev0, line 143: If we are going to have all this scale stuff in our generic `PromotedChildStatus`, then let's at least have a struct to encapsulate it, and clearly comment that it only pertains to some Rollouts and not others

juliev0: I think we'll later need to explore the potential of not all Rollouts using the same `ProgressiveStatus` type. Maybe I can look at that.

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -81,3 +85,213 @@ func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, ex
...
+	rolloutObject ctlrcommon.RolloutObject,
+	promotedChildDef *unstructured.Unstructured,
+	c client.Client,
+) (map[string]apiv1.ScaleValues, bool, error) {
+
+	numaLogger := logger.FromContext(ctx).WithName("ScaleDownPromotedChildSourceVertices").WithName("PipelineRollout")
+
+	numaLogger.Debug("started promoted child source vertices scaling down process")
+
+	promotedChild, err := ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradePromoted, true, c)
```
juliev0, line 115: Was there a reason you needed to call this? Isn't it identical to `promotedChildDef`?

juliev0: Were you thinking that `promotedChildDef` represents the target Pipeline definition, as defined by the Rollout? I think it represents the existing one found in Kubernetes, doesn't it? Given that, I would think that the derivation of `originalMin` and `originalMax` down below would be based on the currently running Pipeline (and thus continually changing as you update it on each reconciliation) - am I making an error?

afugazzotto: It was a leftover from when I moved that logic. Fixed.

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -81,3 +85,213 @@ func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, ex
...
+			}
+
+			promotedChildNeedsUpdate = true
+
+			scaleValue := int64(math.Floor(float64(actualPodsCount) / float64(2)))
+
+			originalMax, _, err := unstructured.NestedInt64(vertexAsMap, "scale", "max")
+			if err != nil {
+				return nil, false, err
+			}
```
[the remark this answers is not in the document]
afugazzotto: As discussed, will address in separate PR.

afugazzotto:
As discussed, unresolved comments will be address in separate PR.

juliev0 (approved):
approving since remaining comments will be addressed in a future PR

Issue #510, opened 2025-01-20
PipelineRollout deletion does not wait for Pipeline deletion


whynowy:
**Describe the bug**

When deleting a `PipelineRollout`, the orchestrated Pipeline is stuck in deletion (e.g. numaflow controller is not running for whatever reason), PipelineRollout is cleaned up and a dangling Pipeline is left in deleting status.

Same thing happens for ISBServiceRollout.

**Expected behavior**

The PipelineRollout object should be in deleting status as well until Pipeline is deleted.

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

juliev0:
Thanks for noticing this and sorry I didn't see this issue until now (I guess I need to change my notification settings for new Issues). 

We will prioritize it

juliev0:
Hey @chandankumar4 - I guess our OwnerReference may need to be adjusted? Can you take a look sometime this sprint?

chandankumar4:
Sure, I'll take a look it on priority

juliev0:
> Sure, I'll take a look it on priority

thank you!

juliev0:
Okay, after discussing with @whynowy, I am okay if we want to have a different behavior than traditionally for background deletion. 

I did find another issue that would be resolved by this today. After the underlying PipelineRollout was deleted from ArgoCD (maybe through background deletion), I had no access to delete the Pipeline since we can't write to our namespace in kubernetes, and since the child Pipeline was hidden in ArgoCD. When I tried to sync the PipelineRollout to create a new Pipeline, I got the error "message: Pipeline numaplane-test-pipeline-0 already exists in namespace, not owned by a PipelineRollout". I guess the background deletion effectively left the Pipeline running but removed the PipelineRollout OwnerReference.

PR #548, opened 2025-02-06
fix: Set foreground deletion policy for rollout deletion

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/510

### Modifications
- Set foreground deletion policy for pipeline, monovertex, isbService and Numaflow Controller

### Verification

- Verified in local Kubernetes cluster

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -109,7 +109,7 @@ jobs:
   e2e-tests:
     name: E2E Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -109,7 +109,7 @@ jobs:
   e2e-tests:
     name: E2E Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 25
+    timeout-minutes: 35 # If we start getting error in CI as "The operation was canceled", we need to increase this value.
     strategy:
       fail-fast: false
       matrix:
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -202,7 +202,7 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		numaLogger.Info("Deleting ISBServiceRollout")
 		if controllerutil.ContainsFinalizer(isbServiceRollout, common.FinalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(isbsvcKey)
-			// Set the foreground deletion policy
+			// Set the foreground deletion policy so that we will block for children to be cleaned up for any type of deletion action
 			foreground := metav1.DeletePropagationForeground
 			if err := r.client.Delete(ctx, isbServiceRollout, &client.DeleteOptions{PropagationPolicy: &foreground}); err != nil {
 				return ctrl.Result{}, err
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -202,6 +202,11 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		numaLogger.Info("Deleting ISBServiceRollout")
 		if controllerutil.ContainsFinalizer(isbServiceRollout, common.FinalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(isbsvcKey)
+			// Set the foreground deletion policy
+			foreground := metav1.DeletePropagationForeground
+			if err := r.client.Delete(ctx, isbServiceRollout, &client.DeleteOptions{PropagationPolicy: &foreground}); err != nil {
+				return ctrl.Result{}, err
+			}
 			controllerutil.RemoveFinalizer(isbServiceRollout, common.FinalizerName)
 		}
 		// generate metrics for ISB Service deletion.
--- a/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
+++ b/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
@@ -231,9 +231,8 @@ func (r *NumaflowControllerReconciler) reconcile(
 		r.recorder.Eventf(controller, corev1.EventTypeNormal, "Deleting", "Deleting NumaflowController")
 		if controllerutil.ContainsFinalizer(controller, common.FinalizerName) {
 			// Check if dependent resources are deleted, if not then requeue
-			if !r.areDependentResourcesDeleted(ctx, controller) {
-				numaLogger.Warn("Dependent resources are not deleted yet, requeue...")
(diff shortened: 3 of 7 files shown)


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -204,6 +204,11 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		numaLogger.Info("Deleting ISBServiceRollout")
 		if controllerutil.ContainsFinalizer(isbServiceRollout, common.FinalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(isbsvcKey)
+			// Set the foreground deletion policy
+			foreground := metav1.DeletePropagationForeground
+			if err := r.client.Delete(ctx, isbServiceRollout, &client.DeleteOptions{PropagationPolicy: &foreground}); err != nil {
+				return ctrl.Result{}, err
+			}
```
juliev0, line 211: This is really interesting. So, from [this](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/) article, it looks like doing this essentially creates a finalizer. 

```
"finalizers": [
      "foregroundDeletion"
    ]
```

And then the GarbageCollector Controller is actually in charge of removing that finalizer. 

Good find!

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -204,6 +204,11 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		numaLogger.Info("Deleting ISBServiceRollout")
 		if controllerutil.ContainsFinalizer(isbServiceRollout, common.FinalizerName) {
 			ppnd.GetPauseModule().DeletePauseRequest(isbsvcKey)
+			// Set the foreground deletion policy
```
juliev0, line 207:
```suggestion
			// Set the foreground deletion policy so that we will block for children to be cleaned up for any type of deletion action
```

juliev0: can you add a similar comment for all?

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -723,20 +723,29 @@ func (r *NumaflowControllerReconciler) ErrorHandler(numaflowController *apiv1.Nu
 
 // areDependentResourcesDeleted checks if dependent resources are deleted.
 func (r *NumaflowControllerReconciler) areDependentResourcesDeleted(ctx context.Context, controller *apiv1.NumaflowController) bool {
-	pipelineList, err := kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion, "pipelines",
+	pipelineList, err := kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion, common.NumaflowPipelineKind,
 		controller.Namespace, common.LabelKeyParentRollout, "")
 	if err != nil {
-		return false
+		if !apierrors.IsNotFound(err) {
+			return false
```
juliev0, line 730: so, this is the case of getting a real error when listing? Should we maybe return an error to `Reconcile()` if that happens so that we'll re-reconcile?

juliev0: Anyway, I'm trying to understand how critical the bug is that you're fixing here. It seems like previously in the case of `apierrors.IsNotFound()`, then we would return false. In that case, we must pretty much never be deleting the `NumaflowController` CR - is that right?

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -234,9 +234,9 @@ func (r *NumaflowControllerReconciler) reconcile(
 		numaLogger.Info("Deleting NumaflowController")
 		r.recorder.Eventf(controller, corev1.EventTypeNormal, "Deleting", "Deleting NumaflowController")
 		if controllerutil.ContainsFinalizer(controller, common.FinalizerName) {
-			// Check if dependent resources are deleted, if not then requeue after 5 seconds
+			// Check if dependent resources are deleted, if not then requeue
 			if !r.areDependentResourcesDeleted(ctx, controller) {
-				numaLogger.Warn("Dependent resources are not deleted yet, requeue after 5 seconds")
+				numaLogger.Warn("Dependent resources are not deleted yet, requeue...")
 				return ctrl.Result{Requeue: true}, nil
```
juliev0, line 240: So, I guess we aren't doing foreground deletion for this one - I guess that's okay since it seems like theoretically there shouldn't be anything holding up the deletion of those underlying children, but then again I wonder if we just want it to be consistent. Any reason you can think of not to just add it here too?

internal/controller/isbservicerollout/isbservicerollout_controller.go
chandankumar4: Yes

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
chandankumar4: 1. Yes, error can be used as well to reconcile.
2. Yes this was the bug that Numaflow controller CR was never get's deleted, I'm not sure why it didn't cause the problem till now

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
chandankumar4: I didn't added it here, because `NumaflowController_rollout` already has it and in `NumaflowController_controller` has this `areDependentResourcesDeleted` check

juliev0: right, so I guess the one reason to add it here would be if say there were some resource being deployed by the NumaflowController - I know right now it's just `ConfigMap`, `Deployment`, etc - but what if one of them had some finalizer that couldn't be deleted for some reason? At least, we'd be consistent everywhere in Numaplane with that behavior. I'm kind of leaning towards adding it, unless you disagree.

chandankumar4: Sure, I'll add it, I don't have strong reason against it, also it will make it consistent.

.github/workflows/ci.yaml
```diff
@@ -109,7 +109,7 @@ jobs:
   e2e-tests:
     name: E2E Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 25
```
chandankumar4, line 112: Had to increase it as tests were getting cancelled while running, [example reference](https://github.com/numaproj/numaplane/actions/runs/13235133722/job/36940124528)

juliev0: good job figuring out the cause for that. If we're anywhere close to that number, how about we increase it even more then? maybe 35? 

I guess separately there is also [this](https://github.com/numaproj/numaplane/blob/main/tests/e2e/common.go#L48) other value in the code itself. Do you mind making both of them 35?

juliev0: I wonder if we should add a comment here as well to help us diagnose the problem if we see this same CI failure again?

juliev0 (approved):
approved but please add the comments mentioned, thanks

PR #552, opened 2025-02-07
feat: introduce ppnd e2e test suite

### LINKED ISSUE #509: e2e test specific to PPND
# Summary

For any cases that only apply to PPND, let's have a separate test. 

This includes:

- "allow data loss" annotation stops pausing: move this from our functional_test.go, and augment it such that numaflow controller, isbsvc, and pipeline are all requesting pause and requiring an update
- "won't pause" case - pipeline is failed - numaflow controller, isbsvc, and pipeline all need updating. Pipeline doesn't need to be paused.

---

dpadhiar:
Fixes #509 

### Modifications

Introduces the pause and drain e2e test suite. This is accomplished by redefining certain e2e test functions to be public within the e2e package and repurposing them to work for multiple test cases including these here.

The PPND test suite runs specific tests for just the pause and drain strategy which includes completing updates that cause data loss even if the Pipeline is failed and adding an annotation to allow data loss for a slower pausing Pipeline that is holding up updates.

### Verification

Running locally and on the CI.

None.


dpadhiar:
@afugazzotto @juliev0 The PR might be a bit messier than I'd like currently as merging main into this had quite a few conflicts so will try to clean it up as we review it.

.github/workflows/ci.yaml
```diff
@@ -178,4 +178,78 @@ jobs:
           path: |
             tests/e2e/output/logs/
           retention-days: 7
+
+
+  ppnd-e2e-test:
```
juliev0, line 183: maybe we can do it as a follow up PR, but we don't want to copy and paste all of these steps. I see looking at the other jobs that we already had an issue there. There seems to be something called "Composite Actions" in Github Actions that maybe we can try.

juliev0: We may also want to do what Numaflow is doing, in which `case` is a matrix: https://github.com/numaproj/numaflow/blob/main/.github/workflows/ci.yaml#L205. Maybe we could have one matrixed job called "MultiStrategyE2E" and another called "SingleStrategyE2E" or something. The first would include our "functional-test" and any other future multi-strategy test. The second would include the ppnd test and any other single-strategy test.

Makefile
```diff
@@ -128,7 +128,13 @@ test: codegen fmt vet envtest ## Run tests.
...
 .PHONY: test-e2e
 test-e2e: codegen fmt vet envtest ## Run e2e tests.
-	GOFLAGS="-count=1" go test -timeout 20m -v ./tests/e2e/... 
+	GOFLAGS="-count=1" go test -timeout 20m -v ./tests/e2e/...
+
+test-functional-e2e:
+test-ppnd-e2e:
+test-%: codegen fmt vet envtest
+	GOFLAGS="-count=1" go test -timeout 20m -v ./tests/e2e/$*
+
```
juliev0, line 137: this looks good

tests/e2e/isbservice.go
```diff
@@ -347,36 +348,39 @@ func updateISBServiceRollout(
...
+	if UpgradeStrategy == config.PPNDStrategyID {
 		if dataLossRisk {
 
-			document("Verify that in-progress-strategy gets set to PPND")
+			Document("Verify that in-progress-strategy gets set to PPND")
 			verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyPPND)
-			verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyPPND)
-			verifyPipelinePaused(Namespace, pipelineRolloutName)
+			// if we expect the pipeline to be failed?
+			// short term workaround
```
juliev0, line 364: maybe we should just pass in a boolean to the function called `pipelineIsFailed`?

tests/e2e/isbservice.go
```diff
@@ -347,36 +348,39 @@ func updateISBServiceRollout(
...
 
-			document("Verify that the pipelines are unpaused by checking the PPND conditions on ISBService Rollout and PipelineRollout")
+			Document("Verify that the pipelines are unpaused by checking the PPND conditions on ISBService Rollout and PipelineRollout")
 			Eventually(func() bool {
 				isbRollout, _ := isbServiceRolloutClient.Get(ctx, isbServiceRolloutName, metav1.GetOptions{})
 				isbCondStatus := getRolloutConditionStatus(isbRollout.Status.Conditions, apiv1.ConditionPausingPipelines)
 				plRollout, _ := pipelineRolloutClient.Get(ctx, pipelineRolloutName, metav1.GetOptions{})
 				plCondStatus := getRolloutConditionStatus(plRollout.Status.Conditions, apiv1.ConditionPipelinePausingOrPaused)
-				if isbCondStatus != metav1.ConditionTrue || plCondStatus != metav1.ConditionTrue {
+				if isbCondStatus == metav1.ConditionTrue || plCondStatus == metav1.ConditionTrue {
```
juliev0, line 376: glad you caught that one. :) I guess it only probably worked before because at the beginning of the test it was paused.

tests/e2e/numaflowcontroller.go
```diff
@@ -198,32 +198,34 @@ func deleteNumaflowControllerRollout() {
...
 
 	// NOTE: we are only checking the "valid" case because in the "non-valid" case the pipeline pausing conditions on
 	// the NumaflowController and Pipeline rollouts change too rapidly making the test flaky (intermittently pass or fail)
-	if upgradeStrategy == config.PPNDStrategyID && valid {
+	if UpgradeStrategy == config.PPNDStrategyID && valid {
 
-		document("Verify that in-progress-strategy gets set to PPND")
-		verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyPPND)
-		verifyPipelinePaused(Namespace, pipelineRolloutName)
+		if !strings.Contains(pipelineRolloutName, "failed") {
```
juliev0, line 216: same comment from before

tests/e2e/pipeline.go
```diff
@@ -138,8 +138,26 @@ func verifyPipelinePaused(namespace string, pipelineRolloutName string) {
 	//verifyPodsRunning(namespace, 0, getVertexLabelSelector(pipelineName))
 }
 
-func verifyInProgressStrategy(pipelineRolloutName string, inProgressStrategy apiv1.UpgradeStrategy) {
-	document("Verifying InProgressStrategy")
+func VerifyPipelineFailed(namespace, pipelineRolloutName string) {
+	Document("Verify that Pipeline Rollout condition is Failed")
```
juliev0, line 142:
```suggestion
	Document("Verify that Pipeline is Failed")
```

tests/e2e/pipeline.go
```diff
@@ -441,40 +461,42 @@ func deletePipelineRollout(name string) {
...
 	Expect(err).ShouldNot(HaveOccurred())
 
 	// update the PipelineRollout
-	updatePipelineRolloutInK8S(Namespace, name, func(rollout apiv1.PipelineRollout) (apiv1.PipelineRollout, error) {
+	UpdatePipelineRolloutInK8S(Namespace, name, func(rollout apiv1.PipelineRollout) (apiv1.PipelineRollout, error) {
 		rollout.Spec.Pipeline.Spec.Raw = rawSpec
 		return rollout, nil
 	})
 
 	// no data loss OR system pause
```
juliev0, line 476: why does it say "no data loss OR system pause"?

tests/e2e/pipeline.go
```diff
@@ -441,40 +461,42 @@ func deletePipelineRollout(name string) {
...
-	if upgradeStrategy == config.PPNDStrategyID && dataLoss {
-
-		document("Verify that in-progress-strategy gets set to PPND")
-		verifyInProgressStrategy(name, apiv1.UpgradeStrategyPPND)
-
-		verifyPipelinePaused(Namespace, name)
+	if UpgradeStrategy == config.PPNDStrategyID && dataLoss {
+
+		Document("Verify that in-progress-strategy gets set to PPND")
```
juliev0, line 479: this no longer pertains to all below

tests/e2e/pipeline.go
```diff
@@ -488,21 +510,73 @@ func updatePipelineRollout(name string, newSpec numaflowv1.PipelineSpec, expecte
...
 	}
 
+	Document("Verifying Pipeline got updated")
 	// get Pipeline to check that spec has been updated to correct spec
-	verifyPipelineSpec(Namespace, name, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
-		return len(retrievedPipelineSpec.Vertices) == numPipelineVertices
-	})
-
-	verifyPipelineRolloutDeployed(name)
+	VerifyPipelineSpec(Namespace, name, verifySpecFunc)
```
juliev0, line 515: hmm, this one shouldn't pass if we're still in a long pause - is it?

tests/e2e/pipeline.go
```diff
@@ -488,21 +510,73 @@ func updatePipelineRollout(name string, newSpec numaflowv1.PipelineSpec, expecte
...
+	}, testTimeout, testPollingInterval).Should(Succeed())
+
+	Document("Verifying that the Pipeline was created")
+	VerifyPipelineSpec(namespace, name, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
+		return len(spec.Vertices) == len(retrievedPipelineSpec.Vertices) // TODO: make less kludgey
+		//return reflect.DeepEqual(pipelineSpec, retrievedPipelineSpec) // this may have had some false negatives due to "lifecycle" field maybe, or null values in one
+	})
+
+	VerifyPipelineRolloutDeployed(name)
+	VerifyInProgressStrategy(name, apiv1.UpgradeStrategyNoOp)
```
juliev0, line 580: How is this function different from `CreatePipelineRollout()`?

tests/e2e/ppnd-e2e/ppnd_test.go
```diff
@@ -0,0 +1,367 @@
...
+	Document("Verifying that the slow pipeline was created")
+	VerifyPipelineSpec(Namespace, slowPipelineRolloutName, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
+		return len(slowPipelineSpec.Vertices) == len(retrievedPipelineSpec.Vertices)
+	})
+
+	VerifyPipelineRunning(Namespace, slowPipelineRolloutName, false)
+	VerifyInProgressStrategy(slowPipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+
+}
+func completeSlowPipelineRolloutTest() {
```
juliev0, line 339: I know you're trying to consolidate as much shared functionality as possible into a single function, but I still think functions should be named to have some kind of meaning, which represents what it does. 

I think it would be slightly less consolidated, but better to divide it into 3, e.g.:
1. verifySlowPipelinePausing()
2. allowDataLoss()
3. DeletePipelineRollout()

tests/e2e/ppnd-e2e/ppnd_test.go
```diff
@@ -0,0 +1,367 @@
...
+		failedPipelineSpec.Edges = append(failedPipelineSpec.Edges, numaflowv1.Edge{From: "not", To: "valid"})
+
+		CreateFailedPipelineRollout(failedPipelineRolloutName, Namespace, failedPipelineSpec)
+		VerifyPipelineFailed(Namespace, failedPipelineRolloutName)
+
+		time.Sleep(5 * time.Second)
+
+		// update spec to have topology change
+		failedPipelineSpec = updatedPipelineSpec
+		failedPipelineSpec.Edges = append(failedPipelineSpec.Edges, numaflowv1.Edge{From: "not", To: "valid"})
```
juliev0, line 250: so, the part that matters for this test is actually the fact that it was failed beforehand - not that the spec afterward was failed. Before the before spec was failed, we did not try to pause and were able to perform the update.

We should be able to simply update to `initialPipelineSpec` even, or otherwise `updatedPipelineSpec`.

Okay to do in a subsequent PR if you prefer.

tests/e2e/isbservice.go
dpadhiar: That would probably be best, just didn't want to continually increase the function signature but this would be more helpful.

tests/e2e/pipeline.go
dpadhiar: `dataLoss` is a bit of a confusing param, it informs us if the update will cause data loss or not. If I remember correctly, I added `system pause` cause I was debugging this function when I was testing the manual pause case. In any case, this is a leftover comment we can remove for now.

tests/e2e/pipeline.go
dpadhiar: It is not meant to pass but I am just passing a function that returns `true` to circumvent that. Should we handle this another way?

tests/e2e/pipeline.go
dpadhiar: The only difference is we don't check if the Pipeline is running and the vertices are up.

dpadhiar: We could probably safely add another `failed` boolean to this to handle that case and be sure we don't have multiple repeating functions.

tests/e2e/ppnd-e2e/ppnd_test.go
dpadhiar: Sure! Yeah that would be a better idea, I was thinking more about consolidating the logic overall.

tests/e2e/pipeline.go
dpadhiar: Added `failed` boolean.

tests/e2e/pipeline.go
juliev0: This one shouldn't be resolved. To clarify, for the "Failed" case below, we don't set the in-progress-strategy to PPND.

juliev0: Can you see if the individual functions below have `Document` statements in them? That may be sufficient then so maybe we don't need one up at the top.

tests/e2e/pipeline.go
juliev0: Oh, that's clever. But I do think it's a little hard to read and the comment makes it sound like it's "been updated to correct spec". I think it will be more apparent if we check for the expectedPhase being PipelinePhasePausing to determine whether this needs to be done. Also, maybe the function passed in should be a pointer since it should be optional if it's PipelinePhasePausing.

tests/e2e/pipeline.go
dpadhiar: Ah I resolved since I had changed it on my working branch (using the PR as a checklist essentially) but that would probably be better to do after updating on Github. Regardless, let me check about the individual funcs and we can change that.

tests/e2e/pipeline.go
juliev0: Why is this resolved?

tests/e2e/pipeline.go
juliev0: sounds good

tests/e2e/ppnd-e2e/ppnd_test.go
dpadhiar: Looking at this again.. I see the issue and will just update to updatedPipelineSpec without the bad edge. I assumed we wanted to keep the Pipeline in a failed state.

.github/workflows/ci.yaml
```diff
@@ -114,6 +114,10 @@ jobs:
       fail-fast: false
       matrix:
         strategy: ["pause-and-drain", "progressive", "no-strategy"]
+        case: ["functional"]
+        include:
+        - strategy: "pause-and-drain"
+          case: "ppnd"
```
juliev0, line 120: oh, that's cool that you were able to consolidate all of that together.

dpadhiar: Thank you! After taking a look at Numaflow's CI, I had forgot that you could control the matrix with the `include` field. The reason I had split them into 2 previously is the matrix would have ran our `ppnd` test with each strategy which wouldn't work anyway (and we don't need to do that). This way, we can control specific scenarios much easier.

tests/e2e/isbservice.go
```diff
@@ -347,36 +350,38 @@ func updateISBServiceRollout(
...
-	if upgradeStrategy == config.PPNDStrategyID {
+	if UpgradeStrategy == config.PPNDStrategyID {
 		if dataLossRisk {
 
-			document("Verify that in-progress-strategy gets set to PPND")
+			Document("Verify that in-progress-strategy gets set to PPND")
 			verifyInProgressStrategyISBService(Namespace, isbServiceRolloutName, apiv1.UpgradeStrategyPPND)
-			verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyPPND)
-			verifyPipelinePaused(Namespace, pipelineRolloutName)
+			// if we expect the pipeline to be failed
```
juliev0, line 365:
```suggestion
			// if we expect the pipeline to be healthy
```

tests/e2e/isbservice.go
```diff
@@ -413,30 +418,35 @@ func updateISBServiceRollout(
...
-	verifyISBServiceSpec(Namespace, isbServiceRolloutName, verifySpecFunc)
+	VerifyISBServiceSpec(Namespace, isbServiceRolloutName, verifySpecFunc)
 
 	verifyISBSvcRolloutReady(isbServiceRolloutName)
 	verifyISBSvcReady(Namespace, isbServiceRolloutName, 3)
 
-	verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
-	verifyPipelineRunning(Namespace, pipelineRolloutName, true)
+	VerifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+	// VerifyPipelineRunning(Namespace, pipelineRolloutName)
```
juliev0, line 427: remove old line

juliev0: oh wait, actually, I guess the logic below still needs to be updated not to use the pipeline name to determine if it's failed, right? I guess this was in the code like that a few places?

dpadhiar: Ack, missed that. Will change.

tests/e2e/numaflowcontroller.go
```diff
@@ -251,7 +254,12 @@ func updateNumaflowControllerRollout(originalVersion, newVersion, pipelineRollou
...
 
-	verifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
+	VerifyInProgressStrategy(pipelineRolloutName, apiv1.UpgradeStrategyNoOp)
 	// need to track vertices of pipeline spec
-	verifyPipelineRunning(Namespace, pipelineRolloutName, false)
+	if strings.Contains(pipelineRolloutName, "failed") {
+		VerifyPipelineFailed(Namespace, pipelineRolloutName)
+	} else {
+		VerifyPipelineRunning(Namespace, pipelineRolloutName, false)
+	}
```
juliev0, line 263: replace this as well

tests/e2e/ppnd-e2e/ppnd_test.go
```diff
@@ -0,0 +1,372 @@
...
+	updatedJetstreamVersion          = "2.10.11"
+)
+
+var (
+	pipelineSpecSourceRPU      = int64(5)
+	pipelineSpecSourceDuration = metav1.Duration{
+		Duration: time.Second,
+	}
+	numVertices         = int32(1)
+	zeroReplicaSleepSec = uint32(15) // if for some reason the Vertex has 0 replicas, this will cause Numaflow to scale it back up
```
juliev0, line 54: @afugazzotto - I believe this `zeroReplicaSleepSec` was supposed to be a temporary thing, right? Just FYI that when you clean up this stuff later, then let's also remove it from this ppnd_test.go file as well. (It may not need to be in this file at all, of course, but it was copied over.)

afugazzotto: This is not related to the workaround I introduced. However, it could be affected once Numaflow allows for scaling vertices to zero.

tests/e2e/ppnd-e2e/ppnd_test.go
```diff
@@ -0,0 +1,372 @@
...
+		failedPipelineSpec := initialPipelineSpec
+		failedPipelineSpec.Edges = append(failedPipelineSpec.Edges, numaflowv1.Edge{From: "not", To: "valid"})
+
+		CreatePipelineRollout(failedPipelineRolloutName, Namespace, failedPipelineSpec, true)
+		VerifyPipelineFailed(Namespace, failedPipelineRolloutName)
+
+		time.Sleep(5 * time.Second)
+
+		// update spec to have topology change
+		failedPipelineSpec = updatedPipelineSpec
```
juliev0, line 255: nit: the variable naming is confusing - in this case, it's a valid pipeline spec, right?

tests/e2e/ppnd-e2e/ppnd_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+		// need to update
+		// this update technically doesnt cause data loss since pipeline for now is failed
```
juliev0: regarding this comment, I think if the `dataLoss` field basically means "the field(s) being changed are fields that result in data loss", then it's good to set this true. Perhaps the comment in the function for what the field means, or possibly even the field, could say that?

juliev0: then you can remove this TODO

tests/e2e/ppnd-e2e/ppnd_test.go
```diff
@@ -0,0 +1,372 @@
...
+		// update would normally cause data loss
+		UpdateISBServiceRollout(isbServiceRolloutName, failedPipelineRolloutName, updatedISBServiceSpec, func(retrievedISBServiceSpec numaflowv1.InterStepBufferServiceSpec) bool {
+			return retrievedISBServiceSpec.JetStream.Version == initialJetstreamVersion
+		}, true, false, false, true)
+
+		DeletePipelineRollout(failedPipelineRolloutName)
+
+	})
+
+	It("Should update a NumaflowControler even if the Pipeline is failed", func() {
```
juliev0, line 292:
```suggestion
	It("Should update a NumaflowController even if the Pipeline is failed", func() {
```

tests/e2e/ppnd-e2e/ppnd_test.go
dpadhiar: Ack, I think the comment originally had to do with me making a few changes to the update function to handle a failed pipeline and the no data loss update followed the logic we needed for that. We should be good now to remove the comment.

tests/e2e/ppnd-e2e/ppnd_test.go
dpadhiar: Good catch, this was a holdover of when I was originally updating the updatePipeline spec to also contain the failing edge.

tests/e2e/numaflowcontroller.go
dpadhiar: I made the change locally for this and pushed it but it's not showing up on Github.. not seen this before.

juliev0: hmm, weird, okay I guess you can do it on a follow-up PR

juliev0 (approved):
any remaining items can be handled in follow-up PR

tests/e2e/ppnd-e2e/ppnd_test.go
juliev0: Okay

Makefile
```diff
@@ -128,7 +128,13 @@ test: codegen fmt vet envtest ## Run tests.
 
 .PHONY: test-e2e
 test-e2e: codegen fmt vet envtest ## Run e2e tests.
-	GOFLAGS="-count=1" go test -timeout 20m -v ./tests/e2e/... 
+	GOFLAGS="-count=1" go test -timeout 20m -v ./tests/e2e/...
```
juliev0, line 131: Actually, I'm kind of thinking maybe `make test-e2e` doesn't make sense to run anymore and I'm wondering if we really need it. Some of the tests require an environment variable; others don't. It would be weird to set the environment variable to "progressive" for example and run this, in which the "progressive" version of the functional test executes and so does the PPND test.

PR #556, opened 2025-02-11
chore: handle cases in which scale/min/max are unset

afugazzotto:
### Modifications

- Handle cases in which scale/min/max are unset
- Created reusable func to calculate scaled min and max
- Pod listing optimization to only get metadata since we only need the pods count

### Verification

Unit tests, e2e tests, and the following manual tests (done for both MonoVertex and Pipeline):
- created MonoVertexRollout with scale.min set to 3 (scale.max unset)
- changed the image to a bad image to first test the failure case. Upgrading child was marked failed and promoted child vertices were restored to original values
- changed the image to a good image which caused the promoted child to scale down to 1 pod and the new upgrading child to be successful.

None.


internal/util/kubernetes/structured_util.go
```diff
@@ -51,3 +51,22 @@ func NewPodDisruptionBudget(name, namespace string, maxUnavailable int32, ownerR
 		},
 	}
 }
+
+func ListPodsMetadataOnly(ctx context.Context, c k8sClient.Client, namespace, labels string) (*metav1.PartialObjectMetadataList, error) {
+	podsMeta := &metav1.PartialObjectMetadataList{}
+
+	err := KubernetesClient.CoreV1().RESTClient().
```
afugazzotto, line 58: Using this instead of previously committed code because that code was giving the following error:
`reflector.go:158] "Unhandled Error" err="pkg/mod/k8s.io/client-go@v0.31.0/tools/cache/reflector.go:243: Failed to watch *v1.PartialObjectMetadata: unknown" logger="UnhandledError"`

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -189,18 +186,22 @@ func scaleDownMonoVertex(
...
+		return true, err
 	}
 
-	actualPodsCount := int64(len(pods.Items))
+	actualPodsCount := int64(len(podsList.Items))
+
+	// If there are no pods, return and do not requeue
+	if actualPodsCount == 0 {
+		return false, nil
+	}
```
juliev0, line 204: I guess this would happen if somebody were to pause the monovertex or pipeline (in their Rollout spec) first - we should try that to make sure it works. How about we add an issue for that if you haven't already tried it?

afugazzotto: This is a very edge case that should not occur. My main concern was related on the pipeline side in which a vertex could be scaled to 0 once I make those changes in Numaflow. For consistency I also brought it to monovertex. Do you think we should remove it though? I'm conflicted on this change tbh.

internal/controller/progressive/progressive.go
```diff
@@ -332,3 +349,60 @@ func IsNumaflowChildReady(upgradingObjectStatus *kubernetes.GenericStatus) bool
...
+	}
+
+	// If max was not found, return nil to later restore the appropriate desired value
+	outMax := &max
+	if !foundMax {
+		outMax = nil
+	}
+
+	return newMin, newMax, outMin, outMax, nil
+}
```
juliev0, line 408: nit: I think this is actually conceptually 2 different functions:
```
func CalculateScaleMinMaxValues(podCount int) (int, int) {
  newMax := int64(math.Floor(float64(podsCount) / float64(2)))
  
  // If min exceeds the newMax, reduce also min to newMax
  newMin := min
  if min > newMax {
     newMin = newMax
  }
  return newMin, newMax
}

func GetSpecDefinedMinMaxValues(object map[string]any, pathToMin, pathToMax []string) (int64, *int64, *int64, error) {
   ....the rest of the logic
}


```

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -189,18 +186,22 @@ func scaleDownMonoVertex(
...
-			// the vertex name for a monovertex is the same as the monovertex name
-			common.LabelKeyNumaflowPodMonoVertexVertexName, promotedChildDef.GetName(),
-		),
-	})
+	podsList, err := kubernetes.ListPodsMetadataOnly(ctx, c, promotedChildDef.GetNamespace(), fmt.Sprintf(
+		"%s=%s, %s=%s",
+		common.LabelKeyNumaflowPodMonoVertexName, promotedChildDef.GetName(),
+		// the vertex name for a monovertex is the same as the monovertex name
+		common.LabelKeyNumaflowPodMonoVertexVertexName, promotedChildDef.GetName(),
+	))
```
juliev0, line 194: nice

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -279,30 +282,40 @@ func scaleMonoVertexToDesiredValues(
...
+		if rolloutPromotedChildStatus.ScaleValues[promotedChildDef.GetName()].DesiredMin != nil {
+			minVal = *rolloutPromotedChildStatus.ScaleValues[promotedChildDef.GetName()].DesiredMin
+		}
+
+		if rolloutPromotedChildStatus.ScaleValues[promotedChildDef.GetName()].DesiredMax != nil {
+			maxVal = *rolloutPromotedChildStatus.ScaleValues[promotedChildDef.GetName()].DesiredMax
+		}
+
+		patchJson = fmt.Sprintf(`{"spec": {"scale": {"min": %v, "max": %v}}}`, minVal, maxVal)
+	}
```
juliev0, line 315: I guess one thing we could've done is just save the original json itself instead of saving these 3 different values but I guess if this is working, that's okay

juliev0: I see there's a function called [unstructed.SetNestedMap()](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#SetNestedMap) - one could theoretically unmarshal the json and then call this

juliev0: Maybe we could explore that as a follow up to see if it simplifies the code? (if it's easy - if not, forget it)

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: > This is a very edge case that should not occur.

I think it will occur. If somebody has a paused pipeline/monovertex and then does a progressive rollout, then the new pipeline will also technically need to be paused as well.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: I decided not to store the entire scale object to avoid one more nesting in the status, especially for pipeline where there could be multiple input sources in that map.

afugazzotto: Using marshalling, this is what it could look like:
`
scale := map[string]any{"min": minVal, "max": maxVal}
patchMap := map[string]any{"spec": map[string]any{"scale": scale}}
patchBytes, err := json.Marshal(patchMap)
if err != nil {
    return true, fmt.Errorf("error marshalling patch JSON: %w", err)
}
patchJson = string(patchBytes)
`
Do you prefer this?

juliev0: > I decided not to store the entire scale object to avoid one more nesting in the status, especially for pipeline where there could be multiple input sources in that map.

Thank you for mentioning multiple sources since I didn't think of that. Now that you mention that, I think the fields we have in there won't work, since there is a single set of fields and really each Source Vertex can have its own fields, right?

internal/controller/progressive/progressive.go
afugazzotto: I agree, I'm just not sure I like to have 2 functions for this logic. I would prefer to rename this function to something like `ExtractScaleValues` or `DestructureScaleValues` or `UnpackScaleValues`.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: This only applies to monovertex which only has one vertex. The pipeline is handled differently. I made that forementioned decision on the status due to the pipeline having multiple sources but I wanted to keep it consistent for monovertex to have some shared code.

internal/controller/progressive/progressive.go
juliev0: Sure...I guess this may be affected by some of the other comments in this PR...

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: I missed that a user could have a paused pipeline and then do progressive upgrade, in which case the vertices could be 0. What I was referring to is related to Numaflow potentially allowing to set a scale value of 0 before starting progressive upgrade process. I see this to be unlikely other than in the cause of pausing a pipeline.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: Oh good, I see. You have a [map](https://github.com/numaproj/numaplane/blob/main/pkg/apis/numaplane/v1alpha1/status_types.go#L142) from vertex name to the ScaleValues struct. 

But then I'm not sure what you mean by "more nesting"?:
> I decided not to store the entire scale object to avoid one more nesting in the status, especially for pipeline where there could be multiple input sources in that map.

juliev0: I feel like the ScaleValues struct could be like:
```
type ScaleValues struct {
	OriginalScaleDefinition string // or alternatively map[string]interface{} I guess
	// ScaleTo indicates how many pods to scale down to
	ScaleTo int64 `json:"scaleTo"`
	// Actual indicates how many pods are actually running for the vertex
	Actual int64 `json:"actual"`
}
```

juliev0: Sorry, I just realized that you do in fact use the "min" that's in there when you scale down. I was originally thinking that we were just saving the spec and restoring it later, and didn't need to look at the actual field values. So, it's up to you I guess if you want to change anything here or not.

internal/controller/progressive/progressive.go
juliev0: I realized these aren't completely orthogonal, since the "min" that's in the spec is used to set the new "min". So, I'm okay if you want to keep as is. Sorry about that.

juliev0:
Can you clarify which manual tests you did?

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: As discussed, we decided to remove this for the moment and reevaluate later.

internal/controller/progressive/progressive.go
afugazzotto: As discussed, we'll keep this as-is.

juliev0 (approved):

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: After our discussion, I decided to try the JSON string solution. Working on it on another branch.

Issue #43, opened 2024-06-06
Be able to define ConfigMap LogLevel as a user-friendly string


juliev0:
# Summary

Currently, LogLevel is an integer, which isn't very user-friendly. We should make it a string (or otherwise have both options).

# Use Cases

When would you use this?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #557, opened 2025-02-11
feat: Be able to define ConfigMap LogLevel as a user-friendly string

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/43

### Modifications

- Define configmap LogLevel as user-friendly string, which supports: "ERROR", "WARN", "INFO", "DEBUG" and "VERBOSE"

### Verification

- Verified in local k8s cluster

- Yes, with this version `LogLevel` should be defined as string instead of `Number`

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"os"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -255,7 +256,7 @@ func loadConfigs() {
 	}
 
 	// anything we need to set based on our main Numaplane Config?
-	numaLogger.SetLevel(config.LogLevel)
+	numaLogger.SetLevel(logger.LogLevelMap[strings.ToUpper(config.LogLevel)])
 	logger.SetBaseLogger(numaLogger)
 	clog.SetLogger(*numaLogger.LogrLogger)
 
--- a/internal/util/logger/logger.go
+++ b/internal/util/logger/logger.go
@@ -49,7 +49,6 @@ const (
 
 // LogLevelMap maps the log level string to the logr verbosity level.
 var LogLevelMap = map[string]int{
-	"ERROR":   FatalLevel,
 	"WARN":    WarnLevel,
 	"INFO":    InfoLevel,
 	"DEBUG":   DebugLevel,
--- a/internal/util/logger/logger.go
+++ b/internal/util/logger/logger.go
@@ -265,6 +265,7 @@ func (nl *NumaLogger) Verbosef(msg string, args ...any) {
 func (nl *NumaLogger) SetLevel(level int) {
 	// default if not defined
 	if level == 0 {
+		log.Printf("LogLevel is not set or unsupported, using default level: INFO")
 		level = defaultLevel
 	}
 
--- a/internal/util/logger/logger.go
+++ b/internal/util/logger/logger.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -45,6 +46,15 @@ const (
 	DebugLevel   = infoLevelShift + 1
 	VerboseLevel = infoLevelShift + 2
 )
+
+// LogLevelMap maps the log level string to the logr verbosity level.
+var LogLevelMap = map[string]int{
+	"ERROR":   FatalLevel,
+	"WARN":    WarnLevel,
+	"INFO":    InfoLevel,
+	"DEBUG":   DebugLevel,
+	"VERBOSE": VerboseLevel,
+}
 
 // The following map define the logr verbosity or NumaLogger semantic levels (constants above) mapping
 // to the zerolog levels (https://github.com/rs/zerolog/blob/master/log.go#L129).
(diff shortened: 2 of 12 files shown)


internal/util/logger/logger.go
```diff
@@ -46,6 +47,15 @@ const (
 	VerboseLevel = infoLevelShift + 2
 )
 
+// LogLevelMap maps the log level string to the logr verbosity level.
+var LogLevelMap = map[string]int{
+	"ERROR":   FatalLevel,
```
juliev0, line 52: Hmm, maybe if this is "FatalLevel", then we should instead do:
`"FATAL": FatalLevel,`?

I am kind of wondering: if we set the level to this value and then we do `numalogger.Error("some error")`, does it get printed or not? If it does, then it is not truly "FatalLevel" but rather "ErrorLevel". 

@afugazzotto you may want to review this PR as well. And if you have any information on why we have a FatalLevel and not an ErrorLevel, let us know. Thanks.

cmd/main.go
```diff
@@ -255,7 +256,7 @@ func loadConfigs() {
 	}
 
 	// anything we need to set based on our main Numaplane Config?
-	numaLogger.SetLevel(config.LogLevel)
+	numaLogger.SetLevel(logger.LogLevelMap[strings.ToUpper(config.LogLevel)])
```
juliev0, line 259: so, if somebody has a string that isn't one of the values, the result of `logger.LogLevelMap[strings.ToUpper(config.LogLevel)]` is 0, and it looks like `numaLogger.SetLevel()` will set to default log level if that happens. Maybe we can at least log an error if that happens?

internal/util/logger/logger.go
chandankumar4: Have tested it via setting `LogLevel=ERROR` in config and observed that it's printing `error` log message as well which it should be if it's fatal.

chandankumar4: Have tested it via setting `LogLevel=ERROR` in config and observed that it's printing `error` log message as well which it should be if it's fatal.

chandankumar4: btw ZeroLog have ErrorLevel logging as well [here](https://github.com/rs/zerolog/blob/master/log.go#L137)

juliev0: > Have tested it via setting LogLevel=ERROR in config and observed that it's printing error log message as well which it should be if it's fatal.

I am thinking that if that truly represented "FatalLevel", then a call to "numaLogger.Error()" should not be logged: only a call to "numaLogger.Fatal()" should be logged. Therefore, I'm thinking that maybe we should change the constant `FatalLevel` to `ErrorLevel`. @chandankumar4 @afugazzotto what do you think?

chandankumar4: We don't have `ErrorLevel` [here](https://github.com/numaproj/numaplane/pull/557/files#diff-72d1bf15c143c98293b1321f13f60e3500c88a4cca40aed1e0bf93c4a165ae3fR43)

chandankumar4: Oh! you mean to rename the constant itself, Currently I'm just trying to introduce error level as well, if that doesn't work, I'll rename the constant

juliev0: > Therefore, I'm thinking that maybe we should change the constant FatalLevel to ErrorLevel

This solution may not be good, because I can see that the [setting](https://github.com/numaproj/numaplane/blob/main/internal/util/logger/logger.go#L55) of `FatalLevel` to 4 is based on what [zerolog says](https://github.com/rs/zerolog) that 4 means ("fatal")

But nonetheless, we need to figure out what the right way is to solve this issue:
> I am thinking that if that truly represented "FatalLevel", then a call to "numaLogger.Error()" should not be logged: only a call to "numaLogger.Fatal()" should be logged.

juliev0: Okay, maybe that constant is somewhat orthogonal to what you're doing @chandankumar4. I can see that from the user perspective, if somebody sets "Error" level and then calls `numaLogger.Error()`, then it is in fact logging an error in our code, and that's good.

juliev0: But it's weird to me to map "ERROR" to "FatalLevel" - so I guess I don't understand why something called `FatalLevel` is allowing Errors to get written in the first place.

chandankumar4: > why something called FatalLevel is allowing Errors to get written in the first place.

yeah, that's the issue I'm trying to figure out!! I'm checking both library https://github.com/go-logr/logr and https://github.com/rs/zerolog how mapping is get setup

chandankumar4: After spending sometime, I observed that our logger implementation is a bit complex which I believe it shouldn't be, I know we need `https://github.com/go-logr/logr` as gitops engine use it, but it should be simple like [this](https://github.com/argoproj/argo-cd/blob/master/util/log/logrus.go#L20), what's your thought?

cmd/main.go
chandankumar4: Added [here](https://github.com/numaproj/numaplane/pull/557/files#diff-72d1bf15c143c98293b1321f13f60e3500c88a4cca40aed1e0bf93c4a165ae3fR268)

internal/util/logger/logger.go
afugazzotto: I think we should remove the FATAL option. Fatal messages and error messages should always be shown independently of the logger level configured. The logger configuration should only allow to set from Warn to Verbose.

juliev0 (approved):

chandankumar4:
I'll merge this PR for now, and if we have to refactor the logger to make it simpler, I'll open up another issue for it

PR #566, opened 2025-02-14
fix: pipelinerollout deletion sequence

### LINKED ISSUE #559: e2e ppnd test - Pipeline is getting stuck during deletion (on the CI)
**Describe the bug**

In the PPND e2e test suite, we have a test case that tries to successfully update a PipelineRollout, ISBServiceRollout and NumaflowControllerRollout if there is a dependent failed Pipeline. Currently the PipelineRollout and NumaflowControllerRollout cases are passing without failure but the ISBServiceRollout test seems to be consistently failing on the CI only.

The failure is consistently occurring after the main feature has been tested successfully and we are cleaning up the failed PipelineRollout. The PipelineRollout is waiting to be verified as deleted and never happens causing a timeout and test suite failure. 

With investigation into the resources artifacts, we see that the child Pipeline has a deletionTimestamp as well as the pipeline-controller finalizer still. This explains why the Pipeline is essentially "hanging" during deletion and informs us the Numaflow controller is not successfully removing the finalizer.

Need to investigate further to see why this only occurs on the CI and not locally. 

```
apiVersion: numaflow.numaproj.io/v1alpha1
kind: Pipeline
metadata:
  creationTimestamp: "2025-02-11T17:09:20Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2025-02-11T17:10:28Z"
  finalizers:
  - pipeline-controller
```

**To Reproduce**

1. Uncomment test and let CI run.

**Expected behavior**

Test to pass successfully on the CI and locally.

**Additional context**

Locally the test is passing which informs us that the issue may be CI related.
A change was made recently to Numaplane which prevents a Rollout from being deleted before its child resource. This change should not have impacted this case or issue but it may have.

---

juliev0:
Fixes #559 

### Modifications

By performing a `Get()` after adding the `foregroundDeletion` finalizer, we can eliminate the resource version conflict issue that occurs. 

### Verification

Running the CI, it seems to pass all of the tests for confirming deletion has occurred.

However, @chandankumar4 will follow up with another PR to address our other CRDs and to do additional testing for different cases. 

N/A

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_controller.go
@@ -347,6 +347,12 @@ func (r *PipelineRolloutReconciler) reconcile(
 			if err := r.client.Delete(ctx, pipelineRollout, &client.DeleteOptions{PropagationPolicy: &foreground}); err != nil {
 				return 0, nil, err
 			}
+			// Get the PipelineRollout live resource
+			livePipelineRollout, err := kubernetes.NumaplaneClient.NumaplaneV1alpha1().PipelineRollouts(pipelineRollout.Namespace).Get(ctx, pipelineRollout.Name, metav1.GetOptions{})
+			if err != nil {
+				return 0, nil, fmt.Errorf("error getting the live PipelineRollout: %w", err)
+			}
+			*pipelineRollout = *livePipelineRollout
 			controllerutil.RemoveFinalizer(pipelineRollout, common.FinalizerName)
 		}
 		// generate the metrics for the Pipeline deletion.
--- a/tests/e2e/ppnd-e2e/ppnd_test.go
+++ b/tests/e2e/ppnd-e2e/ppnd_test.go
@@ -264,35 +264,35 @@ var _ = Describe("Pause and drain e2e", Serial, func() {
 
 	time.Sleep(5 * time.Second)
 
-	// It("Should update an ISBService even if the Pipeline is failed", func() {
+	It("Should update an ISBService even if the Pipeline is failed", func() {
 
-	// 	// add bad edge to automatically fail Pipeline
-	// 	failedPipelineSpec := initialPipelineSpec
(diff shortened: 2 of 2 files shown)


internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -347,6 +347,12 @@ func (r *PipelineRolloutReconciler) reconcile(
 			if err := r.client.Delete(ctx, pipelineRollout, &client.DeleteOptions{PropagationPolicy: &foreground}); err != nil {
```
juliev0, line 347: It looks like calling `r.client.Delete()` doesn't actually update the pipelineRollout to reflect the new Finalizer called `foregroundDeletion` and the new resource version. So, the subsequent call would sometimes fail to update. 

For some reason in the PPND test that was commented out in here, it basically would always fail the same way - in a constant cycle of never being to update. 

We can either do what I have here which is to Get the live object after doing the Update(), or we could change the Update() to a Patch().

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -347,6 +347,12 @@ func (r *PipelineRolloutReconciler) reconcile(
...
 				return 0, nil, err
 			}
+			// Get the PipelineRollout live resource
+			livePipelineRollout, err := kubernetes.NumaplaneClient.NumaplaneV1alpha1().PipelineRollouts(pipelineRollout.Namespace).Get(ctx, pipelineRollout.Name, metav1.GetOptions{})
+			if err != nil {
+				return 0, nil, fmt.Errorf("error getting the live PipelineRollout: %w", err)
+			}
+			*pipelineRollout = *livePipelineRollout
```
juliev0, line 355: have to set this as well since the `pipelineRollout` is what's returned in the function argument

chandankumar4 (approved):

PR #569, opened 2025-02-14
fix: promoted child status was only being copied partially from live rollout to in-memory rollout

juliev0:
### Modifications

There's a bug in which the entire `PromotedPipelineStatus` and `PromotedMonoVertexStatus` is no longer being propagated from the `liveRolloutObject` to the in-memory `rolloutObject` during Progressive update. This happened when I refactored the ProgressiveStatus struct to separate it into multiple. 

To remedy that, I made it so that the individual Rollout objects copy the liveRolloutObject's ProgressiveStatus over to the in-memory object each time we reconcile progressively, and then from there we only use that.  

Also fixes a flakey e2e test for MonoVertexRollout which had potential to fail for progressive case (described inline).

### Verification

I did some manual tests of the Pipeline and figured given that the MonoVertex was changed in the exact same way, it should hopefully be okay as well:

- good pipeline -> bad pipeline -> good pipeline
- good pipeline -> good pipeline

I observed each time that it took the promoted one down to the correct number of vertices before upgrading, and then back up on failure.

I also ran the e2e tests 4 times in a row, and it passed every time.

N/A

### DIFF
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -418,13 +418,16 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 	case apiv1.UpgradeStrategyProgressive:
 		numaLogger.Debug("processing InterstepBufferService with Progressive")
 
-		// Get the ISBServiceRollout live resource
+		// Get the ISBServiceRollout live resource so we can grab the ProgressiveStatus from that for our own local isbServiceRollout
+		// (Note we don't copy the entire Status in case we've updated something locally)
 		liveISBServiceRollout, err := kubernetes.NumaplaneClient.NumaplaneV1alpha1().ISBServiceRollouts(isbServiceRollout.Namespace).Get(ctx, isbServiceRollout.Name, metav1.GetOptions{})
 		if err != nil {
 			return 0, fmt.Errorf("error getting the live ISBServiceRollout for assessment processing: %w", err)
 		}
 
-		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, liveISBServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
+		isbServiceRollout.Status.ProgressiveStatus = *liveISBServiceRollout.Status.ProgressiveStatus.DeepCopy()
+
+		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
 		if err != nil {
 			return 0, fmt.Errorf("Error processing isbsvc with progressive: %s", err.Error())
 		}
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -76,7 +76,6 @@ type ProgressiveRolloutObject interface {
 func ProcessResource(
 	ctx context.Context,
 	rolloutObject ProgressiveRolloutObject,
-	liveRolloutObject ProgressiveRolloutObject,
 	existingPromotedChild *unstructured.Unstructured,
 	promotedDifference bool,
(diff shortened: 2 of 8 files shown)


tests/e2e/monovertex.go
```diff
@@ -352,7 +352,7 @@ func DeleteMonoVertexRollout(name string) {
 	}).WithTimeout(testTimeout).Should(BeTrue(), "The MonoVertex should have been deleted but it was found.")
 }
 
-func UpdateMonoVertexRollout(name string, newSpec numaflowv1.MonoVertexSpec, expectedFinalPhase numaflowv1.MonoVertexPhase) {
+func UpdateMonoVertexRollout(name string, newSpec numaflowv1.MonoVertexSpec, expectedFinalPhase numaflowv1.MonoVertexPhase, verifySpecFunc func(numaflowv1.MonoVertexSpec) bool) {
```
juliev0, line 355: we need to make sure we verify the mono vertex spec before verifying anything else - by verifying the monovertex spec, we ensure our "promoted" monovertex is the new one

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -418,13 +418,16 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
...
+		// (Note we don't copy the entire Status in case we've updated something locally)
 		liveISBServiceRollout, err := kubernetes.NumaplaneClient.NumaplaneV1alpha1().ISBServiceRollouts(isbServiceRollout.Namespace).Get(ctx, isbServiceRollout.Name, metav1.GetOptions{})
 		if err != nil {
 			return 0, fmt.Errorf("error getting the live ISBServiceRollout for assessment processing: %w", err)
 		}
 
-		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, liveISBServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
+		isbServiceRollout.Status.ProgressiveStatus = *liveISBServiceRollout.Status.ProgressiveStatus.DeepCopy()
+
+		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
```
juliev0, line 430: just pass one version of the rollout in here now

internal/controller/progressive/progressive.go
```diff
@@ -274,10 +269,8 @@ func processUpgradingChild(
 
 		// if so, mark the existing one for garbage collection and then create a new upgrading one
 		if needsUpdating {
-			if liveRolloutObject.GetPromotedChildStatus() == nil {
+			if rolloutObject.GetPromotedChildStatus() == nil {
 				rolloutObject.SetPromotedChildStatus(&apiv1.PromotedChildStatus{Name: existingPromotedChildDef.GetName()})
-			} else {
-				rolloutObject.SetPromotedChildStatus(liveRolloutObject.GetPromotedChildStatus().DeepCopy())
```
juliev0, line 280: no need for this copy - we already copied the liveRolloutObject's entire Progressive Status into here

tests/e2e/pipeline.go
```diff
@@ -516,7 +516,7 @@ func UpdatePipelineRollout(name string, newSpec numaflowv1.PipelineSpec, expecte
 	}
 
 	// rollout phase will be pending if we are expecting a long pausing state and Pipeline will not be fully updated
-	if expectedFinalPhase != numaflowv1.PipelinePhasePausing {
+	if !(UpgradeStrategy == config.PPNDStrategyID && expectedFinalPhase == numaflowv1.PipelinePhasePausing) {
```
juliev0, line 519: this is just a clarification that this is related to the PPND strategy

juliev0:
I have a new [PR](https://github.com/numaproj/numaplane/pull/570) which adds more to this one, so putting this one into "Draft"

juliev0:
no need for this PR as I have a new one, so closing

PR #570, opened 2025-02-15
fix: various bugs related to progressive rollout

### LINKED ISSUE #550: race condition on pipeline/isbsvc blue/green deletion
**Describe the bug**
A successful blue/green Pipeline can be marked "recyclable" and then deleted, but ISBService `AssessUpgradingChild()` seems to look for the Pipeline child. Seems like it may still be performing its assessment after the Pipeline gets deleted?

---

juliev0:
Fixes #550 

### Modifications

This fixes a few issues (all of them are described inline):
- Reset the entire Upgrading Child Status and Promoted Child Status instead of just setting the inner structs that are common to all Rollouts
- Make sure to reset the Promoted Child Set if there's a new promoted child (it didn't look like we were)
- Make sure the Assessment of ISBService isn't dependent on the Pipeline still being there (it could be deleted)
- Premature success was being declared if it succeeded just once within the time window
- copy/paste error
- functional test bug for progressive case

### Verification

Ran e2e test many times, all passing.

**Manual testing:**
For **MonoVertex**, I did a test in which I:
1. Create MonoVertexRollout with scale.min/max=3,6
2. Update image to a bad one
3. Observe initial MonoVertex scale down to 1 Pod
4. Observe NextAssessmentTime gets set, then AssessUntil after the first assessment is made
5. Observe it's marked Failed as soon as the first failure is recorded
6. Observed promoted MonoVertex gets scaled back up
7. Restore MonoVertexRollout back to original image 
8. Observe same steps as above happen except that determination of "Success" happens only once "AssessUntil" time is reached

For **InterstepBufferService** and **Pipeline**, I did a test in which I:
1. Create ISBServiceRollout and PipelineRollout
2. Update ISBServiceRollout to a different version
3. Observe initial Pipeline scale down
4. Observe that Successful assessment is only made for each after the AssessUntil time is reached
5. Observe new pipeline and new isbservice get promoted

None

### DIFF
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -418,13 +418,16 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 	case apiv1.UpgradeStrategyProgressive:
 		numaLogger.Debug("processing InterstepBufferService with Progressive")
 
-		// Get the ISBServiceRollout live resource
+		// Get the ISBServiceRollout live resource so we can grab the ProgressiveStatus from that for our own local isbServiceRollout
+		// (Note we don't copy the entire Status in case we've updated something locally)
 		liveISBServiceRollout, err := kubernetes.NumaplaneClient.NumaplaneV1alpha1().ISBServiceRollouts(isbServiceRollout.Namespace).Get(ctx, isbServiceRollout.Name, metav1.GetOptions{})
 		if err != nil {
 			return 0, fmt.Errorf("error getting the live ISBServiceRollout for assessment processing: %w", err)
 		}
 
-		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, liveISBServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
+		isbServiceRollout.Status.ProgressiveStatus = *liveISBServiceRollout.Status.ProgressiveStatus.DeepCopy()
+
+		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
 		if err != nil {
 			return 0, fmt.Errorf("Error processing isbsvc with progressive: %s", err.Error())
 		}
--- a/internal/controller/isbservicerollout/isbservicerollout_progressive.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_progressive.go
@@ -63,7 +63,7 @@ func (r *ISBServiceRolloutReconciler) AssessUpgradingChild(ctx context.Context, 
 	for _, pipelineRollout := range pipelineRollouts {
 		upgradingPipelineStatus := pipelineRollout.Status.ProgressiveStatus.UpgradingPipelineStatus
 		if upgradingPipelineStatus == nil || upgradingPipelineStatus.InterStepBufferServiceName != existingUpgradingChildDef.GetName() {
-			numaLogger.WithValues("pipeline", upgradingPipelineStatus.Name).Debug("can't assess ISBService; pipeline is not yet upgrading with this ISBService")
+			numaLogger.WithValues("pipelinerollout", pipelineRollout.GetName()).Debug("can't assess ISBService; pipeline is not yet upgrading with this ISBService")
 			return apiv1.AssessmentResultUnknown, nil
 		}
 		switch pipelineRollout.Status.ProgressiveStatus.UpgradingPipelineStatus.AssessmentResult {
@@ -73,8 +73,12 @@ func (r *ISBServiceRolloutReconciler) AssessUpgradingChild(ctx context.Context, 
 		case apiv1.AssessmentResultUnknown:
 			numaLogger.WithValues("pipeline", upgradingPipelineStatus.Name).Debug("pipeline assessment is unknown")
 			return apiv1.AssessmentResultUnknown, nil
+		case apiv1.AssessmentResultSuccess:
+			numaLogger.WithValues("pipeline", upgradingPipelineStatus.Name).Debug("pipeline succeeded")
+
 		}
 	}
+
 	return apiv1.AssessmentResultSuccess, nil
 }
 
--- a/internal/controller/isbservicerollout/isbservicerollout_progressive.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_progressive.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/numaproj/numaplane/internal/common"
-	ctlrcommon "github.com/numaproj/numaplane/internal/controller/common"
 	"github.com/numaproj/numaplane/internal/controller/progressive"
 	"github.com/numaproj/numaplane/internal/util/logger"
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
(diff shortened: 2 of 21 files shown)


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -418,13 +418,16 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
...
-		// Get the ISBServiceRollout live resource
+		// Get the ISBServiceRollout live resource so we can grab the ProgressiveStatus from that for our own local isbServiceRollout
+		// (Note we don't copy the entire Status in case we've updated something locally)
 		liveISBServiceRollout, err := kubernetes.NumaplaneClient.NumaplaneV1alpha1().ISBServiceRollouts(isbServiceRollout.Namespace).Get(ctx, isbServiceRollout.Name, metav1.GetOptions{})
 		if err != nil {
 			return 0, fmt.Errorf("error getting the live ISBServiceRollout for assessment processing: %w", err)
 		}
 
-		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, liveISBServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
+		isbServiceRollout.Status.ProgressiveStatus = *liveISBServiceRollout.Status.ProgressiveStatus.DeepCopy()
```
juliev0, line 428: just copy the ProgressiveStatus from the live resource here so that in the progressive code, we can just one have one single version of our rollout object

internal/controller/isbservicerollout/isbservicerollout_progressive.go
```diff
@@ -59,58 +58,27 @@ func (r *ISBServiceRolloutReconciler) AssessUpgradingChild(ctx context.Context,
...
-				foundPipeline = true
-				break
-			}
-		}
-		if !foundPipeline {
-			numaLogger.Debugf("Can't assess isbsvc; didn't find a Pipeline associated with PipelineRollout %s/%s using this isbsvc",
-				pipelineRollout.GetNamespace(), pipelineRollout.GetName())
+		upgradingPipelineStatus := pipelineRollout.Status.ProgressiveStatus.UpgradingPipelineStatus
+		if upgradingPipelineStatus == nil || upgradingPipelineStatus.InterStepBufferServiceName != existingUpgradingChildDef.GetName() {
+			numaLogger.WithValues("pipelinerollout", pipelineRollout.GetName()).Debug("can't assess ISBService; pipeline is not yet upgrading with this ISBService")
```
juliev0, line 66: previously we looked for the `Pipeline` itself to determine if it's using this isbsvc. This was a problem, because the Pipeline could have already been deleted. Now we use the `PipelineRollout.ProgressiveStatus.UpgradingChildStatus.InterstepBufferServiceName` to confirm that the upgrading Pipeline that we're getting an AssessmentResult from is in fact using this isbsvc

internal/controller/isbservicerollout/isbservicerollout_progressive.go
```diff
@@ -123,7 +91,7 @@ func (r *ISBServiceRolloutReconciler) ProcessPromotedChildPreUpgrade(
 	return false, nil
 }
 
-func (r *ISBServiceRolloutReconciler) ProcessPromotedChildPostUpgrade(
+func (r *ISBServiceRolloutReconciler) ProcessPromotedChildPostFailure(
```
juliev0, line 94: I thought maybe this was a more accurate name for this case, since it's only the failure case and not the success case

internal/controller/progressive/progressive.go
```diff
@@ -99,10 +104,12 @@ func ProcessResource(
...
-			if liveRolloutObject.GetPromotedChildStatus() == nil {
-				rolloutObject.SetPromotedChildStatus(&apiv1.PromotedChildStatus{Name: existingPromotedChild.GetName()})
-			} else {
-				rolloutObject.SetPromotedChildStatus(liveRolloutObject.GetPromotedChildStatus().DeepCopy())
+			promotedChildStatus := rolloutObject.GetPromotedChildStatus()
+			if promotedChildStatus == nil || promotedChildStatus.Name != existingPromotedChild.GetName() {
+				err = rolloutObject.ResetPromotedChildStatus(existingPromotedChild)
+				if err != nil {
+					return false, false, 0, err
+				}
```
juliev0, line 112: I figure we need to also Reset this in the case that the promoted child name has changed from before

juliev0: Also, by calling `ResetPromotedChildStatus()`, we now reset the entire encapsulating promoted child struct for that particular Rollout.

internal/controller/progressive/progressive.go
```diff
@@ -274,12 +279,6 @@ func processUpgradingChild(
 
 		// if so, mark the existing one for garbage collection and then create a new upgrading one
 		if needsUpdating {
-			if liveRolloutObject.GetPromotedChildStatus() == nil {
-				rolloutObject.SetPromotedChildStatus(&apiv1.PromotedChildStatus{Name: existingPromotedChildDef.GetName()})
-			} else {
-				rolloutObject.SetPromotedChildStatus(liveRolloutObject.GetPromotedChildStatus().DeepCopy())
-			}
```
juliev0, line 281: I relocated the setting of Promoted Child Status to the beginning of the `progressive.ProcessResource()` call - we should have the same promoted child throughout (up until the end of the Success case at least)

internal/controller/progressive/progressive.go
```diff
@@ -85,6 +90,15 @@ func ProcessResource(
...
 	numaLogger := logger.FromContext(ctx)
 
+	// Make sure that our Promoted Child Status reflects the current promoted child
+	promotedChildStatus := rolloutObject.GetPromotedChildStatus()
+	if promotedChildStatus == nil || promotedChildStatus.Name != existingPromotedChild.GetName() {
+		err := rolloutObject.ResetPromotedChildStatus(existingPromotedChild)
+		if err != nil {
+			return false, false, 0, err
+		}
+	}
```
juliev0, line 100: I think we can just call this once at the beginning if either the PromotedChildStatus is nil or if there's a new promoted child. 

By calling `ResetPromotedChildStatus()`, we can update the [InterstepBufferServiceName](https://github.com/numaproj/numaplane/blob/pipeline-ref-to-isbsvc/pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go#L69) for UpgradingPipelineStatus

pkg/apis/numaplane/v1alpha1/isbservicerollout_types.go
```diff
@@ -105,8 +106,8 @@ func (isbServiceRollout *ISBServiceRollout) GetRolloutGVK() schema.GroupVersionK
 
 func (isbServiceRollout *ISBServiceRollout) GetChildGVR() metav1.GroupVersionResource {
 	return metav1.GroupVersionResource{
-		Group:    numaflowv1.PipelineGroupVersionKind.Group,
-		Version:  numaflowv1.PipelineGroupVersionKind.Version,
+		Group:    numaflowv1.ISBGroupVersionKind.Group,
+		Version:  numaflowv1.ISBGroupVersionKind.Version,
```
juliev0, line 110: this was a copy/paste error

tests/e2e/functional-e2e/functional_test.go
```diff
@@ -303,7 +303,9 @@ var _ = Describe("Functional e2e", Serial, func() {
 		Document("setting desiredPhase=Paused")
 		currentMonoVertexSpec.Lifecycle.DesiredPhase = numaflowv1.MonoVertexPhasePaused
 
-		UpdateMonoVertexRollout(monoVertexRolloutName, currentMonoVertexSpec, numaflowv1.MonoVertexPhasePaused)
+		UpdateMonoVertexRollout(monoVertexRolloutName, currentMonoVertexSpec, numaflowv1.MonoVertexPhasePaused, func(spec numaflowv1.MonoVertexSpec) bool {
+			return spec.Lifecycle.DesiredPhase == numaflowv1.MonoVertexPhasePaused
+		})
```
juliev0, line 308: this fixes the unit test (described below)

tests/e2e/monovertex.go
```diff
@@ -362,8 +362,11 @@ func UpdateMonoVertexRollout(name string, newSpec numaflowv1.MonoVertexSpec, exp
 		rollout.Spec.MonoVertex.Spec.Raw = rawSpec
 		return rollout, nil
 	})
+	Document("Verifying MonoVertex spec got updated")
+	// get Pipeline to check that spec has been updated to correct spec
+	VerifyMonoVertexSpec(Namespace, name, verifySpecFunc)
```
juliev0, line 367: this fixes the unit test - before we weren't waiting for the spec to get updated, which means that we could start verifying the previous "promoted" MonoVertex for various checks, rather than verifying the new one we're updating

tests/e2e/pipeline.go
```diff
@@ -516,7 +516,7 @@ func UpdatePipelineRollout(name string, newSpec numaflowv1.PipelineSpec, expecte
 	}
 
 	// rollout phase will be pending if we are expecting a long pausing state and Pipeline will not be fully updated
-	if expectedFinalPhase != numaflowv1.PipelinePhasePausing {
+	if !(UpgradeStrategy == config.PPNDStrategyID && expectedFinalPhase == numaflowv1.PipelinePhasePausing) {
```
juliev0, line 519: this is a clarification that this check is really intended for PPND

internal/controller/progressive/progressive.go
```diff
@@ -322,26 +318,31 @@ func processUpgradingChild(
...
 
 	case apiv1.AssessmentResultSuccess:
-		// Label the new child as promoted and then remove the label from the old one
-		numaLogger.WithValues("old child", existingPromotedChildDef.GetName(), "new child", existingUpgradingChildDef.GetName(), "replacing 'promoted' child")
-		reasonSuccess := common.LabelValueProgressiveSuccess
-		err := ctlrcommon.UpdateUpgradeState(ctx, c, common.LabelValueUpgradePromoted, &reasonSuccess, existingUpgradingChildDef)
-		if err != nil {
-			return false, false, 0, err
-		}
+		if childStatus.CanDeclareSuccess() {
```
juliev0, line 321: I'm looking for `>=AssessUntil` here. Now the call to `CanAssess()` above only looks for whether it's reached the `NextAssessmentTime` (and whether it's previously failed). 

So, basically:
- if we haven't reached `NextAssessmentTime`, we don't assess
- Once we reach `NextAssessmentTime`, we can start assessing. If we see a failure, we can mark it and stop assessing.
- Once we reach `AssessUntil`, we can declare success if we never saw a failure.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -209,6 +209,8 @@ func scaleDownMonoVertex(
 	// If for the vertex we already set a Scaled scale value, we only need to update the actual pods count
 	// to later verify that the pods were actually scaled down.
 	// We want to skip scaling down again.
+	// TODO: why do we concern ourselves if vertexScaleValues.ScaleTo != 0? If we scale 1 Pod to 0, we go past here even if the scale values
+	// were found
```
juliev0, line 213: need more info from @afugazzotto on this

internal/controller/progressive/progressive.go
juliev0: With this, we basically reset the Promoted status whenever there's a new promoted child that starts the upgrading process.

internal/controller/progressive/progressive.go
```diff
@@ -322,26 +318,31 @@ func processUpgradingChild(
...
+			rolloutObject.GetRolloutStatus().MarkProgressiveUpgradeSucceeded(fmt.Sprintf("New Child Object %s/%s Running", existingUpgradingChildDef.GetNamespace(), existingUpgradingChildDef.GetName()), rolloutObject.GetRolloutObjectMeta().Generation)
+			childStatus.AssessmentResult = apiv1.AssessmentResultSuccess
+			rolloutObject.SetUpgradingChildStatus(childStatus)
+			rolloutObject.GetRolloutStatus().MarkDeployed(rolloutObject.GetRolloutObjectMeta().Generation)
+
+			// we're done
+			return true, false, assessmentInterval, nil
+		} else {
+			return false, false, assessmentInterval, nil
+		}
```
juliev0, line 345: only these lines changed I believe. Now, we can always return true if we've reached the end of the time window

tests/e2e/common.go
```diff
@@ -44,9 +44,7 @@ var (
 	dynamicClient       dynamic.DynamicClient
 	testEnv             *envtest.Environment
 	ctx                 context.Context
-	cancel              context.CancelFunc
-	suiteTimeout        = 35 * time.Minute // Note: if we start seeing "client rate limiter: context deadline exceeded", we need to increase this value
```
juliev0, line 48: I removed `suiteTimeout` and the timeout that was in Github Actions for the e2e test. Now there's just the one timeout in the Makefile target for `go test` itself. I think that should be sufficient.

internal/controller/progressive/progressive.go
```diff
@@ -322,26 +318,31 @@ func processUpgradingChild(
...
-			return false, false, 0, err
-		}
+		if childStatus.CanDeclareSuccess() {
 
-		err = ctlrcommon.UpdateUpgradeState(ctx, c, common.LabelValueUpgradeRecyclable, &reasonSuccess, existingPromotedChildDef)
-		if err != nil {
-			return false, false, 0, err
-		}
+			// Label the new child as promoted and then remove the label from the old one
+			numaLogger.WithValues("old child", existingPromotedChildDef.GetName(), "new child", existingUpgradingChildDef.GetName(), "replacing 'promoted' child")
```
afugazzotto, line 324: You may want to log the message using Debug or other level.

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -132,7 +132,7 @@ func (r *PipelineRolloutReconciler) ProcessPromotedChildPreUpgrade(
 }
 
 /*
-ProcessPromotedChildPostUpgrade handles the post-upgrade processing of a promoted pipeline.
+ProcessPromotedChildPostFailure handles the post-upgrade processing of a promoted pipeline.
```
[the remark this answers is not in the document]
juliev0: good catch

juliev0:
@afugazzotto I made the changes based on your comments

afugazzotto (approved):

PR #574, opened 2025-02-18
chore: update e2e functional test to confirm each Rollout only has one child at the end; also consolidate some logic into a function

### LINKED ISSUE #547: e2e test to check for garbage collection
# Summary

We want to make sure that in the end, assuming everything succeeded, we should only have one child.

In our functional test, after we perform an update of a given Rollout, we can confirm that "Eventually" there is only one child of that Rollout. (For ISBSvc update, this would be related to the isbsvc children as well as the pipeline children)

---

juliev0:
Fixes #547 

### Modifications

Decided to consolidate some common logic into a new `startUpgradeProcess()` function.

Also added a check to the functional test to confirm that there's only one child per Rollout by the end (the others should get garbage collected). 

### Verification

Wanted to verify I didn't mess anything up, so I ran these manual tests:

- Update PipelineRollout from good -> bad -> good and observe the Status of the PipelineRollout to make sure it's right
- Update ISBServiceRollout and observe the Status of the PipelineRollout and ISBServiceRollout to make sure it's right

N/A

### DIFF
--- a/internal/controller/common/rollout_controller.go
+++ b/internal/controller/common/rollout_controller.go
@@ -145,7 +145,6 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
 			}
 		}
 		// recycle the previous children
-		// TODO: do we want this? consider the cases that would actually use it
 		for _, recyclableChild := range recycleList {
 			numaLogger.Debugf("found multiple children of Rollout %s/%s of upgrade state=%q, marking recyclable: %s",
 				rolloutObject.GetRolloutObjectMeta().Namespace, rolloutObject.GetRolloutObjectMeta().Name, upgradeState, recyclableChild.GetName())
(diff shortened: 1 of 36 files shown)


juliev0:
@afugazzotto - I made the changes based on your comments.

internal/controller/progressive/progressive.go
```diff
@@ -303,8 +283,7 @@ func processUpgradingChild(
 				return false, false, 0, err
 			}
 
-			err = kubernetes.CreateResource(ctx, c, newUpgradingChildDef)
-			return false, true, 0, err
+			return false, newUpgradingChildDef != nil, 0, nil
```
juliev0, line 286: One thing to look at is that I ended up changing the order of marking the old "upgrading" one recyclable and creating the new "upgrading" one. Say we successfully create the new one and get a conflict error marking the previous one "recyclable". Then we will temporarily have 2 that are marked "upgrading". When we come back around to reconcile again, we will call [FindMostCurrentChildOfUpgrade()](https://github.com/numaproj/numaplane/blob/main/internal/controller/progressive/progressive.go#L103) and that will actually [mark](https://github.com/numaproj/numaplane/blob/main/internal/controller/common/rollout_controller.go#L150) the older one Recyclable.

internal/controller/common/rollout_controller.go
```diff
@@ -145,7 +145,6 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
 			}
 		}
 		// recycle the previous children
-		// TODO: do we want this? consider the cases that would actually use it
```
juliev0, line 148: I found a good use case, mentioned below, so I'm removing this "TODO"

dpadhiar (approved):
LGTM!

PR #579, opened 2025-02-19
fix: (E2E) Remove Document fn call and refactor

### LINKED ISSUE #500: Improve error messaging in e2e test
# Summary

When there's a failure in a given test, it can be hard to figure out why it failed. 

**Issue 1: forgetting to call `document()`:**
Right now we have the `document()` statements in there which give us some idea where we are in the test. When we forget to add them prior to every test, we can't tell where it failed. Perhaps there's some structural change that would make it so we always have them? Here's an idea: maybe we never call `Eventually` or `Consistently` directly: instead we call some function which calls those, and it also takes a string argument which we print. 

**Issue 2: panics:**
I also know that once when an Eventually() test panicked in its function block, the panic wasn't displayed.

**Issue 3:**
This one is not as critical but is it possible to print more information when an `Eventually` or `Consistently` test fails and contains multiple test conditions in it? Maybe we can start to figure out possible strategies.

---

chandankumar4:
Fixes #500 

### Modifications

- Removing Documentation fn call and created a wrapper method for `Eventually()` and `Consistently()` for printing the test case name as well while run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ test: codegen fmt vet envtest ## Run unit tests.
 test-functional-e2e:
 test-ppnd-e2e:
 test-%: envtest ## Run e2e tests. Note we may need to increase the timeout in the future.
-	GOFLAGS="-count=1" ginkgo -v ./tests/e2e/$*
+	GOFLAGS="-count=1" ginkgo run -v --timeout 35m ./tests/e2e/$*
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 GOLANGCI_LINT_VERSION ?= v1.61.0
--- a/Makefile
+++ b/Makefile
@@ -128,10 +128,8 @@ test: codegen fmt vet envtest ## Run unit tests.
 
 test-functional-e2e:
 test-ppnd-e2e:
-test-%: codegen fmt vet envtest ## Run e2e tests. Note we may need to increase the timeout in the future.
-	GOFLAGS="-count=1" go test -timeout 35m -v ./tests/e2e/$* 
-
-
+test-%: envtest ## Run e2e tests. Note we may need to increase the timeout in the future.
+	GOFLAGS="-count=1" ginkgo -v ./tests/e2e/$*
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 GOLANGCI_LINT_VERSION ?= v1.61.0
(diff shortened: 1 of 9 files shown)


tests/e2e/common.go
```diff
@@ -571,3 +553,13 @@ func setupOutputDir() {
...
+
+func CheckEventually(testData string, actualOrCtx interface{}) AsyncAssertion {
+	By(testData)
+	return Eventually(actualOrCtx, testTimeout, testPollingInterval)
+}
+
+func CheckConsistently(testData string, actualOrCtx interface{}) AsyncAssertion {
+	By(testData)
+	return Consistently(actualOrCtx, testTimeout, testPollingInterval)
+}
```
juliev0, line 565: I want to get this merged sooner than later so you don't have merge conflicts, but one thing I'm thinking is that it would be great to be able to optionally specify the timeout or the polling interval if we need to.

juliev0: oh wait, cool, I see you were able to do that by calling `CheckEventually(...).WithTimeout(...)`, right?

juliev0: maybe we can just comment these functions to clarify that you can override the default timeout and default poll interval that way?

juliev0 (approved):

juliev0:
sorry there are merge conflicts! I've approved, so feel free to fix the conflicts, add the comments I mentioned [here](https://github.com/numaproj/numaplane/pull/579#discussion_r1962871979), and then merge yourself

Makefile
```diff
@@ -128,10 +128,8 @@ test: codegen fmt vet envtest ## Run unit tests.
 
 test-functional-e2e:
 test-ppnd-e2e:
-test-%: codegen fmt vet envtest ## Run e2e tests. Note we may need to increase the timeout in the future.
-	GOFLAGS="-count=1" go test -timeout 35m -v ./tests/e2e/$* 
-
-
+test-%: envtest ## Run e2e tests. Note we may need to increase the timeout in the future.
+	GOFLAGS="-count=1" ginkgo -v ./tests/e2e/$*
```
juliev0, line 132: is there a "timeout" flag for ginkgo call? 

Note that the comment above still refers to a timeout. Also, I should've mentioned it, but in a recent [PR](https://github.com/numaproj/numaplane/pull/570/files), I noticed that we were timing out in 3 different places (here in the Makefile, in the Github action, and in the ginkgo code itself), so I decided to only leave this one and remove the other 2. 

If we can't do it here, then perhaps we should add one of the other ones back.

chandankumar4: Added timeout here with `ginkgo run`

juliev0: nice

PR #582, opened 2025-02-19
fix: scaling to zero behavior fix

### LINKED ISSUE #584: Make sure we only scale down once
**Describe the bug**

Investigate the TODO [here](https://github.com/numaproj/numaplane/blob/577c1943d67f3e829b5087ea2626e62faf2d2b90/internal/controller/monovertexrollout/monovertexrollout_progressive.go#L213).

---

afugazzotto:
Fixes #584 

### Modifications

Removed the check for which the scaleTo value should not be 0.
On the E2E tests, added a `scaleInVertex` arg on the `UpdatePipelineRollout` as a workaround to force scale up the `in` vertex back to 3 pods since after pausing a pipeline and resuming it, the scale values are ignored by Numaflow and we have to manually change the replicas on the vertex. This is similar to the scale down workaround adopted on the other E2E tests and that will be removed once the Numaflow changes/fixes are available. 

### Verification

Manually tested by creating a MonoVertex and Pipeline with no scale values set.

None.

### DIFF
--- a/tests/e2e/functional-e2e/functional_test.go
+++ b/tests/e2e/functional-e2e/functional_test.go
@@ -273,7 +273,7 @@ var _ = Describe("Functional e2e:", Serial, func() {
 		numPipelineVertices := len(updatedPipelineSpec.Vertices)
 		UpdatePipelineRollout(pipelineRolloutName, updatedPipelineSpec, numaflowv1.PipelinePhaseRunning, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
 			return len(retrievedPipelineSpec.Vertices) == numPipelineVertices
-		}, true, true)
+		}, true, true, false)
 
 	})
 
@@ -288,7 +288,7 @@ var _ = Describe("Functional e2e:", Serial, func() {
 
 		UpdatePipelineRollout(pipelineRolloutName, currentPipelineSpec, numaflowv1.PipelinePhasePaused, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
 			return retrievedPipelineSpec.Lifecycle.DesiredPhase == numaflowv1.PipelinePhasePaused
-		}, false, false)
+		}, false, false, false)
 
 		VerifyPipelineStaysPaused(pipelineRolloutName)
 	})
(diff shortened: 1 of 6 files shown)


tests/e2e/numaflowcontroller.go
```diff
@@ -260,7 +260,7 @@ func UpdateNumaflowControllerRollout(originalVersion, newVersion string, pipelin
 		if rolloutInfo.PipelineIsFailed {
 			VerifyPipelineFailed(Namespace, rolloutName)
 		} else {
-			VerifyPipelineRunning(Namespace, rolloutName, false)
+			VerifyPipelineRunning(Namespace, rolloutName, true)
```
juliev0, line 263: cc @dpadhiar (I know we were looking at this boolean). I think setting it to true didn't work for you, right?

dpadhiar: Yes thats right!

afugazzotto: Setting it to `true` will cause the validation func to check the number of pods against the scaled value if in progressive strategy. This will all be removed once the Numaflow changes are available.

dpadhiar: That sounds good. The issue I was seeing in my case was that if we paused and unpaused a pipeline manually, the pipeline would only have 3 pods come back up for the vertices not the 9 we should expect.

dpadhiar: Taking a look at the CI run, this seems to be fixed. Might be able to test my changes out after this is merged into main @juliev0 

`verifying 9 Pods running with label selector "numaflow.numaproj.io/pipeline-name=test-pipeline-rollout-1,app.kubernetes.io/component=vertex"`

tests/e2e/functional-e2e/functional_test.go
```diff
@@ -273,7 +273,7 @@ var _ = Describe("Functional e2e:", Serial, func() {
 		numPipelineVertices := len(updatedPipelineSpec.Vertices)
 		UpdatePipelineRollout(pipelineRolloutName, updatedPipelineSpec, numaflowv1.PipelinePhaseRunning, func(retrievedPipelineSpec numaflowv1.PipelineSpec) bool {
 			return len(retrievedPipelineSpec.Vertices) == numPipelineVertices
-		}, true, true)
```
juliev0, line 276: So, it's because we set `overrideSourceVertexReplicas` to true here that we need to set `scaleInVertex` to true down below, right?

tests/e2e/numaflowcontroller.go
juliev0: let's hold off until Chandan separates the test out - otherwise, I'm worried that our CI will get even longer

juliev0 (approved):

Issue #576, opened 2025-02-18
Configurable resource health assessment window


juliev0:
# Summary

Each PipelineRollout or MonoVertexRollout should be able to override the default `ChildStatusAssessmentSchedule`.

We might as well also have 2 different default schedules: one for Pipeline and one for MonoVertex. 

Note [this](https://github.com/numaproj/numaplane/issues/572) issue is related (remove assessment from ISBService).

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
closed by: https://github.com/numaproj/numaplane/pull/593

PR #593, opened 2025-02-23
feat: ISBService doesn't need to delay before making assessment, like Pipeline and Monovertex do

### LINKED ISSUE #572: Progressive rollout: Consider removing "Assessment time window" code from InterstepBufferService assessment
# Summary

We have this notion of an Assessment time window for all of the Rollouts, but really it's only needed for Pipeline and MonoVertex. This is actually making it so the assessment of InterstepBufferService is taking longer. (It starts its window once it gets the first non-Unknown result, meaning that the Pipelines have already finished their time window.) However, I don't think we need that. 

---

juliev0:
Fixes #572 #576 

### Modifications

Separated the single `childAssessmentSchedule` from numaplane-controller-config into one per "Kind". (This is a string with comma separated values for "delay", "period", and "interval" to check.)

I also enabled the PipelineRollout, MonoVertexRollout, and ISBServiceRollout to optionally specify it. 

The default for ISBServiceRollout is now "0,0,10" - it still needs to have the "interval" set for when its result is "Unknown", since it needs to requeue to try again.

### Verification

I did success and failure case updates to each of these 3 Rollouts using the default rollout schedule. I observed that the ISBServiceRollout used delay=0 and period=0 so as to not delay its measurement and instead derive the result right after the PipelineRollout had derived its result (note that the success of the ISBServiceRollout is purely dependent on whether the PipelineRollouts were successful).

None.


config/install.yaml
```diff
@@ -1923,13 +1971,11 @@ subjects:
...
-    includedResources: "group=apps,kind=Deployment;\
-    group=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=Service;\
-    group=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role"
-    defaultUpgradeStrategy: "pause-and-drain"
-    childStatusAssessmentSchedule: "120,60,10"
+  config.yaml: "logLevel: INFO # Supported log levels are: VERBOSE, DEBUG, INFO, WARN.
+    ERROR, FATAL will be printed regardless of the log level\nincludedResources: \"group=apps,kind=Deployment;\\\ngroup=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=Service;\\\ngroup=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role\"\ndefaultUpgradeStrategy:
+    \"pause-and-drain\"\nprogressiveConfig:\n  defaultAssessmentSchedule: \n    -
+    kind: Pipeline\n      schedule: \"120,60,10\"\n    - kind: MonoVertex\n      schedule:
+    \"120,60,10\"\n    - kind: InterstepBufferService\n      schedule: \"0,0,10\"\n"
```
juliev0, line 1978: I'm not really sure why some of the line breaks got replaced with "\n" but I guess codegen did that

afugazzotto (approved):

PR #594, opened 2025-02-24
chore: incorporate Numaflow 1.4.3 RC3

juliev0:
### Modifications

This updates the e2e test to use the new Numaflow 1.4.3 RC 3 (and a clone of that which was named 1.4.3 RC33 even though it's not really 33). 

### Verification

Just ran the e2e test a few times.

N/A


tests/manifests/special-cases/no-strategy/controller-config.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: numaplane-controller-config
 data:
   config.yaml: |
-    logLevel: INFO # Supported log levels are: VERBOSE, DEBUG, INFO, WARN. ERROR, FATAL will be printed regardless of the log level
+    logLevel: DEBUG # Supported log levels are: VERBOSE, DEBUG, INFO, WARN. ERROR, FATAL will be printed regardless of the log level
```
juliev0, line 7: I'm not sure why we had this set to INFO - I'm guessing we generally didn't look at this test

tests/manifests/default/kustomization.yaml
```diff
@@ -5,6 +5,8 @@ resources:
   - ../../../config/default
   - ./controller_def_1.4.1.yaml
   - ./controller_def_1.4.2.yaml
+  - ./controller_def_1.4.3-rc3.yaml
+  - ./controller_def_1.4.3-rc33.yaml
```
juliev0, line 9: I didn't end up removing the previous versions - we can remove them later.

tests/manifests/special-cases/no-strategy/controller-config.yaml
chandankumar4: Probably, As it was set as `3` earlier, So respectively replaced with `INFO`

afugazzotto (approved):

PR #601, opened 2025-02-25
feat: add "promote full" option for Numaplane rollouts

### LINKED ISSUE #573: "Promote Full" option for Numaplane
# Summary

Argo Rollouts has an option called "Promote full" which means to bypass the standard analysis check and simply promote the new one. Perhaps we could have this option for a Rollout that's either Failed Progressive or has not yet been assessed (i.e. in either case still in "Pending" phase).

We could use a Label on the Rollout called "numaplane.numaproj.io/promote". Either we set it to "true", or it's perhaps a Label with no value, and if the Label exists it means to promote.

If we make it so that that label can be used for all progressive rollouts, then we can just use it in the progressive.go code directly without worrying about which Rollout it is.

---

dpadhiar:
Fixes #573 

This PR introduces the ability to `full-promote` a Numaplane rollout upgrading with the Progressive strategy which is to force it to promote to a new child and bypass all the previous checks. 

To use this new feature, the user adds the label `"numaplane.numaproj.io/promote"` set to `true` to an upgrading rollout that is either still pending or already assessed as a Failure for the update. This will bypass all checks and immediately update to the new child. We will know if a rollout was updated this way by checking the `upgradingChildStatus` which now has a boolean `forcedSuccess` which will be set to `true`.

This PR also updates the unit tests for Progressive strategy to create child objects in Kubernetes to effectively test that we are patching the child resource for success cases.

### Modifications

When the label is set on the rollout, we immediately skip all assessment logic and go straight to the `declareSuccess` function which now houses all of our success logic for updating the promoted and upgrading children.

### Verification

Testing upgrading a failed and pending rollout locally by adding the label and seeing that it completes.

Updating and running unit tests.

N/A


internal/controller/common/rollout_controller.go
```diff
@@ -170,6 +170,9 @@ func UpdateUpgradeState(ctx context.Context, c client.Client, upgradeState commo
 		childObject.GetKind(), childObject.GetNamespace(), childObject.GetName())
 	var patchJson string
 	labels := childObject.GetLabels()
+	if labels == nil {
```
dpadhiar, line 173: This is probably unnecessary but was unsure if this would help with the unit test failures.

juliev0: this seems good just in case

internal/controller/progressive/progressive.go
```diff
@@ -390,6 +376,39 @@ func AssessUpgradingPipelineType(ctx context.Context, existingUpgradingChildDef
...
+		return false, false, 0, err
+	}
+
+	rolloutObject.GetRolloutStatus().MarkProgressiveUpgradeSucceeded(fmt.Sprintf("New Child Object %s/%s Running", existingUpgradingChildDef.GetNamespace(), existingUpgradingChildDef.GetName()), rolloutObject.GetRolloutObjectMeta().Generation)
+	childStatus.AssessmentResult = apiv1.AssessmentResultSuccess
+	rolloutObject.SetUpgradingChildStatus(childStatus)
+	rolloutObject.GetRolloutStatus().MarkDeployed(rolloutObject.GetRolloutObjectMeta().Generation)
+
+	// we're done
+	return true, false, assessmentInterval, nil
```
juliev0, line 408: nit: let's reduce the output to just what we need and just return `done (bool)` and `error`. The caller can use that information to do the right thing

internal/controller/progressive/progressive.go
```diff
@@ -390,6 +384,54 @@ func AssessUpgradingPipelineType(ctx context.Context, existingUpgradingChildDef
 	return apiv1.AssessmentResultUnknown, nil
 }
 
+/*
+declareSuccess handles the success logic for the update of a child resource during a progressive upgrade.
+It patches both the existing and upgrading children, updates the child status and marks the rollout as deployed.
+
+Parameters:
+- ctx: The context for managing request-scoped values, cancellation, and timeouts.
+- rolloutObject: The current rollout object (this could be from cache).
```
juliev0, line 393: actually, let's not say "this could be from cache", since the PR I created today makes it so that it's not

internal/controller/progressive/progressive.go
```diff
@@ -390,6 +384,54 @@ func AssessUpgradingPipelineType(ctx context.Context, existingUpgradingChildDef
...
+- existingPromotedChildDef: The definition of the currently promoted child resource.
+- existingUpgradingChildDef: The definition of the child resource currently being upgraded.
+- childStatus: The upgrading status of the child resource currently being upgraded.
```
juliev0, line 396:
```suggestion
- childStatus: The status of the child resource currently being upgraded (from the Rollout CR)
```

internal/controller/progressive/progressive.go
```diff
@@ -245,6 +245,16 @@ func processUpgradingChild(
...
 	childStatus = rolloutObject.GetUpgradingChildStatus()
 
+	// check for Promote label here to force success logic
+	if rolloutObject.GetRolloutObjectMeta().Labels[common.LabelKeyNumaplanePromote] == "true" {
+		childStatus.ForcedSuccess = true
+		done, err := declareSuccess(ctx, rolloutObject, existingPromotedChildDef, existingUpgradingChildDef, childStatus, c)
+		if err != nil {
+			return done, false, 0, err
+		}
+		return done, false, assessmentSchedule.Interval, err
```
juliev0, line 255: for clarity, we may want to return 0 for the third parameter whenever `done` is true (same comment below), even though I realize that it's a little more code to do so

internal/common/common.go
```diff
@@ -91,6 +91,9 @@ const (
 	// LabelKeyUpgradeStateReason is an optional label to provide more information on top of the LabelKeyUpgradeState
 	LabelKeyUpgradeStateReason = "numaplane.numaproj.io/upgrade-state-reason"
 
+	// LabelKeyNumaplanePromote is the label value indicating that the resource managed by a NumaRollout shoudl be force promoted
```
juliev0, line 94:
```suggestion
	// LabelKeyNumaplanePromote is the label value indicating that the resource managed by a NumaRollout should be force promoted
```

juliev0 (approved):

PR #605, opened 2025-02-26
fix: increased number of pods for the source vertex for all strategies and other e2e tests fixes

afugazzotto:
### Modifications

Fixed E2E tests by:
- removing time.sleep lines that are executed at the beginning of the test suite instead of after each `It` node
- adding more source vertex pods
- increasing `PauseGracePeriodSeconds` since pausing the pipeline was not draining it before the timeout would occur

### Verification

E2E tests successful execution.

None.

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -135,8 +135,7 @@ func verifyVerticesPodsRunning(namespace, pipelineName string, specVertices []nu
 		)
 
 		for _, vtx := range specVertices {
-			// TTODO
-			// vtxLabelSelector := fmt.Sprintf("%s=%s", numaflowv1.KeyVertexName, vtx.Name) // this only works for pipeline
+			// vtxLabelSelector := fmt.Sprintf("%s=%s", numaflowv1.KeyVertexName, vtx.Name) // this only works for pipeline (not monovertex)
 			vtxLabelSelector := fmt.Sprintf("app.kubernetes.io/name=%s-%s", pipelineName, vtx.Name) // this should work also for monovertex
 			labelSelector := fmt.Sprintf("%s,%s", baseLabelSelector, vtxLabelSelector)
 
@@ -152,7 +151,7 @@ func verifyVerticesPodsRunning(namespace, pipelineName string, specVertices []nu
 
 			podsList, err := kubeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 			if err != nil {
-				panic(err) // TTODO: panic or something else?
+				return false
 			}
 
 			if podsList == nil {
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/utils/ptr"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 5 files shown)


tests/e2e/monovertex.go
```diff
@@ -85,7 +85,7 @@ func VerifyMonoVertexReady(namespace, monoVertexRolloutName string) {
 	daemonLabelSelector := fmt.Sprintf("%s=%s,%s=%s", numaflowv1.KeyMonoVertexName, monoVertexName, numaflowv1.KeyComponent, "mono-vertex-daemon")
 
 	By("Verifying that the MonoVertex is ready")
-	verifyPodsRunning(namespace, GetSourceVertexScaleMin(), vertexLabelSelector)
+	verifyPodsRunning(namespace, SourceVertexScaleMin, vertexLabelSelector)
```
juliev0, line 88: looks like this function is still assuming there are 5?

tests/e2e/common.go
```diff
@@ -91,6 +92,8 @@ const (
 	UpgradeStateLabelSelector = "numaplane.numaproj.io/upgrade-state=promoted"
 
 	LogSpacer = "================================"
+
+	SourceVertexScaleMin = 5
```
juliev0, line 96: do we need to have this as a constant across all tests and all pipelines/monovertices? It seems specific to a test.

tests/e2e/common.go
```diff
@@ -127,11 +121,59 @@ func verifyPodsRunning(namespace string, numPods int, labelSelector string) {
...
+
+			for _, pod := range podsList.Items {
+				if pod.Status.Phase != "Running" {
+					return false
+				}
+			}
+		}
+
+		return true
+	}).WithTimeout(TestTimeout).Should(BeTrue())
```
juliev0, line 174: nice

tests/e2e/common.go
juliev0: Any reason we can't set it where we set `sourceVertexScaleMax`?

tests/e2e/functional-e2e/functional_test.go
```diff
@@ -207,6 +210,9 @@ func TestFunctionalE2E(t *testing.T) {
 }
 
 var _ = Describe("Functional e2e:", Serial, func() {
+	// TODO: try to elimiate these assignments
```
juliev0, line 213:
```suggestion
	// TODO: try to eliminate these assignments
```

juliev0: So, they can't be set in the `It()` blocks?

tests/e2e/monovertex.go
afugazzotto: I understood we wanted to do this only for pipeline for now to solve the issue with the bug. I'll make the changes to apply it to MonoVertex as well.

tests/e2e/pipeline.go
```diff
@@ -84,7 +84,8 @@ func VerifyPipelineRolloutHealthy(pipelineRolloutName string) {
 	}).Should(Equal(metav1.ConditionTrue))
 }
 
-func VerifyPipelineRunning(namespace string, pipelineRolloutName string) {
+// TODO: remove tmpNumaflowBugOverride once the Numaflow bug (scale config not applied to replicas field) is fixed
+func VerifyPipelineRunning(namespace string, pipelineRolloutName string, tmpNumaflowBugOverride ...bool) {
```
juliev0, line 88: seems like `UpdatePipelineRollout()` isn't calling this for some reason - it should, right?

tests/e2e/common.go
afugazzotto: Once I apply the changes to MonoVertex, this can be removed and set in the functional and ppnd files directly.

tests/e2e/common.go
juliev0: so, it's not being called for monovertex yet?

tests/e2e/functional-e2e/functional_test.go
afugazzotto: They can be set in the `It` blocks. However, it seems unnecessary to even do that since they are set to their initial values. That is why I propose to remove them. I need to look further into why removing them fails the tests though.

tests/e2e/common.go
juliev0: thanks - if you prefer to do anything as multiple PRs, it's okay

tests/e2e/pipeline.go
afugazzotto: Not sure why it wasn't there. I added this case in the `expectedFinalPhase` switch:
`case numaflowv1.PipelinePhaseRunning:
    VerifyPipelineRunning(Namespace, name)`

tests/e2e/monovertex.go
```diff
@@ -73,21 +73,31 @@ func VerifyMonoVertexRolloutReady(monoVertexRolloutName string) {
...
 
 	By("Verifying that the MonoVertex is running")
 	monoVertexName := VerifyMonoVertexStatus(namespace, monoVertexRolloutName,
 		func(retrievedMonoVertexSpec numaflowv1.MonoVertexSpec, retrievedMonoVertexStatus kubernetes.GenericStatus) bool {
 			return retrievedMonoVertexStatus.Phase == string(numaflowv1.MonoVertexPhaseRunning)
 		})
 
-	vertexLabelSelector := fmt.Sprintf("%s=%s,%s=%s", numaflowv1.KeyMonoVertexName, monoVertexName, numaflowv1.KeyComponent, "mono-vertex")
-	daemonLabelSelector := fmt.Sprintf("%s=%s,%s=%s", numaflowv1.KeyMonoVertexName, monoVertexName, numaflowv1.KeyComponent, "mono-vertex-daemon")
+	unstructMonoVertex, err := GetMonoVertex(namespace, monoVertexRolloutName)
```
juliev0, line 84: We may want to add a comment to this function that says when calling this we should already have verified that the "promoted" one has been updated before getting here. The thing is that calling `GetMonoVertex()` basically finds the "promoted" one. So, if we haven't updated the "promoted" one yet, then that would find the original one.

Ultimately, it may be better to use the spec from the Rollout rather than the spec from the child in all of the functions and try to remove all of these calls to `GetMonoVertex()`, `GetPipeline()`, etc, but I'm fine if we create an issue for someone to do later.

juliev0: (I know the existing code already had this problem btw)

juliev0 (approved):

tests/e2e/monovertex.go
juliev0: Added [issue](https://github.com/numaproj/numaplane/issues/612) - we can do it at some point

PR #606, opened 2025-02-27
fix: get just the live rollout resource when reconciling to avoid issues caused by getting cached status

### LINKED ISSUE #602: Check for possible race condition wrt failure to update RolloutStatus
**Describe the bug**

Currently, in the Success case, we promote the "upgrading" child and mark the old one as "recyclable" before we update the Rollout Status. 

So, there's a possibility that we could promote the "upgrading" child and mark the old one as "recyclable" and then attempt to update the `UpgradingChildStatus` as Successful but fail doing so. Next reconciliation, we would [exit](https://github.com/numaproj/numaplane/blob/v0.14.0/internal/controller/progressive/progressive.go#L125) the progressive strategy. Need to understand the consequences of that. 

---

juliev0:
Mostly Fixes #602

### Modifications

Eventually, I'd like to have a local cache to store the Rollout Status, since we are the only ones writing to it. 

For now, however, I have updated the code to get the live Rollout every time to prevent cases like that identified by #602. 

I ended up doing this for the PipelineRollout, ISBServiceRollout, and MonoVertexRollout, since they are involved in the Progressive rollout case, in which we need to make sure we have the latest and greatest `Status` since we use that information that we previously wrote to inform what we do next.

Thus, I didn't end up changing NumaflowControllerRollout and NumaflowController because I don't believe we do anything like that for those. 

### Verification

I mostly no longer see the resource version conflict errors for these Rollouts in the CI logs. (I occasionally see one, but I think it's due to the test updating the Rollout in the middle of a reconciliation. That can still occur.)

N/A

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_controller.go
@@ -166,16 +166,8 @@ func (r *PipelineRolloutReconciler) processPipelineRollout(ctx context.Context, 
 	ctx = logger.WithLogger(ctx, numaLogger)
 	r.customMetrics.PipelineROSyncs.WithLabelValues().Inc()
 
-	// Get PipelineRollout CR
-	/*pipelineRollout := &apiv1.PipelineRollout{}
-	if err := r.client.Get(ctx, namespacedName, pipelineRollout); err != nil {
-		if apierrors.IsNotFound(err) {
-			return ctrl.Result{}, nil
-		} else {
-			r.ErrorHandler(pipelineRollout, err, "GetPipelineRolloutFailed", "Failed to get PipelineRollout")
-			return ctrl.Result{}, err
-		}
-	}*/
+	// Get the live PipelineRollout since we need latest Status for Progressive rollout case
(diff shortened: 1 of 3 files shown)


internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -236,13 +232,13 @@ func (r *PipelineRolloutReconciler) processPipelineRollout(ctx context.Context,
...
+	r.recorder.Eventf(pipelineRollout, "Normal", "ReconcileSuccess", "Reconciliation successful")
+	numaLogger.Debug("reconciliation successful")
+
 	if requeueDelay > 0 {
 		return ctrl.Result{RequeueAfter: requeueDelay}, nil
 	}
 
-	r.recorder.Eventf(pipelineRollout, "Normal", "ReconcileSuccess", "Reconciliation successful")
-	numaLogger.Debug("reconciliation successful")
-
```
juliev0, line 245: this wasn't getting logged before in the case of requeue

afugazzotto (approved):

PR #616, opened 2025-03-01
chore: add feature flag to disallow progressive upgrades for non-Monovertex resources

### LINKED ISSUE #595: Be able to use PPND for everything except MonoVertex initially
# Summary

Derek prefers not to turn on Progressive for Pipeline right away.

Need to decide how we want to do this

---

afugazzotto:
Fixes #595 

### Modifications

Added a feature flag to disallow progressive upgrades for non-MonoVertex resources.

### Verification

Manually tested the following:
Deployed the sample NumaflowControllerRollout, ISBSvcRollout, and the sample namespace-level ConfigMap with `upgradeStrategy: "progressive"`. Then,
- For MonoVertex:
  1. created the sample MonoVertexRollout
  2. observed the `upgradeInProgress` field of its status
  3. updated the MonoVertexRollout by changing the image
  4. observed that the `upgradeInProgress` field of its status changed to `Progressive`
- For Pipeline:
  1. created the sample PipelineRollout
  2. observed the `upgradeInProgress` field of its status
  3. updated the PipelineRollout by changing the vertices topology
  4. observed that the `upgradeInProgress` field of its status changed to `PipelinePauseAndDrain` instead of `Progressive`

None.

### DIFF
--- a/tests/manifests/special-cases/no-strategy/controller-config.yaml
+++ b/tests/manifests/special-cases/no-strategy/controller-config.yaml
@@ -9,6 +9,7 @@ data:
     group=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=Service;\
     group=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role"
     defaultUpgradeStrategy: "no-strategy"
+    featureFlagDisallowProgressiveForNonMonoVertex: true
     progressive:
       defaultAssessmentSchedule: 
         - kind: Pipeline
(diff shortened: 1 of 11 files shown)


tests/manifests/special-cases/no-strategy/controller-config.yaml
```diff
@@ -9,11 +9,12 @@ data:
     group=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=Service;\
     group=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role"
     defaultUpgradeStrategy: "no-strategy"
+    featureFlagDisallowProgressiveForNonMonoVertex: true
```
afugazzotto, line 12: Decided to only add the feature flag here for manual local testing and not in the other config files because:
- it is a temporary feature flag that will be removed
- numa-manifest-generator will take care of setting it
- avoid breaking unit and E2E tests which we should not change since this is a temporary feature flag

juliev0: I'm confused about why put it here at all?

afugazzotto: I did it for manual tests, I guess I can remove it now.

juliev0 (approved):

PR #634, opened 2025-03-07
feat: Scale down failed MonoVertex to 0 Pods 

juliev0:
### Modifications

This PR takes care of MonoVertex only - Pipeline can be done later. 

It does the following:

- Upon failure, scales the MonoVertex down to 0 Pods
- If "force promote" is done, scale it back up to the value specified in the MonoVertexRollout

(Note that for the latter, we may ultimately want to not get it from the MonoVertexRollout but instead save the `scaleDefinition` in the UpgradingChildStatus just like we do right now for Promoted child. Just thinking of very rare edge cases in which someone just changed the scale in the MonoVertexRollout and we use that instead of the one we were using. Normally, it would be fine to just use that scale value, but thinking of some rare case in which it's important that the scale value gets set atomically with some other change....) 

### Verification

Ran various sequences of failure and success and observed scale set correctly.
Tested the following with a monovertex whose scale.min and scale.max were set and one for which `scale` was not set at all:

- Updated good spec to bad spec - observed good one scaled down to half, bad one came up as normal at full scale, after failure bad one scaled down to 0 and good one scaled back up
- Updated bad spec to use "forcePromote": "true" - observed bad one got promoted, and it scaled back up (spec specified full scale but I think Numaflow wouldn't bring up all the pods because they were unhealthy (not sure if a Numaflow bug or not))
- Updated good spec to good spec - observed original scaled down to half, good one came up as normal at full scale

### DIFF
--- a/internal/controller/monovertexrollout/monovertexrollout_controller.go
+++ b/internal/controller/monovertexrollout/monovertexrollout_controller.go
@@ -608,6 +608,14 @@ func (r *MonoVertexRolloutReconciler) ChildNeedsUpdating(ctx context.Context, fr
 
 		excludedPaths := []string{"replicas", "scale.min", "scale.max"}
 		util.RemovePaths(specAsMap, excludedPaths, ".")
+
+		// if "scale" is there and empty, remove it
+		// (this enables accurate comparison between one monovertex with "scale" empty and one with "scale" not present)
+		scaleMap, found := specAsMap["scale"].(map[string]interface{})
+		if found && len(scaleMap) == 0 {
+			unstructured.RemoveNestedField(specAsMap, "scale")
+		}
+
 		return specAsMap, nil
 	}
 
--- a/internal/controller/monovertexrollout/monovertexrollout_controller.go
+++ b/internal/controller/monovertexrollout/monovertexrollout_controller.go
@@ -43,7 +43,6 @@ import (
 	numaflowv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaplane/internal/common"
 	ctlrcommon "github.com/numaproj/numaplane/internal/controller/common"
-	"github.com/numaproj/numaplane/internal/controller/common/numaflowtypes"
 	"github.com/numaproj/numaplane/internal/controller/progressive"
 	"github.com/numaproj/numaplane/internal/usde"
 	"github.com/numaproj/numaplane/internal/util"
@@ -594,16 +593,29 @@ func getBaseMonoVertexMetadata(monoVertexRollout *apiv1.MonoVertexRollout) (apiv
 
 }
 
-// ChildNeedsUpdating() tests for essential equality, with any irrelevant fields eliminated from the comparison
+// ChildNeedsUpdating() tests for essential equality, with any fields that Numaplane manipulates eliminated from the comparison
 // This implements a function of the progressiveController interface
 func (r *MonoVertexRolloutReconciler) ChildNeedsUpdating(ctx context.Context, from, to *unstructured.Unstructured) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
-	// remove "replicas" field from comparison to test for equality
-	fromNew, err := numaflowtypes.MonoVertexWithoutReplicas(from)
+
+	// remove certain fields (which numaplane needs to set) from comparison to test for equality
+	removeFunc := func(monoVertex *unstructured.Unstructured) (map[string]interface{}, error) {
+		var specAsMap map[string]any
+
+		if err := util.StructToStruct(monoVertex.Object["spec"], &specAsMap); err != nil {
+			return nil, err
+		}
+
+		excludedPaths := []string{"replicas", "scale.min", "scale.max"}
+		util.RemovePaths(specAsMap, excludedPaths, ".")
+		return specAsMap, nil
+	}
+
+	fromNew, err := removeFunc(from)
 	if err != nil {
 		return false, err
 	}
-	toNew, err := numaflowtypes.MonoVertexWithoutReplicas(to)
+	toNew, err := removeFunc(to)
 	if err != nil {
 		return false, err
 	}
--- a/internal/controller/monovertexrollout/monovertexrollout_controller_test.go
+++ b/internal/controller/monovertexrollout/monovertexrollout_controller_test.go
@@ -575,3 +575,61 @@ func TestChildNeedsUpdating(t *testing.T) {
 		})
 	}
 }
+
+func TestGetScaleValuesFromMonoVertexSpec(t *testing.T) {
+	one := int64(1)
+	ten := int64(10)
+	tests := []struct {
+		name        string
+		input       map[string]interface{}
+		expectedMin *int64
+		expectedMax *int64
+		expectError bool
+	}{
+		{
+			name: "BothValuesPresent",
+			input: map[string]interface{}{
+				"scale": map[string]interface{}{
+					"min": int64(1),
+					"max": int64(10),
+				},
+			},
+			expectedMin: &one,
+			expectedMax: &ten,
+			expectError: false,
+		},
+		{
+			name: "NoValuesPresent",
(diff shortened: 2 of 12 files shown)


internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -594,16 +593,29 @@ func getBaseMonoVertexMetadata(monoVertexRollout *apiv1.MonoVertexRollout) (apiv
...
+
+	// remove certain fields (which numaplane needs to set) from comparison to test for equality
+	removeFunc := func(monoVertex *unstructured.Unstructured) (map[string]interface{}, error) {
+		var specAsMap map[string]any
+
+		if err := util.StructToStruct(monoVertex.Object["spec"], &specAsMap); err != nil {
+			return nil, err
+		}
+
+		excludedPaths := []string{"replicas", "scale.min", "scale.max"}
```
afugazzotto, line 609: Same comment made for pipeline.

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
```diff
@@ -575,3 +575,61 @@ func TestChildNeedsUpdating(t *testing.T) {
...
+				"scale": map[string]interface{}{
+					"min": int64(1),
+					"max": int64(10),
+				},
+			},
+			expectedMin: &one,
+			expectedMax: &ten,
+			expectError: false,
+		},
+		{
```
afugazzotto, line 601: Should we add cases for only one of them being set?

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -91,6 +92,127 @@ func (r *MonoVertexRolloutReconciler) ProcessPromotedChildPreUpgrade(
...
+
+	numaLogger := logger.FromContext(ctx).WithName("ProcessUpgradingChildPostFailure").WithName("MonoVertexRollout")
+
+	numaLogger.Debug("started post-failure processing of upgrading monovertex")
+
+	// scale down monovertex to 0 Pods
+	// need to check to see if it's already scaled down before we do this
+	existingSpec, ok := upgradingChildDef.Object["spec"].(map[string]interface{})
+	if !ok {
+		numaLogger.Errorf(errors.New("monovertex spec invalid"), "existing upgrading monovertex spec doesn't start with root 'spec' key (or is of invalid type)?: %+v", upgradingChildDef.Object)
```
afugazzotto, line 125: Is this necessary. At this point in the code we should know that the upgradingChildDef has a valid spec. Maybe we should move this verification at the beginning of the reconciliation process where upgradingChildDef if first retrieved.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -91,6 +92,127 @@ func (r *MonoVertexRolloutReconciler) ProcessPromotedChildPreUpgrade(
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getScaleValuesFromMonoVertexSpec(monovertexSpec map[string]interface{}) (*int64, *int64, error) {
+	min, foundMin, err := unstructured.NestedInt64(monovertexSpec, "scale", "min")
+	if err != nil {
+		// try again using Float64
```
afugazzotto, line 188: Why do you try using Float64? Min and max are int32 in Numaflow.

afugazzotto (commented):
Added some more comments after reviewing over zoom. These new comments can be addressed with a separate PR.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: just being complete - do you think it hurts to have it?

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: For some reason, I was getting an error some of the time when I had it set to `NestedFloat64()` that it needed to be Int64 and other times when I had it set to `NestedInt64()` that it needed to be Float64. In once case, it comes from MonoVertexRollout spec and in the other case comes from Unstructured. Numaplane doesn't know what Numaflow uses.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: It does not hurt to have it, but I'm thinking we do not check that before using the definition at other locations so it may be actually better to check for the spec immediately after getting the definition. I actually like this idea, I just think we should place it at the beginning.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: We can keep it and investigate more later on why that happens.

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -935,15 +935,17 @@ func (r *PipelineRolloutReconciler) drain(ctx context.Context, pipeline *unstruc
...
-// ChildNeedsUpdating() tests for essential equality, with any irrelevant fields eliminated from the comparison
+// ChildNeedsUpdating() tests for essential equality, with any fields that Numaplane manipulates eliminated from the comparison
 // This implements a function of the progressiveController interface
 func (r *PipelineRolloutReconciler) ChildNeedsUpdating(ctx context.Context, from, to *unstructured.Unstructured) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
 	fromCopy := from.DeepCopy()
 	toCopy := to.DeepCopy()
 	// remove lifecycle.desiredPhase field from comparison to test for equality
 	numaflowtypes.PipelineWithoutDesiredPhase(fromCopy)
+	numaflowtypes.PipelineWithoutScaleMinMax(fromCopy)
```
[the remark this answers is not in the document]
juliev0: So, for the progressive case this is called only in the case of our being in the Failed state and checking to see if there's a new update. I guess that means that when we're in the Failed state, we would only perform an update if they change a field other than `desiredPhase` and other than `scale.min` and `scale.max`. That means that only when done with Progressive would they process those. 

The reason I had to change that was this: if we update scale.min and scale.max to 0 on failure, then on the next reconciliation we will think there's a change, and create a new "upgrading" child.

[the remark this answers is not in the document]
afugazzotto: That make sense, thank you for explaining.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: if you think it's too messy, I'd be okay to take it out I guess, though.

afugazzotto: This can be done later in a separate PR.

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
juliev0: you're right - I actually just let Qodo write the test, and I didn't look at it closely - going too fast...

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: I ended up taking it out and not putting it at the beginning of the reconciliation loop at all. I don't think we're realistically going to have a case like that, do you? I think it's probably just clutter which will be an extra code coverage line we're not covering. :)

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
juliev0: added

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: Sounds good

afugazzotto (approved):

PR #640, opened 2025-03-10
chore: during progressive upgrade only store min and max scale values

afugazzotto:
### Modifications

During progressive upgrade only store min and max for promoted scale values to restore in case of failure

### Verification

Manually verified that while upgrading the MonoVertex/Pipeline, the stored value only has min and max and that restoring it only overwrites min and max (not other scale fields).
Tested the following (for both MonoVertex and Pipeline):
- both min and max set (and zeroReplicaSleepSeconds)
- only min set (and zeroReplicaSleepSeconds)
- scale unset (nil)

None


internal/controller/progressive/progressive.go
```diff
@@ -421,9 +421,9 @@ Returns:
 - A boolean indicating if the upgrade is done.
 - An error if any issues occur during the process.
 */
-func declareSuccess(ctx context.Context,
+func declareSuccess(
+	ctx context.Context,
 	rolloutObject ProgressiveRolloutObject,
-	controller progressiveController,
```
afugazzotto, line 426: This was unused

internal/controller/common/in_progress_strategy.go
```diff
@@ -119,6 +121,32 @@ func (store *inProgressStrategyStore) SetStrategy(namespacedName k8stypes.Namesp
...
+
+	jsonBytes, err := json.Marshal(originalScaleMinMaxOnly)
+	if err != nil {
+		return nil, err
+	}
+
+	jsonString := string(jsonBytes)
+	return &jsonString, nil
+}
+
```
juliev0, line 149: this file is only related to getting and setting in progress strategy. Since this is only related to progressive strategy, maybe put it there?

internal/controller/common/in_progress_strategy.go
```diff
@@ -119,6 +121,32 @@ func (store *inProgressStrategyStore) SetStrategy(namespacedName k8stypes.Namesp
...
+	}
+
+	if !foundScale {
+		return nil, nil
+	}
+
+	originalScaleMinMaxOnly := map[string]any{
+		"min": originalScaleDef["min"],
+		"max": originalScaleDef["max"],
+	}
```
juliev0, line 139: so, basically if "min" and "max" aren't set at all, then it's something like `{"min": 0, "max": 0 }`?

juliev0, line 139: Did you test the case of `scale` set but not `min` and `max`?

internal/controller/progressive/progressive.go
juliev0: thanks for the cleanup

internal/controller/common/in_progress_strategy.go
afugazzotto: Yes, I updated the description to add the manual tests I verified. However, this comment makes me think that the func `ExtractOriginalScaleMinMax` should return "null" string instead of nil in case scale is not set. I think it would be more appropriate to return `string` instead of `*string`.

internal/controller/progressive/progressive_test.go
```diff
@@ -470,3 +470,63 @@ func Test_getChildStatusAssessmentSchedule(t *testing.T) {
...
+			expectedJSONString: "{\"max\":7,\"min\":3}",
+			expectedError:      nil,
+		},
+		{
+			name:               "both min and max set and also another field",
+			object:             map[string]any{"scale": map[string]any{"min": int64(3), "max": int64(7), "zeroReplicaSleepSeconds": int64(123)}},
+			pathToScale:        []string{"scale"},
+			expectedJSONString: "{\"max\":7,\"min\":3}",
+			expectedError:      nil,
+		},
```
juliev0, line 516: nice tests

internal/controller/common/in_progress_strategy.go
juliev0: That's good - now I think you're essentially returning the content necessary for patching

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -343,13 +331,13 @@ func scaleDownPromotedMonoVertex(
...
 		"newMax", newMax,
-		"originalScaleDefAsString", originalScaleDefAsString,
+		"originalScaleMinMax", originalScaleMinMax,
 	).Debugf("found %d pod(s) for the monovertex, scaling down to %d", actualPodsCount, newMax)
 
 	scaleValuesMap[promotedChildDef.GetName()] = apiv1.ScaleValues{
-		OriginalScaleDefinition: originalScaleDefAsString,
-		ScaleTo:                 newMax,
-		Actual:                  actualPodsCount,
+		OriginalScaleMinMax: &originalScaleMinMax,
```
juliev0, line 338: nit: do you still want this to be a pointer to a string or do you want to change it to a simple string type?

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -394,16 +383,20 @@ func scalePipelineSourceVerticesToOriginalValues(
 				return true, fmt.Errorf("the scale values for vertex '%s' are not present in the rollout promotedChildStatus", vertexName)
 			}
 
-			if vertexScaleValues.OriginalScaleDefinition == nil {
+			if vertexScaleValues.OriginalScaleMinMax == "null" {
```
juliev0, line 386: Did you do any testing for pipeline? (such as this case)

afugazzotto: Yes, updated above description. I tested the same cases for both MonoVertex and Pipeline.

juliev0 (approved):

PR #642, opened 2025-03-12
feat: Watch AnalysisRuns that are owned by Pipelines/MonoVertices which are owned by Numa Rollouts

juliev0:
### Modifications

Add a Watch() for AnalysisRuns - but only include in our cache the ones which are owned by either:
- a MonoVertex which is owned by a MonoVertexRollout, or
- a Pipeline which is owned by a PipelineRollout

This will cause the PipelineRollout or the MonoVertexRollout to be enqueued (`Reconcile()` called).

### Verification

You can see the [commented out code](https://github.com/numaproj/numaplane/blob/watch-rollouts/internal/controller/monovertexrollout/monovertexrollout_progressive.go#L44-L75) in there where I create an `AnalysisRun` during progressive rollout for MonoVertex. I ran that to create the AnalysisRun and then added some temporary print statements [in the Watch() code](https://github.com/numaproj/numaplane/blob/watch-rollouts/internal/controller/monovertexrollout/monovertexrollout_controller.go#L431-L436) to confirm that it found the MonoVertexRollout and enqueued it. 

I also verified PipelineRollout Watch() command by creating an AnalysisRun manually with OwnerReference to Pipeline.

N/A


internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -40,6 +40,40 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
+					},
+				},
+			}
+			if err = r.client.Create(ctx, analysisRun); err != nil {
+				return apiv1.AssessmentResultUnknown, err
+			}
+		} else {
+			return apiv1.AssessmentResultUnknown, err
+		}
+	}*/
```
juliev0, line 75: Is it worth keeping this in here for now until we implement it (as a reference)?

tests/manifests/default/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 
 resources:
   - ../../../config/default
+  - https://github.com/argoproj/argo-rollouts/manifests/crds?ref=stable
   - ./controller_def_1.4.1.yaml
   - ./controller_def_1.4.2.yaml
   - ./controller_def_1.4.3-rc3.yaml
```
juliev0, line 9: deploy Argo Rollouts as part of our test

internal/controller/monovertexrollout/monovertexrollout_progressive.go
afugazzotto: I'm ok with keeping it here, but add a TODO with some details.

afugazzotto (approved):

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: will do, thanks

dpadhiar (approved):

PR #646, opened 2025-03-12
fix: update streamPodLogs func to not fail when pod isn't found

dpadhiar:
### Modifications

Quick fix to make not finding a pod while streaming logs no longer a fatal error. This pod can simply be skipped over and goroutine killed when this occurs.

Was the result of a flakey test which tried to stream a pod that was quickly deleted.

### Verification

Running tests on CI.

N/A

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -266,12 +266,12 @@ func streamPodLogs(ctx context.Context, client clientgo.Interface, namespace, po
 				return true, nil
 			}
 
-			fmt.Printf("Got error %v, retrying.\n", err)
 			return false, nil
 		})
 
 		if err != nil {
-			log.Fatalf("Failed to stream pod %q logs: %v", podName, err)
+			fmt.Printf("Failed to stream pod %q logs: %v", podName, err)
+			return
 		}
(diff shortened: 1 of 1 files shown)


tests/e2e/common.go
```diff
@@ -266,12 +265,12 @@ func streamPodLogs(ctx context.Context, client clientgo.Interface, namespace, po
 				return true, nil
 			}
 
-			fmt.Printf("Got error %v, retrying.\n", err)
 			return false, nil
 		})
 
 		if err != nil {
-			log.Fatalf("Failed to stream pod %q logs: %v", podName, err)
+			fmt.Printf("Failed to stream pod %q logs: %v", podName, err)
```
chandankumar4, line 272: Maybe we can use some logger library and log it as error, for example we are already using `logrus`/`zerolog` in this project.
But can be as an improvement in upcoming PR.

chandankumar4 (approved):

tests/e2e/common.go
dpadhiar: Thanks Chandan! Let me mark that as TODO.

juliev0 (approved):

PR #648, opened 2025-03-13
fix: scale upgrading MonoVertex to desired scale values once deemed successful for promotion

afugazzotto:
### Modifications

Scale the upgrading MonoVertex deemed successful for promotion to the Rollout scale values immediately after the assessment has been completed.

### Verification

https://drive.google.com/file/d/187BIbKxIyzDhk5Jw59Uf6quMuVErXzJy/view?usp=drive_link

None

### DIFF
--- a/internal/controller/monovertexrollout/monovertexrollout_progressive.go
+++ b/internal/controller/monovertexrollout/monovertexrollout_progressive.go
@@ -201,19 +201,37 @@ func (r *MonoVertexRolloutReconciler) ProcessUpgradingChildPostSuccess(
 	upgradingMonoVertexDef *unstructured.Unstructured,
 	c client.Client,
 ) error {
-	return scaleMonoVertexAsPerRolloutSpec(ctx, rolloutObject, upgradingMonoVertexDef, c)
-}
 
-/*
-ProcessUpgradingChildPreForcedPromotion handles an upgrading monovertex that's either failed or still being assessed.
-*/
-func (r *MonoVertexRolloutReconciler) ProcessUpgradingChildPreForcedPromotion(
-	ctx context.Context,
-	rolloutObject progressive.ProgressiveRolloutObject,
-	upgradingMonoVertexDef *unstructured.Unstructured,
-	c client.Client,
-) error {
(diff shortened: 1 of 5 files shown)


internal/controller/progressive/progressive.go
```diff
@@ -438,17 +441,23 @@ Returns:
...
 	rolloutObject ProgressiveRolloutObject,
+	controller progressiveController,
 	existingPromotedChildDef, existingUpgradingChildDef *unstructured.Unstructured,
 	childStatus *apiv1.UpgradingChildStatus,
 	c client.Client,
 ) (bool, error) {
 
 	numaLogger := logger.FromContext(ctx)
 
+	err := controller.ProcessUpgradingChildPostSuccess(ctx, rolloutObject, existingUpgradingChildDef, c)
```
afugazzotto, line 452: @juliev0 what is your opinion on performing the scaling before or after setting the promotion and recyclable labels?

juliev0: Thanks for asking. The way you have it is good. If we still have our "upgrading" child label, we'll be able to get back into this code if for some reason we fail at this call. 

(Also, given that we're using patch, we're far less likely to fail than if we were doing Update, so that's good - although a separate thing).

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -539,3 +540,47 @@ func scaleMonoVertex(
 	patchJson := fmt.Sprintf(`{"spec": {"scale": %s}}`, scaleValue)
 	return kubernetes.PatchResource(ctx, c, monovertex, patchJson, k8stypes.MergePatchType)
 }
+
+/*
+scaleMonoVertexAsPerRolloutSpec scales a MonoVertex according to the specified rollout scale min ad max.
```
juliev0, line 545:
```suggestion
scaleMonoVertexAsPerRolloutSpec scales a MonoVertex according to the specified rollout scale min and max.
```

juliev0 (approved):

PR #651, opened 2025-03-16
chore: Enable AnalysisTemplate specification for pipelines and monovertices

juliev0:
### Modifications

Enable the ability to specify AnalysisTemplates for Pipelines and MonoVertices (including both spec and status). After some deliberation, I made the decision not to include it for ISBServiceRollouts. I don't think we have a need for it, and I think it may be better to restrict the possibility of what users may do so we don't need to concern ourselves with that case.

(If it ever changes, and somebody does need it for ISBServiceRollout, then the types can easily be refactored while maintaining backward compatibility in the json/yaml serialization.)

### Verification

For the PipelineRollout, MonoVertexRollout, and ISBServiceRollout in `config/samples` directory, I applied them and then verified that the fields got set appropriately.

None. I maintained the original json/yaml serialization, i.e. this is the same before/after:
```
apiVersion: numaplane.numaproj.io/v1alpha1
kind: PipelineRollout
metadata:
  name: my-pipeline
  namespace: example-namespace
spec:
  strategy:
    progressive:
      assessmentSchedule: "60,60,10"
```


pkg/apis/numaplane/v1alpha1/shared_types.go
```diff
@@ -40,13 +44,25 @@ const (
...
 
-// PipelineTypeRolloutStrategy specifies the RolloutStrategy for fields shared by Pipeline and MonoVertex
-type PipelineTypeRolloutStrategy struct {
-	RolloutStrategy `json:",inline"`
+// PipelineTypeProgressiveStrategy specifies the Progressive Rollout Strategy for fields shared by Pipeline and MonoVertex
+type PipelineTypeProgressiveStrategy struct {
+	Progressive ProgressiveStrategy `json:",inline"`
+
+	Analysis Analysis `json:"analysis,omitempty"`
+}
```
juliev0, line 57: My main motivation for this refactoring was to separate the `Analysis` section from the `ProgressiveStrategy` such that PipelineRollout and MonoVertexRollout would include `Analysis` but ISBServiceRollout would not.

pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
```diff
@@ -62,7 +62,7 @@ type MonoVertexProgressiveStatus struct {
 
 // UpgradingMonoVertexStatus describes the status of an upgrading child
 type UpgradingMonoVertexStatus struct {
-	UpgradingChildStatus `json:",inline"`
+	UpgradingPipelineTypeStatus `json:",inline"`
```
afugazzotto, line 65: It's a bit confusing to have something called `UpgradingPipelineTypeStatus` inside the `UpgradingMonoVertexStatus`. Can you explain the relationship between them in a comment wherever appropriate?

juliev0: yeah, let me add a comment for that. Just FYI, "pipelineType" is what I use in various places to mean either "Pipeline" or "MonoVertex"

juliev0: added to both here and the same place in `UpgradingPipelineStatus` struct

juliev0: (oops, forgot I need to run `make codegen` after adding a comment, let me do that)

afugazzotto: That's why I suggested to add a comment, it is not obvious what is meant by "pipelineType" in the context of a MonoVertex. And I understand it is difficult to find a naming that fits both of them.

afugazzotto (approved):

PR #657, opened 2025-03-17
fix: update when pausing metrics are set to avoid edge case for quick pause showing as a longer pause

dpadhiar:
Fixes an internal issue.

### Modifications

We were seeing a case where a quick pause for updating the image of a PipelineRollout could cause a resource conflict error when updating the child and result in the pauseStatus of the rollout not being updated properly. This would cause the reported pause metrics to use the previously set status resulting in a much longer pause time than we should expect.

This is fixed by moving where we set the metric to avoid any erroneous values.

### Verification

Testing locally with a long pause and seeing that the metric updates properly.

N/A

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_controller.go
@@ -703,19 +703,25 @@ func (r *PipelineRolloutReconciler) setChildResourcesPauseCondition(pipelineRoll
 		// if BeginTime hasn't been set yet, we must have just started pausing - set it
 		if pipelineRollout.Status.PauseStatus.LastPauseBeginTime == metav1.NewTime(initTime) || !pipelineRollout.Status.PauseStatus.LastPauseBeginTime.After(pipelineRollout.Status.PauseStatus.LastPauseEndTime.Time) {
 			pipelineRollout.Status.PauseStatus.LastPauseBeginTime = metav1.NewTime(time.Now())
-			r.updatePauseMetric(pipelineRollout, pipelinePhase)
 		}
 		reason := fmt.Sprintf("Pipeline%s", string(pipelinePhase))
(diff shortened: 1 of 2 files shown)


internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -706,7 +706,10 @@ func (r *PipelineRolloutReconciler) setChildResourcesPauseCondition(pipelineRoll
 		}
 		reason := fmt.Sprintf("Pipeline%s", string(pipelinePhase))
 		msg := fmt.Sprintf("Pipeline %s", strings.ToLower(string(pipelinePhase)))
-		r.updatePauseMetric(pipelineRollout, pipelinePhase)
+		// if TransitionTime is still before or equal to BeginTime, keep updating Pausing metric
```
juliev0, line 709:
```suggestion
		// if TransitionTime is before or equal to BeginTime, update Pausing metric
		// if it's not, then something is wrong and we don't want to make a bad calculation
```

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -715,15 +718,19 @@ func (r *PipelineRolloutReconciler) setChildResourcesPauseCondition(pipelineRoll
 		}
 		reason := fmt.Sprintf("Pipeline%s", string(pipelinePhase))
 		msg := fmt.Sprintf("Pipeline %s", strings.ToLower(string(pipelinePhase)))
-		r.updatePauseMetric(pipelineRollout, pipelinePhase)
+		// if EndTime is still before or equal to Begintime, keep updating Paused metric
```
juliev0, line 721:
```suggestion
		// if EndTime is before or equal to BeginTime, update Paused metric
		// if it's not, then something is wrong, and we don't want to make a bad calculation
```

juliev0 (approved):

PR #658, opened 2025-03-18
chore: Run Argo Rollouts when we run Numaplane

juliev0:
### Modifications

Since Numaplane will now require Argo Rollouts, we can run Argo Rollouts whenever we run Numaplane.

This runs argo-rollouts in the "argo-rollouts" namespace, and in a "cluster mode", such that it reconcile from any namespace.

### Verification

standard CI

N/A

### DIFF
--- a/tests/manifests/default/kustomization.yaml
+++ b/tests/manifests/default/kustomization.yaml
@@ -3,7 +3,7 @@ kind: Kustomization
 
 resources:
   - ../../../config/default
-  - https://github.com/argoproj/argo-rollouts/manifests/crds?ref=stable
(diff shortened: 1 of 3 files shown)


tests/manifests/default/kustomization.yaml
```diff
@@ -3,7 +3,6 @@ kind: Kustomization
 
 resources:
   - ../../../config/default
-  - https://github.com/argoproj/argo-rollouts/manifests/crds?ref=stable
```
juliev0, line 6: no need to deploy the CRDs here anymore

afugazzotto (approved):

PR #663, opened 2025-03-21
chore: install Prometheus when running 

juliev0:
### Modifications

Install/upgrade prometheus if not installed. If user already has prometheus installed, they can skip it by setting PROMETHEUS_REQUIRED=false. This installs it to "prometheus" namespace.

### Verification

Ran both:
```
STRATEGY=progressive make start
```

and

```
PROMETHEUS_REQUIRED=false STRATEGY=progressive make start
```

Also ran this in the CI 3 times in a row successfully (since changing prometheus size to "small").

N/A


tests/manifests/default/prometheus-monitors.yaml
```diff
@@ -0,0 +1,58 @@
...
+  selector:
+    matchExpressions:
+    - key: app.kubernetes.io/part-of
+      operator: In
+      values:
+      - numaflow
+    - key: app.kubernetes.io/component
+      operator: In
+      values:
+      - mono-vertex
```
afugazzotto, line 52: Any reason we do not call this `monovertex` like in other places in the code?

afugazzotto (approved):

tests/manifests/default/prometheus-monitors.yaml
juliev0: numaflow is setting that, not us

PR #666, opened 2025-03-21
fix: Store the original scale min/max for the upgrading MonoVertex in the Status and use that on success

### LINKED ISSUE #637: Save the scale definition used for forcePromote in the upgradingChildStatus
**Describe the bug**

In [this](https://github.com/numaproj/numaplane/pull/634) PR, the MonoVertexRollout-defined `scale` definition is used when force promoting a failed (or upgrading) MonoVertex. Better would be to save the original monovertex's scale and use that, however, as mentioned in the PR.

---

juliev0:
Fixes #637 

### Modifications

The Main modification was to add a new field in the Upgrading Status called `ScaleMinMaxValues`. Note that contrary to the promoted child status, I did not feel that we needed the other fields, as we don't need to keep track of the "current" scale. 

I think we just need to be able to scale the "upgrading" child back up after Success. As long as we patch it correctly, that's sufficient.

I also made a couple of other small changes, described inline.

### Verification

I tried these tests:
1. update MonoVertexRollout from successful to successful (observe new status field gets set correctly; observe rescale up when new child gets promoted)
2. update MonoVertexRollout from successful to bad, and then "force promote"
3. update MonoVertexRollout from successful to successful and change scale at the same time
4. update MonoVertexRollout from successful to successful and remove "scale" field altogether (this sets value to "null")
5. update MonoVertexRollout from successful to successful and set "scale: {}" (this sets value to `{"max":null,"min":null}`)

I guess if there are any existing MonoVertexRollouts in a "Failed" state and somebody does a "Force Promote" on one, it would look for this new `ScaleMinMaxValues` field to determine what to scale to, which wouldn't exist. The code in that case sets it to "null". That's actually okay since once it's done with Progressive strategy, it will notice that the scale values are different from the desired ones, and then it would reset it.


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -425,7 +425,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 	case apiv1.UpgradeStrategyProgressive:
 		numaLogger.Debug("processing InterstepBufferService with Progressive")
 
-		done, _, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
+		done, progressiveRequeueDelay, err := progressive.ProcessResource(ctx, isbServiceRollout, existingISBServiceDef, isbServiceNeedsToUpdate, r, r.client)
```
juliev0, line 428: We were no longer using this value `newChild` from the call to `ProcessResource()` so I thought I'd remove it from that function altogether to clean things up.

internal/controller/progressive/progressive.go
```diff
@@ -469,7 +467,7 @@ func declareSuccess(
 }
 
 // return:
-// - whether we just created a new child
+// - the new child we created, if any
```
juliev0, line 470: this was forgotten when the value changed type in the past

internal/controller/progressive/progressive.go
```diff
@@ -494,13 +492,18 @@ func startUpgradeProcess(
...
+	}
 
 	requeue, err = controller.ProcessUpgradingChildPreUpgrade(ctx, rolloutObject, newUpgradingChildDef, c)
 	if err != nil {
-		return nil, false, err
+		return newUpgradingChildDef, false, err
 	}
 	if requeue {
-		return nil, true, nil
+		return newUpgradingChildDef, true, nil
```
juliev0, line 506: We were being inconsistent about returning this in the cases where it wasn't really being used on the other side. In these 2 cases we weren't returning it, but in other cases (like below) we were.

internal/controller/progressive/progressive.go
```diff
@@ -494,13 +492,18 @@ func startUpgradeProcess(
 	if err != nil {
 		return newUpgradingChildDef, false, err
 	}
+	// reset the Status to reflect our new Upgrading child
+	err = rolloutObject.ResetUpgradingChildStatus(newUpgradingChildDef)
+	if err != nil {
+		return newUpgradingChildDef, false, fmt.Errorf("processing upgrading child, failed to reset the upgrading child status for child %s/%s", newUpgradingChildDef.GetNamespace(), newUpgradingChildDef.GetName())
+	}
```
juliev0, line 499: We were previously calling this too late in the game (in [processUpgradingChild()](https://github.com/numaproj/numaplane/blob/main/internal/controller/progressive/progressive.go#L242-L252)) - now reset the upgrading status as soon as we have a new upgrading child. Note, I left the other call to this in there just in case it doesn't get set here.

afugazzotto (approved):

PR #667, opened 2025-03-21
feat: create AnalysisRun for MonoVertexRollout and PipelineRollout during progressive rollout

### LINKED ISSUE #653: Create AnalysisRuns for MonoVertexRollout and PipelineRollout during progressive rollout
# Summary

If somebody's spec indicates they want to use AnalysisTemplates/ClusterAnalysisTemplates, then there is a [function](https://github.com/argoproj/argo-rollouts/blob/master/utils/analysis/helpers.go#L292) which can create a single AnalysisRun from those. 

## Arguments
We can pass into that function the Arguments that are configured in the MonoVertexRollout. In addition, we can have 2 special Arguments which Numaplane computes:
- `monovertex-name`
- `monovertex-namespace`

## Name of the AnalysisRun

I was originally thinking we could name it the same as the MonoVertex or the Pipeline since there should only be one AnalysisRun (or otherwise zero) for each, but how about instead of that we use the MonoVertex/Pipeline name as the `generateName` (just in case we end up needing to create multiple for some reason)?  We should label it so it's identifiable as being part of numaplane (there's a function argument for that) - possibly `"submitted-by": "numaplane"`?

## Timeline

AnalysisRun gets created at the same time as the ChildAssessmentWindow starts. (AnalysisRun will be able to run as long as it needs, up until some configurable timeout, but don't worry about this for this PR.)

## Code Structure 

Since PipelineRollout and MonoVertexRollout both need this, we will share as much code as possible, which we could potentially put in a new file `internal/controller/progressive/analysis.go`. 

Note that I added some [temporary code which is commented out](https://github.com/numaproj/numaplane/blob/main/internal/controller/monovertexrollout/monovertexrollout_progressive.go#L45), which I was able to test. 

## Update Status

We also want to update the Status with the name of the analysis run and the start time (there are now fields for this).

## CheckList
- [ ] owner reference (monovertex)
- [ ] Use arguments from MonoVertexRollout
- [ ] `monovertex-name` and `monovertex-namespace` arguments calculated and passed in
- [ ] labeled `"submitted-by": "numaplane"`
- [ ] Status fields updated

---

dpadhiar:
Fixes #653 

### Modifications

If a user's spec indicates they want to use Analysis/ClusterAnalysisTemplates, Numaplane will create an AnalysisRun during progressive rollout for Pipeline and MonoVertex rollouts. These AnalysisRuns are reconciled by Argo Rollouts and their output and endTime for completion is reported in the `upgradingChildStatus` of the rollout.

WIP

### Verification

Running progressive upgrades locally for a MonoVertex rollout by observing that the AnalysisRun is created and reconciled successfully.

More testing to be done.


juliev0 (changes requested):

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -39,43 +44,44 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
+				analysisStatus.StartTime = &timeNow
+				mvtxRollout.SetAnalysisStatus(analysisStatus)
+			} else {
 				return apiv1.AssessmentResultUnknown, err
 			}
+		}
+
+		// assess analysisRun status and set endTime and phase if completed
+		analysisStatus := mvtxRollout.GetAnalysisStatus()
+		if analysisRun.Status.Phase.Completed() && analysisStatus.EndTime == nil {
```
juliev0, line 73: nit: I know `analysisStatus` shouldn't be nil if we get here, but should we check for nil just in case in this if statement?

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -87,7 +92,40 @@ func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, ex
...
+				analysisStatus.StartTime = &timeNow
+				pipelineRollout.SetAnalysisStatus(analysisStatus)
+			} else {
+				return apiv1.AssessmentResultUnknown, err
+			}
+		}
+
+		// assess analysisRun status and set endTime and phase if completed
+		analysisStatus := pipelineRollout.GetAnalysisStatus()
+		if analysisRun.Status.Phase.Completed() && analysisStatus.EndTime == nil {
```
juliev0, line 118: maybe check for nil just in case?

pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go
```diff
@@ -181,6 +203,12 @@ func (pipelineRollout *PipelineRollout) ResetUpgradingChildStatus(upgradingPipel
 		},
 	}
 
+	// only set analysisStatus if Analysis is set
+	if len(pipelineRollout.GetAnalysis().Templates) > 0 {
+		upgradingPipelineStatus.Analysis = AnalysisStatus{}
```
juliev0, line 208: We should just clear this regardless of whether any there are any Templates configured

pkg/apis/numaplane/v1alpha1/monovertexrollout_types.go
```diff
@@ -157,6 +178,16 @@ func (monoVertexRollout *MonoVertexRollout) ResetUpgradingChildStatus(upgradingM
 			},
 		},
 	}
+
+	// only set analysisStatus if Analysis is set
+	if len(monoVertexRollout.GetAnalysis().Templates) > 0 {
```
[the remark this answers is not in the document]
juliev0: We should just clear this regardless of whether any there are any Templates configured

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -39,43 +44,44 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
+			analysisStatus.EndTime = analysisRun.Status.CompletedAt
+			analysisStatus.Phase = analysisRun.Status.Phase
+			mvtxRollout.SetAnalysisStatus(analysisStatus)
 		} else {
-			return apiv1.AssessmentResultUnknown, err
+			analysisStatus.Phase = analysisRun.Status.Phase
+			mvtxRollout.SetAnalysisStatus(analysisStatus)
```
juliev0, line 79: maybe just in case somehow we failed to write the AnalysisRunName in the first place (i.e. resource version conflict issue), we can do it here just in case?

Also, you can take this out of the if statement since it pertains to both cases:
```
  analysisStatus.AnalysisRunName = existingUpgradingChildDef.GetName()
  analysisStatus.Phase = analysisRun.Status.Phase
  mvtxRollout.SetAnalysisStatus(analysisStatus)
```

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -87,7 +92,40 @@ func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, ex
...
+			analysisStatus.EndTime = analysisRun.Status.CompletedAt
+			analysisStatus.Phase = analysisRun.Status.Phase
+			pipelineRollout.SetAnalysisStatus(analysisStatus)
+		} else {
+			analysisStatus.Phase = analysisRun.Status.Phase
+			pipelineRollout.SetAnalysisStatus(analysisStatus)
```
juliev0, line 124: same comments as for monovertexrollout

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -39,43 +44,47 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
 
-	return progressive.AssessUpgradingPipelineType(ctx, existingUpgradingChildDef, progressive.AreVertexReplicasReady)
+		// assess analysisRun status and set endTime if completed
+		analysisStatus := mvtxRollout.GetAnalysisStatus()
+		// prevents us from continually setting the endTime each time the child is assessed after the analysis run completes
+		if analysisStatus != nil && analysisRun.Status.Phase.Completed() && analysisStatus.EndTime == nil {
+			analysisStatus.EndTime = analysisRun.Status.CompletedAt
+		}
+		analysisStatus.AnalysisRunName = existingUpgradingChildDef.GetName()
+		analysisStatus.Phase = analysisRun.Status.Phase
```
juliev0, line 81: it looks like we actually need to check for nil around all of this

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -39,43 +44,47 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
 			}
-		} else {
-			return apiv1.AssessmentResultUnknown, err
 		}
-	}*/
 
-	return progressive.AssessUpgradingPipelineType(ctx, existingUpgradingChildDef, progressive.AreVertexReplicasReady)
+		// assess analysisRun status and set endTime if completed
+		analysisStatus := mvtxRollout.GetAnalysisStatus()
+		// prevents us from continually setting the endTime each time the child is assessed after the analysis run completes
```
juliev0, line 78: seeing this comment - is there a reason not to just set it every time?

dpadhiar: Looking at this again.. I had previously added the `analysisStatus.EndTime == nil` guard when I was using `time.Now()` (during that assessment) as the EndTime which wasn't accurate to the `analysisRun.Status.CompletedAt` which I use now.

In this case, we actually don't need that guard and I can update this comment to be more accurate since we will always only update the EndTime when the AnalysisRun is completed anyway.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
dpadhiar: Moved nil guard from line 77 to outside this if loop. Same for PipelineRollout.

internal/controller/monovertexrollout/monovertexrollout_progressive.go
juliev0: that's good, thank you

juliev0 (approved):
just one typo but otherwise approved

PR #672, opened 2025-03-25
feat: Scale Pipeline Vertices to 0 on failure

### LINKED ISSUE #636: Scale down Failed Pipeline to 0 Pods
# Summary

Similar to MonoVertex, which was already implemented [here](https://github.com/numaproj/numaplane/pull/634), we need to be able to:
1. scale down the Failed Pipeline such that every Vertex goes to 0 Pods (alternative is to just scale Source Vertices: the former would have parity with what we do during the upgrade, but the latter seems preferable in that it saves overall resources, which after talking to Zach Aller, seems like a main motivation for Argo Rollouts to scale down.
2. in the case of "force promote", scale the promoted Pipeline back up again.

One thing to note: the strategy for determining how to scale back up in the case of "forcePromote" done for [MonoVertex](https://github.com/numaproj/numaplane/pull/634) should be improved - it uses the scale values from the Rollout, but better would be to save the scale definition in the upgradingChildStatus (as mentioned in the description of that PR). 

---

juliev0:
Fixes #636 

### Modifications

This performs the first set of changes required to update Pipeline logic to be more similar to MonoVertex as far as scaling.

With this PR, we:
1. scale down all Vertices of "upgrading" pipeline to 0 on failure.
2. if force promote is done, we scale them all back up to their original values.

In order to save the min/max, I have added a new field to the UpgradingPipelineStatus called [OriginalScaleMinMax](https://github.com/numaproj/numaplane/blob/pipeline-scale-0/pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go#L75) which stores the original patch strings for each Vertex using an array. 

### Verification

For Pipeline (and also for MonoVertex just in case), I tested the following:
I performed the following test for 3 different types of source vertices:
1. source vertex has no `scale` defined
4. source vertex has `scale` defined but no `min` or `max`, but does have an additional field under `scale`
5. source vertex has `scale.min` and `scale.max` both defined plus an additional field under `scale`
Test performed:
Update to a bad change, observe standard scale down of "promoted" during upgrade. 
After failure, observe all Vertices have scale.min and scale.max = 0 (plus any other fields under `scale` preserved). Observe standard scale up of "promoted" child.
Force promote and observe the newly promoted child is scaled back up (plus any other fields under `scale` preserved).


config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -19,6 +19,7 @@ spec:
       scale:
         min: 2
         max: 3
+        lookbackSeconds: 60
```
juliev0, line 22: just kept this in here after my testing, since it's useful I guess to verify it doesn't get touched during all of the rescaling ?

internal/controller/common/numaflowtypes/pipeline.go
```diff
@@ -198,14 +198,30 @@ func PipelineWithoutDesiredPhase(pipeline *unstructured.Unstructured) {
...
+
+			modifiedVertices = append(modifiedVertices, vertexAsMap)
+		}
 	}
+
+	err := unstructured.SetNestedSlice(pipeline.Object, modifiedVertices, "spec", "vertices")
+	if err != nil {
+		return err
+	}
+	return nil
```
juliev0, line 226: this was a total goof on my previous PR when I was going a little too fast. This code hadn't been used and tested at the time, which was why it got in there.

juliev0: Note that I have a unit test which is checking this now - https://github.com/numaproj/numaplane/blob/pipeline-scale-0/internal/controller/pipelinerollout/pipelinerollout_controller_test.go#L532-L545

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -169,20 +169,20 @@ func (r *MonoVertexRolloutReconciler) ProcessUpgradingChildPostFailure(
 	// need to check to see if it's already scaled down before we do this
 	existingSpec := upgradingMonoVertexDef.Object["spec"].(map[string]interface{})
 
-	existingScaleMin, existingScaleMax, err := getScaleValuesFromMonoVertexSpec(existingSpec)
+	existingScale, err := getScaleValuesFromMonoVertexSpec(existingSpec)
```
juliev0, line 172: This is basically just a new struct I added:
```
type ScaleDefinition struct {
	Min *int64
	Max *int64
}
```

internal/controller/pipelinerollout/pipelinerollout_controller_test.go
```diff
@@ -127,7 +127,7 @@ func init() {
 	runningPipelineSpecWithTopologyChange = withInterstepBufferService(pipelineSpecWithTopologyChange, ctlrcommon.DefaultTestISBSvcName)
 }
 
-var yamlHasDesiredPhase = `
+var specHasDesiredPhase = `
```
juliev0, line 130: realized this was json, not yaml :) so I changed the names

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -246,7 +246,45 @@ func (r *PipelineRolloutReconciler) ProcessUpgradingChildPostFailure(
 
 	numaLogger.Debug("started post-failure processing of upgrading pipeline")
 
-	// TODO
+	// scale down every Vertex to 0 Pods
```
afugazzotto, line 249: nit: I prefer to put the scaling down code in a separate func and than call that func here so that `ProcessUpgradingChildPostFailure` can just be a collection of operations to perform post failure (instead of being strictly for scaling down).

afugazzotto (approved):

internal/controller/common/numaflowtypes/pipeline.go
```diff
@@ -198,14 +198,30 @@ func PipelineWithoutDesiredPhase(pipeline *unstructured.Unstructured) {
...
+		if vertexAsMap, ok := vertex.(map[string]any); ok {
+
+			unstructured.RemoveNestedField(vertexAsMap, "scale", "min")
+			unstructured.RemoveNestedField(vertexAsMap, "scale", "max")
+
+			// if "scale" is there and empty, remove it
+			scaleMap, found := vertexAsMap["scale"].(map[string]interface{})
+			if found && len(scaleMap) == 0 {
+				unstructured.RemoveNestedField(vertexAsMap, "scale")
+			}
```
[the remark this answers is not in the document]
juliev0: so, we're trying to effectively compare spec1 to spec2. Say one of them defines `scale` and the other one does not. We want to call them "equal" in that case.

internal/controller/pipelinerollout/pipelinerollout_progressive.go
juliev0: yeah, that's a good idea

juliev0: done

Issue #670, opened 2025-03-24
Add more information to Failure assessment result to know why failure occurred


juliev0:
# Summary

In the [upgradingChildStatus](https://github.com/numaproj/numaplane/blob/v0.18.0/pkg/apis/numaplane/v1alpha1/progressive_status_types.go#L35-L46), let's add 2 fields which get set when [AssessmentResult](https://github.com/numaproj/numaplane/blob/v0.18.0/pkg/apis/numaplane/v1alpha1/progressive_status_types.go#L39) is set to "Failure":
1. `FailureReason`: for Pipeline and MonoVertex this should be set to things like:
- "readyReplicas (1) < desiredReplicas (3)"
- "phase: Failed"
- "Condition 'DaemonHealthy' set False"
(Note, we only need to list one of those in the case that multiple failed; I think that's sufficient)
For ISBService, we could write something like "Pipeline my-pipeline-3 Failed" as the Reason.
2. `ChildStatus`: a full dump of the child `Status`. I'm not sure if we want to bother setting this for isbsvc or not. Probably just leave that unset, unless you want to set it. 

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
Actually, reassigned to @chandankumar4 - I think it would be a good one for him to pick up.

PR #673, opened 2025-03-26
feat: Add failure reason for progressive rollout

chandankumar4:
Part of: https://github.com/numaproj/numaplane/issues/670

### Modifications

- Add progressive rollout failure reason in case of failure

### Verification

Verified in local kubernetes cluster
```bash
  Progressive Status:
    Promoted Mono Vertex Status:
      Name:                               test-monovertex-rollout-0
      Scale Values Restored To Original:  true
    Upgrading Mono Vertex Status:
      Analysis:
        Phase:
      Assessment End Time:     2025-03-26T14:57:58Z
      Assessment Result:       Failure
      Assessment Start Time:   2025-03-26T14:56:58Z
      Failure Reason:          readyReplicas=1 is less than desiredReplicas=3
      Name:                    test-monovertex-rollout-1
      Original Scale Min Max:  {"max":null,"min":5}
  Upgrade In Progress:         Progressive
```

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_progressive.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_progressive.go
@@ -63,7 +63,7 @@ func (r *PipelineRolloutReconciler) CreateUpgradingChildDefinition(ctx context.C
 
 // AssessUpgradingChild makes an assessment of the upgrading child to determine if it was successful, failed, or still not known
 // This implements a function of the progressiveController interface
-func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, rolloutObject progressive.ProgressiveRolloutObject, existingUpgradingChildDef *unstructured.Unstructured) (apiv1.AssessmentResult, apiv1.AssessmentFailureReason, error) {
+func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, rolloutObject progressive.ProgressiveRolloutObject, existingUpgradingChildDef *unstructured.Unstructured) (apiv1.AssessmentResult, string, error) {
 	verifyReplicasFunc := func(existingUpgradingChildDef *unstructured.Unstructured) (bool, string, error) {
 		verticesList, err := kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion,
 			numaflowv1.VertexGroupVersionResource.Resource, existingUpgradingChildDef.GetNamespace(),
@@ -90,11 +90,8 @@ func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, ro
 				break
 			}
 		}
-		if !areAllVerticesReplicasReady {
-			return areAllVerticesReplicasReady, replicasFailureReason, nil
-		}
 
-		return areAllVerticesReplicasReady, "", nil
+		return areAllVerticesReplicasReady, replicasFailureReason, nil
 	}
 
 	pipelineRollout := rolloutObject.(*apiv1.PipelineRollout)
--- a/internal/controller/pipelinerollout/pipelinerollout_progressive.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_progressive.go
@@ -63,33 +63,38 @@ func (r *PipelineRolloutReconciler) CreateUpgradingChildDefinition(ctx context.C
 
 // AssessUpgradingChild makes an assessment of the upgrading child to determine if it was successful, failed, or still not known
 // This implements a function of the progressiveController interface
-func (r *PipelineRolloutReconciler) AssessUpgradingChild(ctx context.Context, rolloutObject progressive.ProgressiveRolloutObject, existingUpgradingChildDef *unstructured.Unstructured) (apiv1.AssessmentResult, error) {
-	verifyReplicasFunc := func(existingUpgradingChildDef *unstructured.Unstructured) (bool, error) {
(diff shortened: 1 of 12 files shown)


pkg/apis/numaplane/v1alpha1/progressive_status_types.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 type AssessmentResult string
+type AssessmentFailureReason string
```
juliev0, line 27: is there any reason to have this as its own type? Were you just imitating the fact that `AssessmentResult` is its own type? The reason we have that as its own type is because it has a set of very specific values, but I don't think this does, right?

internal/controller/progressive/progressive.go
```diff
@@ -604,16 +616,20 @@ Returns:
   - A boolean indicating whether the ready replicas are sufficient.
   - An error if there is an issue retrieving the replica counts.
```
juliev0, line 617:
```suggestion
  - In the case that ready replicas aren't sufficient, a message providing more information 
  - An error if there is an issue retrieving the replica counts.
```

internal/controller/progressive/progressive.go
```diff
@@ -520,16 +532,16 @@ func startUpgradeProcess(
 }
 
 // return true if all Conditions are true
-func checkChildConditions(upgradingObjectStatus *kubernetes.GenericStatus) bool {
+func checkChildConditions(upgradingObjectStatus *kubernetes.GenericStatus) (bool, string) {
```
juliev0, line 535: from this function instead of returning a string, can we return the first Condition that failed?

Then, on the caller side create the string which includes both the name of the Condition, the fact that it was false, and the `c.Reason`.

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -63,33 +63,38 @@ func (r *PipelineRolloutReconciler) CreateUpgradingChildDefinition(ctx context.C
...
+				replicasFailureReason = failureReason
 				break
 			}
 		}
+		if !areAllVerticesReplicasReady {
+			return areAllVerticesReplicasReady, replicasFailureReason, nil
+		}
 
-		return areAllVerticesReplicasReady, nil
+		return areAllVerticesReplicasReady, "", nil
```
juliev0, line 97: Unless I'm mistaken, I believe we should be able to consolidate this into just:
`return areAllVerticesReplicasReady, replicasFailureReason, nil`

internal/controller/progressive/progressive.go
```diff
@@ -405,19 +406,30 @@ func AssessUpgradingPipelineType(
...
+}
+
+func CalculateFailureReason(replicasFailureReason, phase, conditionFailureReason string) apiv1.AssessmentFailureReason {
+	if replicasFailureReason != "" {
+		return apiv1.AssessmentFailureReason(replicasFailureReason)
+	} else if phase == "Failed" {
+		return "child phase is in Failed state"
+	} else {
+		return apiv1.AssessmentFailureReason(conditionFailureReason)
+	}
```
juliev0, line 432: Let's reorder these, such that:
- phase == "Failed" is first
- condition Failure is second
- replicas is last

internal/controller/progressive/progressive.go
```diff
@@ -405,19 +406,30 @@ func AssessUpgradingPipelineType(
...
-		return apiv1.AssessmentResultFailure, nil
+		failureReason := CalculateFailureReason(replicasFailureReason, upgradingObjectStatus.Phase, conditionFailureReason)
+		return apiv1.AssessmentResultFailure, failureReason, nil
 	}
 
-	return apiv1.AssessmentResultUnknown, nil
+	return apiv1.AssessmentResultUnknown, "", nil
+}
+
+func CalculateFailureReason(replicasFailureReason, phase, conditionFailureReason string) apiv1.AssessmentFailureReason {
```
juliev0, line 425:
```suggestion
// CalculateFailureReason issues a reason for failure; if there are multiple reasons, it returns one of them
func CalculateFailureReason(replicasFailureReason, phase, conditionFailureReason string) apiv1.AssessmentFailureReason {
```

pkg/apis/numaplane/v1alpha1/progressive_status_types.go
chandankumar4: For keeping it consistent with `AssessmentResult`, I have added this type. I don't have any solid reason, I'll remove it for now.

internal/controller/pipelinerollout/pipelinerollout_progressive.go
chandankumar4: Yaa, It can be merged to one

internal/controller/progressive/progressive.go
```diff
@@ -405,19 +406,31 @@ func AssessUpgradingPipelineType(
...
-	return apiv1.AssessmentResultUnknown, nil
+	return apiv1.AssessmentResultUnknown, "", nil
+}
+
+// CalculateFailureReason issues a reason for failure; if there are multiple reasons, it returns one of them
+func CalculateFailureReason(replicasFailureReason, phase string, failedCondition *metav1.Condition) string {
+	if phase == "Failed" {
+		return "child phase is in Failed state"
+	} else if failedCondition != nil {
+		return fmt.Sprintf("condition %s is %s", failedCondition.Type, failedCondition.Reason)
```
juliev0, line 430:
```suggestion
		return fmt.Sprintf("condition %s is False for Reason: %s", failedCondition.Type, failedCondition.Reason)
```

juliev0 (approved):

Issue #677, opened 2025-03-27
Incorporate AnalysisRun failure into failure reasons


juliev0:
# Summary

[This](https://github.com/numaproj/numaplane/pull/671) PR got added to cause failure for PipelineRollout and MonoVertexRollout in the case that user has configured AnalysisTemplates in their Rollouts and gets a failure from that. We need to incorporate this as one of our failure reasons. Perhaps we add it last to the `CalculateFailureReason()` function.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

dpadhiar:
@juliev0 Think we should be able to close this now since @chandankumar4 PR #676 is merged. Can reopen if otherwise.

PR #679, opened 2025-03-27
feat: add analysisRunTimeout to Numaplane controller config

### LINKED ISSUE #656: Have a timeout for AnalysisRun run time
# Summary

We should configure a time out time in Numaplane Controller Config, beyond which AnalysisRuns cannot continue to run. Probably should fail the progressive rollout if it's not complete by that time.

---

dpadhiar:
Fixes #656 

### Modifications

Adds option to set `analysisRunTimeout` to the Numaplane controller config in the progressive field.

When using AnalysisRuns during a progressive upgrade, we should consider a progressive upgrade a failure if the AnalysisRun is taking too long to come to a Completed phase which is a Success, Failure, Error or Inconclusive.

```
switch analysisStatus.Phase {
case argorolloutsv1.AnalysisPhaseSuccessful:
	return apiv1.AssessmentResultSuccess, "", nil
case argorolloutsv1.AnalysisPhaseError, argorolloutsv1.AnalysisPhaseFailed, argorolloutsv1.AnalysisPhaseInconclusive:
	return apiv1.AssessmentResultFailure, "", nil
default:
	// if analysisRun is in a non Completed phase, we check if it has exceeded the analysisRunTimeout
	if time.Since(analysisStatus.StartTime.Time) >= analysisRunTimeout {
		return apiv1.AssessmentResultFailure, "", nil
	}
	return apiv1.AssessmentResultUnknown, "", nil
}
```

Example config (default timeout is currently 10 minutes)
```
progressive:
  defaultAssessmentSchedule: 
    - kind: Pipeline
      schedule: "120,60,10"
    - kind: MonoVertex
      schedule: "120,60,10"
    - kind: InterstepBufferService
      schedule: "0,0,10"
  analysisRunTimeout: "600"
```

### Verification

Ran multiple progressive upgrades on a MonoVertexRollout and saw that each AnalysisRun completed well under the timeout.

Not setting the analysisRunTimeout in the config while using the AnalysisRun for progressive upgrades may cause issues but can be fixed by simply updating config.


internal/controller/progressive/progressive.go
```diff
@@ -404,6 +417,11 @@ func AssessUpgradingPipelineType(
 		WithValues("namespace", existingUpgradingChildDef.GetNamespace(), "name", existingUpgradingChildDef.GetName()).
 		Debugf("Upgrading child is in phase %s, conditions healthy=%t, ready replicas match desired replicas=%t", upgradingObjectStatus.Phase, healthyConditions, healthyReplicas)
 
+	analysisRunTimeout, err := getAnalysisRunTimeout(ctx)
+	if err != nil {
+		return apiv1.AssessmentResultUnknown, "", err
+	}
```
juliev0, line 423: So, @chandankumar4 recently added a [PR](https://github.com/numaproj/numaplane/pull/673) to return the reason for failure if there was a failure. I wrote up [this](https://github.com/numaproj/numaplane/issues/677) and originally assigned to him but now reassigning to you so you can maybe just take care of it here.

See the `if` statement below?:
```
if upgradingObjectStatus.Phase == "Failed" || !healthyConditions || !healthyReplicas {
...
}
```

Can we move that above line 420? 

So, we'll return failure and the corresponding reason if there's a more fundamental failure than the AnalysisRun itself. 

Then down here make sure you return a reason if these other cases of AssessmentResultFailure occur.

internal/controller/progressive/progressive.go
```diff
@@ -416,6 +434,10 @@ func AssessUpgradingPipelineType(
 			case argorolloutsv1.AnalysisPhaseError, argorolloutsv1.AnalysisPhaseFailed, argorolloutsv1.AnalysisPhaseInconclusive:
 				return apiv1.AssessmentResultFailure, "", nil
 			default:
+				// if analysisRun is a non Completed phase yet, we check if it has exceeded the analysisRunTimeout
```
juliev0, line 437:
```suggestion
				// if analysisRun is not completed yet, we check if it has exceeded the analysisRunTimeout
```

juliev0:
to verify it, you can try temporarily configuring it to a very small time duration and maybe update some of [these](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/v1alpha1/analysis_types.go#L95-L101) AnalysisTemplate fields

internal/controller/progressive/progressive.go
dpadhiar: I move the block to above 420 and added a failure reason for when the timeout is exceeded.

internal/controller/config/config.go
```diff
@@ -358,3 +361,15 @@ func (config *ProgressiveConfig) GetChildStatusAssessmentSchedule(kind string) (
...
+
+// GetAnalysisRunTimeout parses the GlobalConfig for the analysisRunTimeout string to convert into a time.Duration
+func (config *ProgressiveConfig) GetAnalysisRunTimeout() (time.Duration, error) {
+
+	analysisRunTimeout, err := strconv.Atoi(config.AnalysisRunTimeout)
+	if err != nil {
+		return 0, fmt.Errorf("invalid analysisRunTimeout value: %w", err)
+	}
+
+	return time.Duration(analysisRunTimeout) * time.Second, nil
```
juliev0, line 373: sorry, just thought of this - I think a typical thing to do is to have a default constant in the code if it's not configured. Can we set ours to 15 or 20 minutes?

config/install.yaml
```diff
@@ -2241,6 +2241,7 @@ data:
           schedule: "120,60,10"
         - kind: InterstepBufferService
           schedule: "0,0,10"
+      analysisRunTimeout: 600
```
juliev0, line 2244: let's make it 15 or 20 minutes

internal/controller/config/config.go
dpadhiar: Got it, so we should both set the default in our manifests to 20 minutes and also return 20 minutes if its unset here.

config/install.yaml
dpadhiar: Going with 20 minutes for now.

internal/controller/progressive/progressive.go
juliev0: looks good

internal/controller/config/config.go
juliev0: thanks, did you test it not being set just in case?

juliev0: I'm thinking you'll get an error in your Atoi conversion if it's an empty string

internal/controller/progressive/progressive.go
```diff
@@ -410,10 +423,20 @@ func AssessUpgradingPipelineType(
 		return apiv1.AssessmentResultUnknown, replicasFailureReason, err
 	}
 
+	if upgradingObjectStatus.Phase == "Failed" || !healthyConditions || !healthyReplicas {
+		failureReason := CalculateFailureReason(replicasFailureReason, upgradingObjectStatus.Phase, failedCondition)
+		return apiv1.AssessmentResultFailure, failureReason, nil
+	}
```
juliev0, line 429: sorry if I may have said the wrong thing, but let's actually make sure the logging that's just below this occurs prior to this

internal/controller/config/config.go
dpadhiar: That's probably the case, let me change the way this checks the string then.

dpadhiar:
> to verify it, you can try temporarily configuring it to a very small time duration and maybe update some of [these](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/v1alpha1/analysis_types.go#L95-L101) AnalysisTemplate fields

Tested with a super quick timeout, here is the resulting progressive status.

```
progressiveStatus:
    promotedMonoVertexStatus:
      name: my-monovertex-0
      scaleValuesRestoredToOriginal: true
    upgradingMonoVertexStatus:
      analysis:
        analysisRunName: my-monovertex-1
        phase: Running
        startTime: "2025-03-28T22:30:07Z"
      assessmentEndTime: "2025-03-28T22:31:17Z"
      assessmentResult: Failure
      assessmentStartTime: "2025-03-28T22:30:06Z"
      childStatus:
        conditions:
        - lastTransitionTime: "2025-03-28T22:30:24Z"
          message: Successful
          reason: Successful
          status: "True"
          type: DaemonHealthy
        - lastTransitionTime: "2025-03-28T22:29:06Z"
          message: Successful
          reason: Successful
          status: "True"
          type: Deployed
        - lastTransitionTime: "2025-03-28T22:29:06Z"
          message: All pods are healthy
          reason: Running
          status: "True"
          type: PodsHealthy
        currentHash: 66e0ab00439f0bae923ffe4ffd98913db4157121480958822d357bc1101d5e97
        desiredReplicas: 0
        lastScaledAt: "2025-03-28T22:30:17Z"
        lastUpdated: "2025-03-28T22:30:24Z"
        observedGeneration: 2
        phase: Running
        replicas: 0
        selector: app.kubernetes.io/component=mono-vertex,numaflow.numaproj.io/mono-vertex-name=my-monovertex-1
        updateHash: 66e0ab00439f0bae923ffe4ffd98913db4157121480958822d357bc1101d5e97
        updatedReadyReplicas: 1
      failureReason: AnalysisRun my-monovertex-1 in phase Running has exceeded the
        analysisRunTimeout
      name: my-monovertex-1
      originalScaleMinMax: '{"max":3,"min":2}'
  upgradeInProgress: Progressive
```

internal/controller/config/config.go
```diff
@@ -358,3 +361,21 @@ func (config *ProgressiveConfig) GetChildStatusAssessmentSchedule(kind string) (
...
+// GetAnalysisRunTimeout parses the GlobalConfig for the analysisRunTimeout string to convert into a time.Duration
+func (config *ProgressiveConfig) GetAnalysisRunTimeout() (time.Duration, error) {
+
+	// default timeout is 20 minutes
+	analysisRunTimeout := 1200
+
+	if config.AnalysisRunTimeout != "" {
+		analysisRunTimeout, err := strconv.Atoi(config.AnalysisRunTimeout)
+		if err != nil {
+			return 0, fmt.Errorf("invalid analysisRunTimeout value: %w", err)
```
juliev0, line 374: returning an error would effectively put the Controller in a crash loop. Instead let's log an Error and then just use the default

internal/controller/config/config.go
```diff
@@ -358,3 +361,21 @@ func (config *ProgressiveConfig) GetChildStatusAssessmentSchedule(kind string) (
...
+		}
+		return time.Duration(analysisRunTimeout) * time.Second, nil
```
juliev0, line 376: this line isn't needed since it will fall through to the line below

dpadhiar: Like this right? VSCode is complaining about the analysisrunTimeout being returned by strconv for some reason.

```
// default timeout is 20 minutes
	analysisRunTimeout := 1200

	if config.AnalysisRunTimeout != "" {
		analysisRunTimeout, err := strconv.Atoi(config.AnalysisRunTimeout)
		if err != nil {
			return 0, fmt.Errorf("invalid analysisRunTimeout value: %w", err)
		}
		// return time.Duration(analysisRunTimeout) * time.Second, nil
	}

	return time.Duration(analysisRunTimeout) * time.Second, nil
```

internal/controller/config/config.go
```diff
@@ -358,3 +361,18 @@ func (config *ProgressiveConfig) GetChildStatusAssessmentSchedule(kind string) (
...
+// GetAnalysisRunTimeout parses the GlobalConfig for the analysisRunTimeout string to convert into a time.Duration
+func (config *ProgressiveConfig) GetAnalysisRunTimeout() (time.Duration, error) {
+
+	// default timeout is 20 minutes
+	defaultAnalysisRunTimeout := 1200
+
+	analysisRunTimeout, err := strconv.Atoi(config.AnalysisRunTimeout)
+	if err != nil {
+		return time.Duration(defaultAnalysisRunTimeout), fmt.Errorf("invalid analysisRunTimeout value: %w", err)
+	}
```
juliev0, line 374: won't this return an error if it's empty string? we don't want that. just return the default if it is

juliev0 (approved):

PR #683, opened 2025-04-01
fix: default timeout was returning as nanoseconds instead of seconds

dpadhiar:
### Modifications

Default timeout for analysisRunTimeout was returning as 1200 nanoseconds instead of 1200 seconds (20 minutes).

### Verification

Running with unit tests locally (will be pushed in another PR).

n/a

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -369,7 +369,7 @@ func (config *ProgressiveConfig) GetAnalysisRunTimeout() (time.Duration, error) 
 	defaultAnalysisRunTimeout := 1200
 
 	if config.AnalysisRunTimeout == "" {
-		return time.Duration(defaultAnalysisRunTimeout), nil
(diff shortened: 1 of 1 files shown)


internal/controller/config/config.go
```diff
@@ -369,7 +369,7 @@ func (config *ProgressiveConfig) GetAnalysisRunTimeout() (time.Duration, error)
 	defaultAnalysisRunTimeout := 1200
 
 	if config.AnalysisRunTimeout == "" {
-		return time.Duration(defaultAnalysisRunTimeout), nil
+		return time.Duration(defaultAnalysisRunTimeout) * time.Second, nil
```
juliev0, line 372: it looks like below on line 377 you're returning `time.Duration(defaultAnalysisRunTimeout)` - do you need to change that too?

dpadhiar: Yes, hadn't changed that from before since we were just only returning the error in that case. Will change that.

juliev0 (approved):

Issue #680, opened 2025-03-28
mismatch between FailureReason and child status


juliev0:
**Describe the bug**

In the code, the call [CanAssess()](https://github.com/numaproj/numaplane/blob/73e1c5af1dee1e61c18374252ca3e3fbc78272d1/internal/controller/progressive/progressive.go#L281-L285) prevents an assessment if it's AssessmentResultFailure already. This causes the failureReason to only get set once.

However, the child status field is getting set multiple times. We need to make it so that that child status only gets set if it hasn't been set yet.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

PR #687, opened 2025-04-03
fix: mismatch between FailureReason and child status

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/680

### Modifications

- Fixed the mismatch between FailureReason and child status

### Verification

- Verified in local k8s cluster after deploying the changes, ran `STRATEGY=progressive make test-progressive-e2e` and keep deleting the newly created pod during upgrade to cause failure.

### DIFF
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -299,13 +299,13 @@ func processUpgradingChild(
 	// Otherwise, assess the previous child status.
 	assessment := childStatus.AssessmentResult
 	failureReason := childStatus.FailureReason
-	failureChildStatus := childStatus.ChildStatus.Raw
+	childSts := childStatus.ChildStatus.Raw
 	if childStatus.CanAssess() {
 		assessment, failureReason, err = controller.AssessUpgradingChild(ctx, rolloutObject, existingUpgradingChildDef)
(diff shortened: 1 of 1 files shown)


internal/controller/progressive/progressive.go
```diff
@@ -299,11 +299,16 @@ func processUpgradingChild(
 	// Otherwise, assess the previous child status.
 	assessment := childStatus.AssessmentResult
 	failureReason := childStatus.FailureReason
+	failureChildStatus := childStatus.ChildStatus.Raw
 	if childStatus.CanAssess() {
 		assessment, failureReason, err = controller.AssessUpgradingChild(ctx, rolloutObject, existingUpgradingChildDef)
 		if err != nil {
 			return false, 0, err
 		}
+		failureChildStatus, err = json.Marshal(existingUpgradingChildDef.Object["status"])
```
juliev0, line 308: nit: do you mind calling it `childStatus` instead of `failureChildStatus`? I'm thinking at this line of code we don't really know if it's a failure, success, or unknown.

chandankumar4: `childStatus` variable is already used there, so probably I can use `childSts`

Issue #649, opened 2025-03-14
Update Pipeline to assess "readyReplicas" as part of Progressive assessment


juliev0:
# Summary

We need to accommodate older versions of Numaflow (<1.4.4) by checking for readyReplicas>=desiredReplicas for each Vertex.

Like with MonoVertex we need to:
1. Prevent autoscaling from making that check fail, by purposely setting the upgrading Pipeline Vertices max=min.
2. Update the Pipeline back to correct min and max once Promoted

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
Just realized this was [done](https://github.com/numaproj/numaplane/blob/v0.20.0/internal/controller/pipelinerollout/pipelinerollout_progressive.go#L82) so closing

PR #689, opened 2025-04-05
feat: Pipeline scaling

juliev0:
### Modifications

This PR updates the Pipeline Vertex scaling that occurs as part of Progressive rollout, which is described as Option 1 [here](https://docs.google.com/document/d/1aZfyHX_C3vj3YncqFx4AJk4HcIaUnuFezt7eBYHeAfk/edit?tab=t.0#heading=h.ah1he7sh8zk6).

It is a precursor to updating the resource health check described by [this](https://github.com/numaproj/numaplane/issues/649) Issue.

It does the following:

1. Pre-upgrade: 
    - store all original Vertex scale values from "promoted" pipeline in Promoted Status
    - store all original Vertex scale values from "upgrading" pipeline in Upgrading Status
    - set the upgrading pipeline scale values to new min/max before creating pipeline
2. On failure:
    - scale up Promoted pipeline vertices all the way
    - scale down failed pipeline vertices to 0 Pods
3. On success (including forced success):
    - scale up upgrading pipeline vertices all the way
    - in order to delete promoted pipeline, we need to make sure no vertices have min=max=0; otherwise, they can't be drained; if that happens, patch them to 1 Pod

### Verification

Tested a pipeline with each vertex having a different scale definition:
1. no "scale" set at all
2. "scale" set but "min" and "max" not set
3. "scale" set with both "min" and "max" set

Tested case of failed pipeline and success pipeline and observed that promoted pipeline and upgrading pipeline scaled to the right values each time.

We haven't turned progressive on for Pipeline on clusters yet, so we don't need to worry about backward compatibility for that.

For MonoVertexRollout, this PR does rename a Status field `allSourceVerticesScaledDown` to `allVerticesScaledDown`. This field is only looked at while at the beginning of the progressive rollout, when the upgrading monovertex is being created, which is very unlikely to occur while updating numaplane. (Also, I think the net effect would be to simply re-scale down.)


internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -277,15 +277,31 @@ func (r *MonoVertexRolloutReconciler) ProcessUpgradingChildPreUpgrade(
 		return true, fmt.Errorf("unexpected type for ProgressiveRolloutObject: %+v; can't process upgrading monovertex pre-upgrade", rolloutObject)
 	}
 
+	err := scaleDownUpgradingMonoVertex(monoVertexRollout, upgradingMonoVertexDef)
```
juliev0, line 280: just separated this into a separate function

tests/e2e/progressive.go
```diff
@@ -138,16 +138,9 @@ func VerifyMonoVertexRolloutScaledDownForProgressive(
...
-// This function should be modified if the E2E tests will be changed
-// to have more than one source vertex.
 func VerifyPipelineRolloutScaledDownForProgressive(
 	pipelineRolloutName string,
-	expectedPromotedName string,
-	sourceVertexName string,
-	expectedCurrent int64,
-	expectedOriginalScaleMinMaxAsJSONString string,
-	expectedScaleTo int64,
+	expectedPromotedPipelineName string,
```
juliev0, line 143: We no longer treat the Source Vertex differently from any other vertex, so it doesn't make sense to treat Source in a special way. that's why I removed those Source-related checks.

A separate issue is written and prioritized to update the e2e testing for this PR.

pkg/apis/numaplane/v1alpha1/pipelinerollout_types.go
```diff
@@ -72,13 +72,20 @@ type UpgradingPipelineStatus struct {
 	InterStepBufferServiceName string `json:"interStepBufferServiceName,omitempty"`
 
 	// OriginalScaleMinMax stores for each vertex, the original scale min and max values as JSON string
-	OriginalScaleMinMax []VertexScale `json:"originalScaleMinMax,omitempty"`
+	//OriginalScaleMinMax []VertexScale `json:"originalScaleMinMax,omitempty"`
+	OriginalScaleMinMax []VertexScaleDefinition `json:"originalScaleMinMax,omitempty"`
```
juliev0, line 76: in the end, breaking the vertices down into Min and Max values made more sense than creating a json string to hold them, since it wouldn't work to patch with the json string anyway

pkg/apis/numaplane/v1alpha1/progressive_status_types.go
```diff
@@ -90,12 +90,12 @@ type PromotedChildStatus struct {
...
-	// ScaleValues is a map where the keys are the promoted child source vertices names
-	// and the values are the scale values of the source vertices
+	// ScaleValues is a map where the keys are the promoted child vertices names
+	// and the values are the scale values of the vertices
 	ScaleValues map[string]ScaleValues `json:"scaleValues,omitempty"`
-	// AllSourceVerticesScaledDown indicates if ALL the promoted child source vertices have been scaled down
-	AllSourceVerticesScaledDown bool `json:"allSourceVerticesScaledDown,omitempty"`
-	// ScaleValuesRestoredToOriginal indicates if ALL the promoted child source vertices have been set back to the original min and max scale values.
+	// AllVerticesScaledDown indicates if ALL the promoted child vertices have been scaled down
+	AllVerticesScaledDown bool `json:"allVerticesScaledDown,omitempty"`
```
juliev0, line 97: we no longer treat "Source" vertices different from other vertices. All vertices need to be scaled down in the promoted pipeline during the upgrade

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -1118,10 +1118,22 @@ func (r *PipelineRolloutReconciler) Recycle(ctx context.Context,
...
+		}
+	}
 
+	if requireDrain {
+		modified, err := r.ensurePipelineIsDrainable(ctx, pipeline)
+		if err != nil {
+			return false, err
+		}
+		if modified {
+			// come back later
```
chandankumar4, line 1135: Should we add "TODO" here to easily find later?

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -1118,10 +1118,22 @@ func (r *PipelineRolloutReconciler) Recycle(ctx context.Context,
 
 		case common.LabelValueProgressiveSuccess:
 			pause = true
-			requireDrain = true // TODO: make configurable (https://github.com/numaproj/numaplane/issues/512)
+			// TODO: make configurable (https://github.com/numaproj/numaplane/issues/512)
+			requireDrain = true
 
 		case common.LabelValueProgressiveFailure:
+			// this is the case of a previously failed "upgrading" pipeline which was replaced with a new one
+		}
```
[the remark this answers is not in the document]
juliev0: no, the only real purpose of this switch statement was to add `requireDrain = true`. Otherwise, it gets set to false. Let me add a comment to the switch statement so it's clear.

internal/controller/pipelinerollout/pipelinerollout_controller.go
juliev0: sorry, that didn't mean that we need to come back later to look at it, it just meant that we are requeuing here - let me know if I should improve the comment

chandankumar4: okay, It's sounds confusing, maybe we can mention "requeuing" also in the comment

juliev0: will do

internal/controller/pipelinerollout/pipelinerollout_controller.go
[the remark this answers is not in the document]
chandankumar4: But "requireDrain = true" is set in previous case(case common.LabelValueProgressiveSuccess)  but this (case common.LabelValueProgressiveFailure) case is not doing anything?

chandankumar4 (approved):
looks good

PR #692, opened 2025-04-07
fix: add OwnerReferences to created AnalysisRuns

### LINKED ISSUE #691: AnalysisRun OwnerReference to MonoVertex
**Describe the bug**
We want the AnalysisRun to get deleted when the MonoVertex/Pipeline gets deleted

---

dpadhiar:
Fixes #691 

### Modifications

Fixes issue where we were not adding an OwnerReference to related AnalysisRuns that are created during a Progressive upgrade that uses them.

### Verification

Running progressive upgrades locally on MonoVertexRollout and checking the created AnalysisRun is created with a ownerReference and is deleted when the rollout is.

```
- apiVersion: argoproj.io/v1alpha1
  kind: AnalysisRun
  metadata:
    creationTimestamp: "2025-04-07T23:59:28Z"
    generation: 2
    labels:
      app.kubernetes.io/part-of: numaplane
    name: my-monovertex-1
    namespace: example-namespace
    ownerReferences:
    - apiVersion: numaflow.numaproj.io/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: MonoVertex
      name: my-monovertex-1
      uid: 1cf0d96d-0cda-4fd7-890c-30ff3caf2f55
    resourceVersion: "890142"
    uid: 8ed46273-5310-4cee-922e-415e10909d8f
```

n/a

### DIFF
--- a/internal/controller/progressive/analysis.go
+++ b/internal/controller/progressive/analysis.go
@@ -91,6 +91,7 @@ Parameters:
   - analysis: struct which contains templateRefs to AnalysisTemplates and ClusterAnalysisTemplates and arguments that can be passed
     and override values already specified in the templates
   - existingUpgradingChildDef: the definition of the upgrading child as an unstructured object.
+  - ownerReference: reference to the upgrading child this AnalysisRun is associated with - ensures cleanup
   - client: the client used for interacting with the Kubernetes API.
 
 Returns:
@@ -124,7 +125,7 @@ func CreateAnalysisRun(ctx context.Context, analysis apiv1.Analysis, existingUpg
 		return err
 	}
 
-	// set ownerReference to guarantee AnalysisRun deletion when rollout is cleaned up
+	// set ownerReference to guarantee AnalysisRun deletion when owner is cleaned up
 	analysisRun.SetOwnerReferences([]metav1.OwnerReference{ownerReference})
 	if err = client.Create(ctx, analysisRun); err != nil {
 		return err
--- a/internal/controller/progressive/analysis.go
+++ b/internal/controller/progressive/analysis.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


juliev0 (changes requested):
it needs to be owned by the MonoVertex/Pipeline, not the rollout. Otherwise, we would be accumulating these forever until the rollout gets deleted.

dpadhiar:
> it needs to be owned by the MonoVertex/Pipeline, not the rollout. Otherwise, we would be accumulating these forever until the rollout gets deleted.

Ack, just changed that now. Will do a quick test.

dpadhiar:
> > it needs to be owned by the MonoVertex/Pipeline, not the rollout. Otherwise, we would be accumulating these forever until the rollout gets deleted.
> 
> Ack, just changed that now. Will do a quick test.

Quickly tested, new analysisrun was created and previous was deleted after two successive progressive upgrades. Changing PR back to ready to review.

internal/controller/progressive/analysis.go
```diff
@@ -123,6 +124,8 @@ func CreateAnalysisRun(ctx context.Context, analysis apiv1.Analysis, existingUpg
 		return err
 	}
 
+	// set ownerReference to guarantee AnalysisRun deletion when rollout is cleaned up
```
juliev0, line 127:
```suggestion
	// set ownerReference to guarantee AnalysisRun deletion when owner is cleaned up
```

internal/controller/progressive/analysis.go
```diff
@@ -95,7 +96,7 @@ Parameters:
 Returns:
   - An error if any issues occur during processing.
 */
-func CreateAnalysisRun(ctx context.Context, analysis apiv1.Analysis, existingUpgradingChildDef *unstructured.Unstructured, client client.Client) error {
+func CreateAnalysisRun(ctx context.Context, analysis apiv1.Analysis, existingUpgradingChildDef *unstructured.Unstructured, ownerReference metav1.OwnerReference, client client.Client) error {
```
juliev0, line 99: need to update the parameters described above to include the new one

juliev0 (approved):

PR #695, opened 2025-04-10
feat: AnalysisRun implementation - unit tests 

### LINKED ISSUE #682: Unit testing for AnalysisRuns
# Summary

What change needs making?

Add unit test cases for progressive upgrades that have an AnalysisStatus set to simulate having an AnalysisRun be created and ran during progressive upgrades.

# Use Cases

When would you use this?

Testing progressive upgrade code changes.

---

dpadhiar:
Fixes #682 

### Modifications

Adds unit testing for our AnalysisRun implementation. 

Tests that our progressive upgrades are evaluated correctly when an AnalysisTemplate/Run are present and the AnalysisStatus is set (we set this manually in the test since our unit test cannot reconcile the AnalysisRuns itself).

### Verification

Running our new unit tests locally and on the CI.

n/a


internal/controller/monovertexrollout/monovertexrollout_controller_test.go
```diff
@@ -366,6 +508,69 @@ func Test_processExistingMonoVertex_Progressive(t *testing.T) {
...
+						Name:                ctlrcommon.DefaultTestMonoVertexRolloutName + "-1",
+						AssessmentStartTime: &metav1.Time{Time: time.Now().Add(-1 * time.Minute)},
+						AssessmentEndTime:   &metav1.Time{Time: time.Now().Add(-30 * time.Second)},
+						AssessmentResult:    apiv1.AssessmentResultFailure,
+					},
+					Analysis: apiv1.AnalysisStatus{
+						AnalysisRunName: ctlrcommon.DefaultTestMonoVertexRolloutName + "-1",
+						StartTime:       &metav1.Time{Time: time.Now().Add(-45 * time.Second)},
+						EndTime:         &metav1.Time{Time: time.Now().Add(-40 * time.Second)},
+						Phase:           argorolloutsv1.AnalysisPhaseFailed,
```
juliev0, line 561: So, is the main goal of this test case to show that if the AnalysisRun has failed, then we should still be in "Pending" with `expectedInProgressStrategy: Progressive`? Do you know if this is testing something beyond what the "Progressive deployment failed" test case above is testing?

juliev0: My thought is whether the `AssessmentResult: Failure` part that's common to both is really the determining factor for whether we arrive in "Pending" with `expectedInProgressStrategy: Progressive`? Do you think there's any added value to having this one as well? 

I hate to ask since I know you put effort into this as well as the other test case, but I want to make sure that it's worth maintaining both of them before we add them.

juliev0 (commented):
I guess what's not being tested is the deployment of the AnalysisRun in the first place. Theoretically there could be something which checks that the right AnalysisRun got deployed with the required templating. Although, if we want to plan to just do that in the e2e tests, that may be enough

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
dpadhiar: Good callout, how about if we try this initial upgrading status:
```
UpgradingChildStatus: apiv1.UpgradingChildStatus{
	Name:                ctlrcommon.DefaultTestMonoVertexRolloutName + "-1",
	AssessmentStartTime: &metav1.Time{Time: time.Now().Add(-1 * time.Minute)},
	// AssessmentEndTime:   &metav1.Time{Time: time.Now().Add(-30 * time.Second)},
	AssessmentResult: apiv1.AssessmentResultUnknown,
},
```

This should end with the same result of Pending and Progressive still in progress.

juliev0: yes, you could do that

This actually makes me wonder if we should be testing the Condition `ProgressiveUpgradeSucceeded` in this test: https://github.com/numaproj/numaplane/blob/main/pkg/apis/numaplane/v1alpha1/status_types.go#L59

dpadhiar:
> I guess what's not being tested is the deployment of the AnalysisRun in the first place. Theoretically there could be something which checks that the right AnalysisRun got deployed with the required templating. Although, if we want to plan to just do that in the e2e tests, that may be enough

I could create an `analysis_test.go` file to test the functions that create the resulting AnalysisRun for us for the unit tests. 

These tests do create the resulting AnalysisRun if we don't create them manually like we are, they just are not reconciled and make it so the tests will not conclude.

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
dpadhiar: I see, should we add an expectedCondition to the test checks then?

juliev0:
> These tests do create the resulting AnalysisRun if we don't create them manually like we are, they just are not reconciled and make it so the tests will not conclude.

What do you mean that the test would not conclude?

For a unit test, you should only be testing as part of one functional call, which is of course just one reconciliation. The test could do a Kubernetes Get to get the AnalysisRun, verify it's there, read it to make sure it has the right content, and the test could also delete it.

juliev0:
Anyway, feel free to do as a separate PR if you like

dpadhiar:
> > These tests do create the resulting AnalysisRun if we don't create them manually like we are, they just are not reconciled and make it so the tests will not conclude.
> 
> What do you mean that the test would not conclude?
> 
> For a unit test, you should only be testing as part of one functional call, which is of course just one reconciliation. The test could do a Kubernetes Get to get the AnalysisRun, verify it's there, read it to make sure it has the right content, and the test could also delete it.

By "not conclude" I meant it would always reconcile as Pending and still in Progressive strategy since the AnalysisRun will never get reconciled.

juliev0:
> > > These tests do create the resulting AnalysisRun if we don't create them manually like we are, they just are not reconciled and make it so the tests will not conclude.
> > 
> > 
> > What do you mean that the test would not conclude?
> > For a unit test, you should only be testing as part of one functional call, which is of course just one reconciliation. The test could do a Kubernetes Get to get the AnalysisRun, verify it's there, read it to make sure it has the right content, and the test could also delete it.
> 
> By "not conclude" I meant it would the controller would reconcile the rollout as Pending and still in Progressive strategy since the AnalysisRun will never get reconciled since we aren't using the Argo client.

Yes, I think that would be worthwhile to have a test which checks that the AnalysisRun is generated correctly.

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
```diff
@@ -256,6 +329,74 @@ func Test_processExistingMonoVertex_Progressive(t *testing.T) {
...
+			initialRolloutPhase:       apiv1.PhasePending,
+			initialRolloutNameCount:   2,
+			initialInProgressStrategy: &progressiveUpgradeStrategy,
+			initialUpgradingChildStatus: &apiv1.UpgradingMonoVertexStatus{
+				UpgradingPipelineTypeStatus: apiv1.UpgradingPipelineTypeStatus{
+					UpgradingChildStatus: apiv1.UpgradingChildStatus{
+						Name:                ctlrcommon.DefaultTestMonoVertexRolloutName + "-1",
+						AssessmentStartTime: &metav1.Time{Time: time.Now().Add(-1 * time.Minute)},
+						AssessmentEndTime:   &metav1.Time{Time: time.Now().Add(-30 * time.Second)},
+						AssessmentResult:    apiv1.AssessmentResultSuccess,
```
juliev0, line 374: if you set this one to Unknown, does it take the MonoVertexRollout out of pending/in-progress-strategy?

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
juliev0: I think that would be great to test that it failed and not just that it's in Pending if possible

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
dpadhiar: Yes, can make that change as well.

internal/controller/monovertexrollout/monovertexrollout_controller_test.go
dpadhiar: Added now.

juliev0 (approved):

PR #696, opened 2025-04-12
fix: replace all uses of patchPipelineVertices() with applyScaleValuesToLivePipeline()

juliev0:
### Modifications

This fixes an issue I saw. Previously, we were calling [patchPipelineVertices()](https://github.com/numaproj/numaplane/blob/v0.19.0/internal/controller/pipelinerollout/pipelinerollout_progressive.go#L648), which actually patches the Vertices in a pipeline with all of the fields of those vertices.

This was normally fine, except say a field's value were originally set to null. Our patch would also include `{"field": null}`. This would have the effect of removing the field altogether. That would be fine, except that it triggers the [ChildNeedsUpdating() ](https://github.com/numaproj/numaplane/blob/v0.19.0/internal/controller/pipelinerollout/pipelinerollout_controller.go#L1008)function to return true even though there's no real difference. 

Therefore, I replaced all of the calls to `patchPipelineVertices()` with the function `applyScaleValuesToLivePipeline()`, which only patches the `min` and `max` fields and doesn't touch anything else.

### Verification

Did a considerable amount of testing with PipelineRollout:
Tested failure and success cases and observed the scale values of both promoted and upgrading pipelines.

n/a


internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -469,10 +472,13 @@ func scaleDownPromotedMonoVertex(
 		return true, fmt.Errorf("cannot extract the scale min and max values from the promoted monovertex: %w", err)
 	}
 
-	newMin, newMax, err := progressive.CalculateScaleMinMaxValues(promotedMonoVertexDef.Object, int(currentPodsCount), []string{"spec", "scale", "min"})
+	/*newMin, newMax, err := progressive.CalculateScaleMinMaxValues(promotedMonoVertexDef.Object, int(currentPodsCount), []string{"spec", "scale", "min"})
 	if err != nil {
 		return true, fmt.Errorf("cannot calculate the scale min and max values: %+w", err)
-	}
+	}*/
+	scaleTo := progressive.CalculateScaleMinMaxValues(int(currentPodsCount))
```
juliev0, line 479: according to the new design, we set min and max the same on the promoted child

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -315,6 +315,9 @@ func scaleDownUpgradingMonoVertex(
 		// which is necessary when performing the health check for "ready replicas >= desired replicas"
 		// 2. that the total number of Pods (between the 2 monovertices) before and during upgrade remains the same
 		upgradingChildScaleTo := scaleValue.Initial - scaleValue.ScaleTo
+		if upgradingChildScaleTo <= 0 { // if for some reason the Initial value was 0, we don't want to set our Pods to 0
+			upgradingChildScaleTo = 1
+		}
```
juliev0, line 320: make sure we don't accidentally scale a new upgrading vertex to 0 Pods just because of the calculation - we need to have at least 1 Pod to test it

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -403,26 +406,13 @@ func scalePipelineVerticesToZero(
...
-				}
-			}
+		zero := int64(0)
+		for i := range vertexScaleDefinitions {
+			vertexScaleDefinitions[i].ScaleDefinition = &apiv1.ScaleDefinition{Min: &zero, Max: &zero}
 		}
 
 		numaLogger.Debug("Scaling down all vertices to 0 Pods")
-		if err := patchPipelineVertices(ctx, pipelineDef, vertices, c); err != nil {
+		if err := applyScaleValuesToLivePipeline(ctx, pipelineDef, vertexScaleDefinitions, c); err != nil {
```
juliev0, line 415: this call just patches in the new min and max as opposed to patching all fields of the vertices

internal/controller/progressive/progressive_test.go
```diff
@@ -335,71 +335,6 @@ var forcePromoteMonoVertexRollout = &apiv1.MonoVertexRollout{
 	},
 }
 
-func Test_CalculateScaleMinMaxValues(t *testing.T) {
```
juliev0, line 338: the function is now a trivial one line function so unit test is not needed

dpadhiar (approved):

PR #697, opened 2025-04-14
fix: Pipeline e2e checks need to be updated based on latest code

juliev0:
### Modifications

Since we incorporated the changes to Pipeline recently described [here](https://github.com/numaproj/numaplane/pull/689), the e2e test needed to be updated. 

For one, "Source Vertex" is no longer handled differently from other vertices.

Also, the Upgrading Pipeline now changes its scale values, so that needed to be tested as well.  

In addition to the CI, one minor modification was done to the MonoVertexRollout progressive code, which is described inline. This is related to a change recently made to Pipeline that didn't get reflected into MonoVertex.

### Verification

Running e2e tests through CI. 

N/A


internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -315,6 +315,9 @@ func scaleDownUpgradingMonoVertex(
 		// which is necessary when performing the health check for "ready replicas >= desired replicas"
 		// 2. that the total number of Pods (between the 2 monovertices) before and during upgrade remains the same
 		upgradingChildScaleTo := scaleValue.Initial - scaleValue.ScaleTo
+		if upgradingChildScaleTo <= 0 { // if for some reason the Initial value was 0, we don't want to set our Pods to 0
+			upgradingChildScaleTo = 1
+		}
```
juliev0, line 320: this was a fix made to Pipeline that I'm now incorporating into Monovertex as well.

tests/e2e/common.go
```diff
@@ -147,33 +148,35 @@ func VerifyVerticesPodsRunning(namespace, rolloutChildName string, specVertices
...
+		switch component {
+		case ComponentVertex:
+			vtxLabelSelector = fmt.Sprintf("app.kubernetes.io/name=%s-%s", rolloutChildName, vtx.Name)
+		case ComponentMonoVertex:
+			vtxLabelSelector = fmt.Sprintf("app.kubernetes.io/name=%s", rolloutChildName)
+		}
+
+		labelSelector := fmt.Sprintf("%s,%s", baseLabelSelector, vtxLabelSelector)
+
+		CheckEventually(fmt.Sprintf("verifying that the correct number of Pods is running %s (%s): between min=%d and max=%d with label selector %s", msg, vtx.Name, *min, *max, labelSelector), func() bool {
```
juliev0, line 179: created a separate `CheckEventually` for each Vertex instead of one for all, so that we can print exactly what we're looking for each time around, and better pinpoint the issue if it fails

tests/e2e/functional-pipeline-e2e/pipeline_test.go
```diff
@@ -266,24 +257,24 @@ var _ = Describe("Functional e2e:", Serial, func() {
...
 		updatedISBServiceSpec := isbServiceSpec
 		updatedISBServiceSpec.JetStream.Version = updatedJetstreamVersion
 
-		expectedPipelineTypeProgressiveStatusInProgress, expectedPipelineTypeProgressiveStatusOnDone := MakeExpectedPipelineTypeProgressiveStatus(
+		/*expectedPipelineTypeProgressiveStatusInProgress, expectedPipelineTypeProgressiveStatusOnDone := MakeExpectedPipelineTypeProgressiveStatus(
 			GetInstanceName(pipelineRolloutName, 1), GetInstanceName(pipelineRolloutName, 2), sourceVertexName,
 			sourceVertexScaleTo, sourceVertexScaleTo,
 			fmt.Sprintf("{\"max\":%d,\"min\":%d}", sourceVertexScaleMax, sourceVertexScaleMin),
 			apiv1.AssessmentResultUnknown, apiv1.AssessmentResultSuccess,
-		)
```
juliev0, line 282: I don't think it makes sense for this generic code, which applies to all strategies, to be aware of Progressive status.

dpadhiar (approved):

PR #698, opened 2025-04-15
feat:  adding label to upgrading child will force promotion during progressive upgrade

dpadhiar:
### Modifications

Currently we have a force-promote field on the parent rollout which can be set to force promote a child even if the progressive upgrade is failed. This may cause issues for later progressive upgrades however if the user does not remember to remove or set this field to false for their next upgrades.

To remedy this, we can check for a `force-promote` label on the upgrading child during the assessment and force a success that way. Thus when the current upgrade happens, this child becomes the currently promoted and the label will be ignored. 

### Verification

Testing locally with failed progressive upgrade that passes after setting the label on the upgrading child.

n/a

### DIFF
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -278,8 +278,14 @@ func processUpgradingChild(
 	}
 	childStatus = rolloutObject.GetUpgradingChildStatus()
 
-	// check for Force Promote set in Progressive strategy to force success logic
-	if rolloutObject.GetProgressiveStrategy().ForcePromote {
+	forcePromote := false
+	_, ok := existingUpgradingChildDef.GetLabels()[common.LabelKeyForcePromote]
+	if ok {
+		forcePromote = true
+	}
+
+	// check for Force Promote set in Progressive strategy to force success logic OR if promoted child has force-promote label
+	if rolloutObject.GetProgressiveStrategy().ForcePromote || forcePromote {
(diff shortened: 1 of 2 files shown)


juliev0:
> To remedy this, we can check for a force-promote label on the upgrading child during the assessment and force a success that way. Thus when the next upgrade happens, this child becomes the currently promoted and the label will be ignored.

This should say "for the current upgrade" as opposed to "the next upgrade"

internal/controller/progressive/progressive.go
```diff
@@ -278,8 +278,14 @@ func processUpgradingChild(
...
 
-	// check for Force Promote set in Progressive strategy to force success logic
-	if rolloutObject.GetProgressiveStrategy().ForcePromote {
+	forcePromote := false
+	_, ok := existingUpgradingChildDef.GetLabels()[common.LabelKeyForcePromote]
+	if ok {
+		forcePromote = true
+	}
+
+	// check for Force Promote set in Progressive strategy to force success logic OR if promoted child has force-promote label
```
juliev0, line 287: why does it say "if promoted child" rather than "if upgrading child"? did you test it with promoted child or upgrading child?

juliev0:
> Testing locally with failed progressive upgrade that passes after setting the label on the rollout.

Do you mean "after setting the label on the upgrading child"?

juliev0 (approved):

PR #702, opened 2025-04-17
fix: ensure we perform post-upgrade functions idempotently

juliev0:
### Modifications

I realized that it's possible in the current code to have this problem:
1. get some error when calling [controller.ProcessUpgradingChildPostUpgrade()](https://github.com/numaproj/numaplane/blob/v0.20.0/internal/controller/progressive/progressive.go#L580) 
2. Re-reconcile and never have an opportunity to try that function again

That is because the function this code lives in - `startUpgradeProcess()` - is only called when the Upgrading child hasn't been created yet. 

Therefore, we need to maintain some other state which indicates if we have completed the initialization process or not. That is why I have added a new field called `InitializationComplete` in the UpgradingChildStatus. Once this is set to true, we know that we don't need to perform initialization anymore. I have also divided the `startUpgradeProcess()` into two separate functions: 
1. `startUpgradeProcess()` does the initialization up until creating the upgrading child
2. `startPostUpgradeProcess()` does the initialization that comes after creating the upgrading child

This PR also adds an additional function - `ProcessPromotedChildPostUpgrade()` - which will be used for a later issue even though none of the Controllers currently implement it yet.

### Verification

did various success and fail cases of the various Rollouts

The current rollouts will not have the `InitializationComplete` value set, which will be treated as false. We will only execute anything in the code if `!childStatus.InitializationComplete && childStatus.AssessmentResult == apiv1.AssessmentResultUnknown`, so this would only be the rare rollout which is in the middle of a progressive upgrade, which isn't too likely. 

Currently, the only one of the rollouts that even implements any function in there is isbservicerollout creating a PDB for the isbsvc (in which case it would just error trying to create the pdb that already exists).


internal/controller/progressive/progressive.go
```diff
@@ -154,6 +158,22 @@ func ProcessResource(
...
+
+	// if the Upgrading child status exists but indicates that we aren't done with upgrade process, then do postupgrade process
+	if !childStatus.InitializationComplete {
+		needsRequeue, err := startPostUpgradeProcess(ctx, rolloutObject, existingPromotedChild, currentUpgradingChildDef, controller, c)
+		if needsRequeue {
+			return false, common.DefaultRequeueDelay, err
+		} else {
+			return false, 0, err
+		}
+	}
```
juliev0, line 175: see that `startUpgradeProcess()` is called above if Upgrading child doesn't exist, and `startPostUpgradeProcess()` called here if `InitializationComplete==false`

internal/controller/progressive/progressive.go
```diff
@@ -181,6 +201,31 @@ func makeUpgradingObjectDefinition(ctx context.Context, rolloutObject Progressiv
...
+
+		err := rolloutObject.ResetUpgradingChildStatus(currentUpgradingChildDef)
+		if err != nil {
+			return nil, fmt.Errorf("processing upgrading child, failed to reset the upgrading child status for child %s/%s", currentUpgradingChildDef.GetNamespace(), currentUpgradingChildDef.GetName())
+		}
+	} else {
+		numaLogger.WithValues("childStatus", *childStatus).Debug("live upgrading child previously set")
+	}
+	return rolloutObject.GetUpgradingChildStatus(), nil
+}
```
juliev0, line 227: moved this logic into its own function

internal/controller/progressive/progressive_test.go
```diff
@@ -107,39 +111,17 @@ func Test_processUpgradingChild(t *testing.T) {
...
-					UpgradingPipelineTypeStatus: apiv1.UpgradingPipelineTypeStatus{
-						UpgradingChildStatus: apiv1.UpgradingChildStatus{Name: "test"}},
-				},
-				nil),
-			//setRolloutObjectChildStatus(defaultMonoVertexRollout.DeepCopy(), &apiv1.UpgradingChildStatus{Name: "test"}, &apiv1.PromotedChildStatus{}),
-			existingUpgradingChildDef: createMonoVertex("test-1"),
-			expectedDone:              false,
-			expectedRequeueDelay:      assessmentSchedule.Interval,
-			expectedError:             nil,
-		},
```
juliev0, line 131: these test cases no longer make sense - this function now assumes that the UpgradingChildStatus is the correct one

dpadhiar (approved):

PR #709, opened 2025-04-19
chore: log errors in numaplane that we issue events for

juliev0:
### Modifications

If we are going to create an event for an error, we should log it in the numaplane logs as well.

(BTW, I had trouble finding the event in splunk.)

### Verification

This is just a new log message.

n/a


chandankumar4 (approved):
Lgtm

PR #711, opened 2025-04-21
fix: Removed foreground deletion and updated deletion process for ISBSVC

### LINKED ISSUE #704: Move/modify check for dependent resources on deletion
Instead of having a finalizer on `NumaflowController` to check whether the dependent resources - `pipelines`, `isbsvc`, `monovertices` - still exist, @whynowy thinks it would be a better approach to have the finalizer on `NumaflowControllerRollout` to check for whether `PipelineRollout`, `ISBServiceRollout`, and `NumaflowControllerRollout` still exist.

The benefit is that if somebody manually deletes their `NumaflowController`, it will simply get deleted and then auto-healed by NumaflowControllerRollout reconciler. 

In either case, though, if somebody deletes their `NumaflowControllerRollout`, it will then have a deletionTimestamp, which cannot be undone, and there's nothing anyone can do except delete everything and recreate anything in that case. 

---

chandankumar4:
Partly Fixes #704 

### Modifications
- Removed the foreground deletion
- Delete the ISBService child object first before removing finalizer whenever deletion request is received for ISBServiceRollout.

### Verification

- Verified in local k8s cluster

### DIFF
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -980,6 +980,7 @@ func getLiveISBServiceRollout(ctx context.Context, name, namespace string) (*api
 }
 
 // listAndDeleteChildISBServices lists all child ISBServices and deletes them
+// return true if we need to requeue
 func (r *ISBServiceRolloutReconciler) listAndDeleteChildISBServices(ctx context.Context, isbServiceRollout *apiv1.ISBServiceRollout) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
 	isbServiceList, err := kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion, numaflowv1.ISBGroupVersionResource.Resource,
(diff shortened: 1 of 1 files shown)


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -984,3 +978,24 @@ func getLiveISBServiceRollout(ctx context.Context, name, namespace string) (*api
...
+
+// listAndDeleteChildISBServices lists all child ISBServices and deletes them
+func (r *ISBServiceRolloutReconciler) listAndDeleteChildISBServices(ctx context.Context, isbServiceRollout *apiv1.ISBServiceRollout) (bool, error) {
+	numaLogger := logger.FromContext(ctx)
+	isbServiceList, err := kubernetes.ListLiveResource(ctx, common.NumaflowAPIGroup, common.NumaflowAPIVersion, numaflowv1.ISBGroupVersionResource.Resource,
+		isbServiceRollout.Namespace, fmt.Sprintf("%s=%s", common.LabelKeyParentRollout, isbServiceRollout.Name), "")
+	if err != nil {
+		return false, err
+	}
+	if len(isbServiceList.Items) > 0 {
```
juliev0, line 990:
```suggestion
	if isbServiceList != nil && len(isbServiceList.Items) > 0 {
```

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -984,3 +978,24 @@ func getLiveISBServiceRollout(ctx context.Context, name, namespace string) (*api
 
 	return isbServiceRollout, err
 }
+
+// listAndDeleteChildISBServices lists all child ISBServices and deletes them
```
juliev0, line 982:
```suggestion
// listAndDeleteChildISBServices lists all child ISBServices and deletes them
// return true if we need to requeue
```

juliev0 (approved):

PR #716, opened 2025-04-23
feat: improve templating by evaluating rollout specs with hyphenated arguments

dpadhiar:
### Modifications

Instead of having the kludgey approach of replacing the hyphenated arguments in a rollout spec before evaluating it as a template, we can utilize the `fasttemplate` package which can evaluate these arguments without error. 

### Verification

Testing multiple Pipeline and MonoVertex rollouts after progressive upgrades to see the values in the child resources are updated properly.

n/a

### DIFF
--- a/internal/controller/common/rollout_controller.go
+++ b/internal/controller/common/rollout_controller.go
@@ -259,8 +259,7 @@ func GetChildName(ctx context.Context, rolloutObject RolloutObject, controller R
 	}
 }
 
-// resolves templated definitions of a child resource with arguments for name and namespace
-// for resources that reference each other, arguments will be dynamically updated with current name of child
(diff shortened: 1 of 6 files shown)


internal/controller/common/rollout_controller.go
```diff
@@ -262,31 +261,24 @@ func GetChildName(ctx context.Context, rolloutObject RolloutObject, controller R
 
 // resolves templated definitions of a child resource with arguments for name and namespace
```
juliev0, line 262: this comment is too specific for what this function is really doing - it's really just resolving a templated definition of anything with any arguments

dpadhiar: Changed to be more generic.

juliev0 (approved):

PR #717, opened 2025-04-23
fix: post-failure code was referring to the old upgrading child rather than the new one

juliev0:
### Modifications

This fixes a bug in which after failure case, the previous "upgrading child" was being used to perform `ProcessUpgradingChildPostUpgrade()` rather than the new one.

Thankfully, only ISBServiceRollout implements anything that happens in `ProcessUpgradingChildPostUpgrade()` at least, but it was operating on the wrong child.

### Verification

Updated ISBServiceRollout to cause failure. Then updated a second time. Observed that the PDB created was for the new isbsvc.

n/a

### DIFF
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -380,9 +380,12 @@ func processUpgradingChild(
 			return false, 0, err
 		}
 
-		// if so, create a new upgrading one and mark the existing one for garbage collection
+		///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+		// Replace existing Upgrading child with new one and mark the existing one for garbage collection
+		///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 		if needsUpdating {
-			_, needRequeue, err := startUpgradeProcess(ctx, rolloutObject, existingPromotedChildDef, controller, c)
+			needRequeue := false
+			newUpgradingChildDef, needRequeue, err = startUpgradeProcess(ctx, rolloutObject, existingPromotedChildDef, controller, c)
 			if err != nil {
 				return false, 0, err
 			}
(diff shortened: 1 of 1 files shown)


internal/controller/progressive/progressive.go
```diff
@@ -380,9 +380,12 @@ func processUpgradingChild(
...
 		}
 
-		// if so, create a new upgrading one and mark the existing one for garbage collection
+		///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+		// Replace existing Upgrading child with new one and mark the existing one for garbage collection
+		///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 		if needsUpdating {
-			_, needRequeue, err := startUpgradeProcess(ctx, rolloutObject, existingPromotedChildDef, controller, c)
+			needRequeue := false
+			newUpgradingChildDef, needRequeue, err = startUpgradeProcess(ctx, rolloutObject, existingPromotedChildDef, controller, c)
```
juliev0, line 388: set `newUpgradingChildDef` to the new one we just created so we can use it below

internal/controller/progressive/progressive.go
```diff
@@ -392,24 +395,27 @@ func processUpgradingChild(
...
 			}
 		}
 		childStatus = rolloutObject.GetUpgradingChildStatus()
 
-		// do post-upgrade process if we haven't (check that AssessmentResult is not set just in case)
+		// After creating the new Upgradng child, do post-upgrade process (check that AssessmentResult is not set just in case) and return
 		if !childStatus.InitializationComplete && childStatus.AssessmentResult == apiv1.AssessmentResultUnknown {
 
-			needsRequeue, err := startPostUpgradeProcess(ctx, rolloutObject, existingPromotedChildDef, existingUpgradingChildDef, controller, c)
+			needsRequeue, err := startPostUpgradeProcess(ctx, rolloutObject, existingPromotedChildDef, newUpgradingChildDef, controller, c)
```
juliev0, line 409: this function needs to operate on the newly created Upgrading child. We were accidentally operating on the existing Upgrading child

internal/controller/progressive/progressive.go
```diff
@@ -653,7 +659,7 @@ func startPostUpgradeProcess(
 	if requeue {
 		return true, nil
 	}
-	requeue, err = controller.ProcessUpgradingChildPostUpgrade(ctx, rolloutObject, existingUpgradingChild, c)
+	requeue, err = controller.ProcessUpgradingChildPostUpgrade(ctx, rolloutObject, newUpgradingChild, c)
```
juliev0, line 662: renaming to clarify that this is the new one that we created

PR #723, opened 2025-04-25
feat: add permittedRiders to numaplane ctrl cm

### LINKED ISSUE #719: Disallow Riders that aren't whitelisted
# Summary

Let's add something to our global config like `whitelistedRiders` or `permittedRiders`.

I'm thinking maybe in the `GetDesiredRiders()` function for each rollout we could call a function in either rolloutController.go or rider.go which compares the rider(s?) passed in to determine if they are allowed. Return an error if not.

We can set this value to include the VPA and ConfigMap in our test config files. Maybe for now don't bother setting it in the default config yaml for Numaplane.

---

dpadhiar:
Fixes #719 

### Modifications

Adds the `permittedRiders` field to the Numaplane controller configmap. 

If a Kind is specified here, it will allow Riders of this Kind to created for MonoVertex/Pipeline/ISBService rollouts. If a non-specified Kind is used for a submitted Rider, the controller will return an error.

### Verification

Created a MonoVertexRollout locally with VPA and ConfigMap riders to test good case. Added Secret rider and checked both that Secret was not created and the controller was returning an error.

n/a

### DIFF
--- a/internal/controller/common/riders/rider.go
+++ b/internal/controller/common/riders/rider.go
@@ -8,12 +8,14 @@ import (
 	"fmt"
 
 	"github.com/numaproj/numaplane/internal/common"
+	"github.com/numaproj/numaplane/internal/controller/config"
 	"github.com/numaproj/numaplane/internal/util/kubernetes"
 	"github.com/numaproj/numaplane/internal/util/logger"
 	apiv1 "github.com/numaproj/numaplane/pkg/apis/numaplane/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8stypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -171,4 +173,22 @@ func GetRidersFromK8S(ctx context.Context, namespace string, ridersList []apiv1.
 
 	}
 	return existingRiders, nil
+}
+
+func VerifyRiders(riders []Rider) error {
+
+	// get global configmap
+	globalConfig, err := config.GetConfigManagerInstance().GetConfig()
+	if err != nil {
+		return err
+	}
+
+	// check that each rider is a permitted Kind set in the Numplane controller config
+	for _, rider := range riders {
+		if !slices.Contains(globalConfig.PermittedRiders, rider.Definition.GetKind()) {
+			return fmt.Errorf("Rider %s of kind %s is not a permitted Kind", rider.Definition.GetName(), rider.Definition.GetKind())
+		}
+	}
+
+	return nil
 }
--- a/tests/manifests/default/config.yaml
+++ b/tests/manifests/default/config.yaml
@@ -7,3 +7,6 @@ group=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=
 group=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role"
 defaultUpgradeStrategy: "pause-and-drain"
 childStatusAssessmentSchedule: "120,60,10"
+permittedRiders:
+- "VerticalPodAutoscaler"
+- "ConfigMap"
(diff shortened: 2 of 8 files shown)


tests/manifests/default/config.yaml
```diff
@@ -7,3 +7,6 @@ group=,kind=ConfigMap;group=,kind=ServiceAccount;group=,kind=Secret;group=,kind=
 group=rbac.authorization.k8s.io,kind=RoleBinding;group=rbac.authorization.k8s.io,kind=Role"
 defaultUpgradeStrategy: "pause-and-drain"
 childStatusAssessmentSchedule: "120,60,10"
+permittedRiders:
```
juliev0, line 10: can we format this the same as `includedResources` above for similarity? It needs group, not just kind.

internal/sync/cache.go
```diff
@@ -389,13 +389,15 @@ func (c *liveStateCache) loadCacheSettings() (*cacheSettings, error) {
 	return &cacheSettings{clusterSettings, ignoreResourceUpdatesEnabled}, nil
 }
 
-// parseResourceFilter parse the given rules to generate the
+// TODO: move to shared package (util?) since we also use this for riders
```
dpadhiar, line 392: Since we're formatting `permittedRiders` the same way as `includedResources` we should also use the same functions and structs to evaluate them. In that case, we could move these all to a shared package.

juliev0: I guess you will need to move the `ResourceType` struct there too

internal/controller/common/riders/rider.go
```diff
@@ -172,3 +174,31 @@ func GetRidersFromK8S(ctx context.Context, namespace string, ridersList []apiv1.
...
+func VerifyRiders(riders []Rider) error {
+
+	// get global configmap
+	globalConfig, err := config.GetConfigManagerInstance().GetConfig()
+	if err != nil {
+		return err
+	}
+
+	// parse permitted Riders and create resource filter
+	riderRules, err := sync.ParseResourceFilter(globalConfig.PermittedRiders)
```
juliev0, line 187: not related to your code, but I noticed this function returns a pointer - I wonder if we can change that to just return an array slice instead? then we don't need to check for nil over here

internal/sync/cache.go
```diff
@@ -471,12 +473,17 @@ type ResourceType struct {
...
 		if resource.Kind == "" {
+			// When Kind is empty, we only check if Group matches
 			if group == resource.Group {
 				return false
 			}
+		} else if resource.Group == "" {
+			// When Group is empty, we only check if Kind matches
+			if group == resource.Kind {
+				return false
+			}
```
juliev0, line 486: I wonder if we need this. Maybe we shouldn't encourage users to only specify `Kind` and not `Group` since the reason `Group` exists is to serve as a "namespace" which can distinguish Kinds that happen to have the same name.

juliev0: Did you have an issue of specifying ConfigMap that it didn't work? I think that may have `group==""`?

dpadhiar: I was previously seeing an error adding `Secret` to the controller configmap but it was a formatting issue instead of a code error. Just tested this out and it is back to working like we intend. In this case, I'll remove this extra clause I added.

internal/controller/common/riders/rider.go
dpadhiar: Was able to make this change to have this function no longer return a pointer.

internal/controller/common/riders/rider.go
```diff
@@ -172,3 +174,31 @@ func GetRidersFromK8S(ctx context.Context, namespace string, ridersList []apiv1.
...
+	}
+	resourcesFilter := &sync.ResourceFilter{
+		IncludedResources: *riderRules,
+	}
+
+	// check that each rider is a permitted Group/Kind set in the Numplane controller config
+	for _, rider := range riders {
+		gvk := rider.Definition.GroupVersionKind()
+		if resourcesFilter.IsExcludedResource(gvk.Group, gvk.Kind, "") {
+			return fmt.Errorf("Rider %s of %s/%s is not a permitted Kind", rider.Definition.GetName(), gvk.Group, gvk.Kind)
```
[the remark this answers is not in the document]
dpadhiar: Changed `VerifyRiders` to not return an error however in `GetDesiredRiders` we still need to return an error. Should we instead return an empty list of desiredRiders and no error?

juliev0 (approved):

PR #724, opened 2025-04-26
fix: Removing existing finalizer for monovertex and controller

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/704

### Modifications
This change will fix the issue regarding finalizer left on resource after deletion.
- Removing existing finalizer for monovertex and controller

### Verification
- Verified on local k8s

### DIFF
--- a/internal/controller/monovertexrollout/monovertexrollout_controller.go
+++ b/internal/controller/monovertexrollout/monovertexrollout_controller.go
@@ -19,7 +19,6 @@ package monovertexrollout
 import (
 	"context"
 	"fmt"
-
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


juliev0 (approved):

juliev0 (commented):
will wait until you look into the CI failure, thanks

juliev0 (changes requested):
will wait until you look into the CI failure, thanks

juliev0 (approved):
okay, I noticed that your test failure also happened here for a separate PR: https://github.com/numaproj/numaplane/actions/runs/14697017565/job/41240051750?pr=727

let me look into that failure

juliev0:
> okay, I noticed that your test failure also happened here for a separate PR: https://github.com/numaproj/numaplane/actions/runs/14697017565/job/41240051750?pr=727
> 
> let me look into that failure

If you download at the MonoVertexRollout from the CI, it seems that for some reason the new MonoVertexRollout is created before the previous MonoVertexRollout has actually been deleted. (You can tell because the progressiveStatus from the old one is still showing.) It's pretty weird because the test already checked that the MonoVertexRollout was no longer there [here](https://github.com/numaproj/numaplane/blob/main/tests/e2e/monovertex.go#L333-L342).

PR #726, opened 2025-04-27
fix: log line formatting

juliev0:
### Modifications

I noticed some log lines which were accidentally written using `Info()` rather than `Infof()`.

### Verification

N/A

N/A


chandankumar4 (approved):
Thanks

PR #730, opened 2025-04-28
go back to background deletion for pipeline in e2e

juliev0:
### Modifications

The ppnd test started intermittently failing due to a race condition that occurs when Numaflow's Create Buffers/Buckets Job (when pipeline is created) runs at the same time as its Delete Buffers/Buckets Job (when previous version of the same pipeline is deleted), and because of that the needed buffers/buckets don't exist.

Somehow updating the e2e test to use Foreground deletion caused the timing to change.

So, I'm putting it back to background deletion. 

### Verification

CI

N/A

### DIFF
--- a/tests/e2e/pipeline.go
+++ b/tests/e2e/pipeline.go
@@ -423,8 +423,12 @@ func createPipelineRolloutSpec(name, namespace string, pipelineSpec numaflowv1.P
 func DeletePipelineRollout(name string) {
 
 	By("Deleting PipelineRollout")
-	foregroundDeletion := metav1.DeletePropagationForeground
-	err := pipelineRolloutClient.Delete(ctx, name, metav1.DeleteOptions{PropagationPolicy: &foregroundDeletion})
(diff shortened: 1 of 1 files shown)


tests/e2e/pipeline.go
```diff
@@ -423,8 +423,12 @@ func createPipelineRolloutSpec(name, namespace string, pipelineSpec numaflowv1.P
 func DeletePipelineRollout(name string) {
 
 	By("Deleting PipelineRollout")
-	foregroundDeletion := metav1.DeletePropagationForeground
-	err := pipelineRolloutClient.Delete(ctx, name, metav1.DeleteOptions{PropagationPolicy: &foregroundDeletion})
+	// Foreground deletion somehow changes the timing of things to cause a race condition in the PPND test, in which the same pipeline deleted then re-created causes Numaflow's "Delete Buffers/Buckets" Job
+	// to sometimes run at the same time as its "Create Buffers/Buckets" Job, which causes expected buffers/buckets to not exist
+	// Therefore, changing to Background Deletion
+	//propagationPolicy := metav1.DeletePropagationForeground
```
chandankumar4, line 429:
```suggestion
```

chandankumar4: I believe this commented line is not required but can be removed later also.

PR #741, opened 2025-05-02
fix: Handle deletions of promoted child gracefully

### LINKED ISSUE #739: Handle use case in which user deletes a promoted or upgrading pipeline during progressive rollout
# Summary

We saw a use case in which user deleted their Promoted pipeline after the Upgrading one was failed. The upgrade was still in progress, and this confused it. A new "promoted" Pipeline was created in that case. The "promotedPipelineStatus" showed the new pipeline's name but no other information which caused this failure in PipelineRollout:
```
Failed to reconcile PipelineRollout unable to restore scale values for the promoted pipeline vertices because the rollout does not have promotedChildStatus set
```

---

juliev0:
Fixes #739 

### Modifications

If a user deletes a Promoted child while in the middle of Progressive upgrade (could be prior to Assessment or after it's failed), then in addition to the Promoted child being recreated, any Upgrading children are deleted and then `upgradeInProgress` field is set back to "".

### Verification

For each of the 3 rollouts, tested prior to Assessment and after Assessment failure. Then I also, just in case, did a subsequent Progressive upgrade just to make sure that worked fine too. 

N/A

### DIFF
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -401,6 +401,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				numaLogger.WithValues("isbsvcDefinition", *newISBServiceDef).Warn("InterstepBufferService not found.")
+				return 0, nil
 			} else {
 				return 0, fmt.Errorf("error getting InterstepBufferService for status processing: %v", err)
 			}
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -252,13 +252,13 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 		} else {
 
 			// create an object as it doesn't exist
-			numaLogger.Debugf("ISBService %s/%s doesn't exist so creating", isbServiceRollout.Namespace, isbServiceRollout.Name)
-			isbServiceRollout.Status.MarkPending()
-
 			newISBServiceDef, err := r.makeTargetISBServiceDef(ctx, isbServiceRollout)
 			if err != nil {
 				return ctrl.Result{}, fmt.Errorf("error generating ISBService: %v", err)
 			}
+
+			numaLogger.Debugf("ISBService %s/%s doesn't exist so creating", isbServiceRollout.Namespace, newISBServiceDef.GetName())
+			isbServiceRollout.Status.MarkPending()
 			if newISBServiceDef != nil {
 
 				if err = r.createPromotedISBService(ctx, isbServiceRollout, newISBServiceDef); err != nil {
--- a/internal/controller/isbservicerollout/isbservicerollout_controller.go
+++ b/internal/controller/isbservicerollout/isbservicerollout_controller.go
@@ -260,14 +260,9 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
 				return ctrl.Result{}, fmt.Errorf("error generating ISBService: %v", err)
 			}
 			if newISBServiceDef != nil {
-				if err = kubernetes.CreateResource(ctx, r.client, newISBServiceDef); err != nil {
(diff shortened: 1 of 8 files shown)


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -252,22 +252,17 @@ func (r *ISBServiceRolloutReconciler) reconcile(ctx context.Context, isbServiceR
...
-				if err = kubernetes.CreateResource(ctx, r.client, newISBServiceDef); err != nil {
-					return ctrl.Result{}, fmt.Errorf("error creating ISBService: %v", err)
-				}
-				if err = r.applyPodDisruptionBudget(ctx, newISBServiceDef); err != nil {
-					return ctrl.Result{}, fmt.Errorf("failed to apply PodDisruptionBudget for ISBService %s, err: %v", newISBServiceDef.GetName(), err)
-				}
-				if err := ctlrcommon.CreateRidersForNewChild(ctx, r, isbServiceRollout, newISBServiceDef, r.client); err != nil {
-					return ctrl.Result{}, fmt.Errorf("error creating riders: %s", err)
+
+				if err = r.createPromotedISBService(ctx, isbServiceRollout, newISBServiceDef); err != nil {
```
juliev0, line 264: for each rollout, I moved the logic into a new function

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -401,6 +396,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				numaLogger.WithValues("isbsvcDefinition", *newISBServiceDef).Warn("InterstepBufferService not found.")
+				return 0, nil
```
juliev0, line 399: we should've been returning here (otherwise, it will cause panic later)

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -789,6 +785,31 @@ func (r *ISBServiceRolloutReconciler) needsUpdate(old, new *apiv1.ISBServiceRoll
...
+	// if user somehow has no Promoted ISBService and is in the middle of Progressive, this isn't right - user may have deleted the Promoted one
+	// if this happens, we need to stop the Progressive upgrade and remove any Upgrading children
+	inProgressStrategy := r.inProgressStrategyMgr.GetStrategy(ctx, isbServiceRollout)
+
+	if inProgressStrategy == apiv1.UpgradeStrategyProgressive {
+		r.inProgressStrategyMgr.UnsetStrategy(ctx, isbServiceRollout)
+		if err := progressive.Discontinue(ctx, isbServiceRollout, r, r.client); err != nil {
+			return err
+		}
+	}
```
juliev0, line 808: this last `if` condition is what I added to each rollout

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -789,6 +785,31 @@ func (r *ISBServiceRolloutReconciler) needsUpdate(old, new *apiv1.ISBServiceRoll
...
+	return nil
```
juliev0, line 809: this is the block I added for each rollout

internal/controller/isbservicerollout/isbservicerollout_controller_test.go
```diff
@@ -404,12 +404,44 @@ func Test_reconcile_isbservicerollout_Progressive(t *testing.T) {
...
+		UpgradingChildStatus: apiv1.UpgradingChildStatus{
+			Name:                   ctlrcommon.DefaultTestISBSvcRolloutName + "-1",
+			AssessmentStartTime:    &metav1.Time{Time: time.Now().Add(-1 * time.Minute)},
+			AssessmentEndTime:      &metav1.Time{Time: time.Now().Add(-30 * time.Second)},
+			AssessmentResult:       apiv1.AssessmentResultSuccess,
+			InitializationComplete: true,
+		},
+	}
+	failedUpgradingISBServiceStatus := successfulUpgradingISBServiceStatus.DeepCopy()
+	failedUpgradingISBServiceStatus.UpgradingChildStatus.AssessmentResult = apiv1.AssessmentResultFailure
```
juliev0, line 437: As part of adding tests, I decided to refactor a bit (finding some definitions that were common and creating variables out of them)

afugazzotto (approved):

PR #743, opened 2025-05-06
chore: Add e2e test for Riders

juliev0:
Fixes #TODO


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -416,10 +416,10 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 			if err := riders.UpdateRidersInK8S(ctx, newISBServiceDef, riderAdditions, riderModifications, riderDeletions, r.client); err != nil {
 				return 0, err
 			}
-		}
 
-		// update the list of riders in the Status
-		r.SetCurrentRiderList(isbServiceRollout, currentRiderList)
+			// update the list of riders in the Status
+			r.SetCurrentRiderList(ctx, isbServiceRollout, currentRiderList)
```
juliev0, line 421: Originally, I was always calling this function no matter what. But I realized this should only be called after we've successfully called `UpdateRidersInK8S()`. 

This Status is supposed to represent what's truly on the cluster. If `UpdateRidersInK8S()` fails to make the update, and then we set this anyway to reflect what we intended to do, it causes issues on a future reconciliation when we are using this Status to determine what exists.

internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -459,11 +459,16 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 		}
 		if done {
 			// update the list of riders in the Status based on our child which was just promoted
-			currentRiderList, err := r.GetDesiredRiders(isbServiceRollout, existingISBServiceDef.GetName(), newISBServiceDef)
+			promotedISBService, err := ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, isbServiceRollout, common.LabelValueUpgradePromoted, nil, true, r.client)
+			if err != nil {
+				return 0, err
+			}
+
+			currentRiderList, err := r.GetDesiredRiders(isbServiceRollout, promotedISBService.GetName(), promotedISBService)
```
juliev0, line 467: fixed the call to `GetDesiredRiders()` to get the riders from the actual promoted child, which is the right list to set in our `Status` below

tests/manifests/default/kustomization.yaml
```diff
@@ -3,11 +3,12 @@ kind: Kustomization
 
 resources:
   - ../../../config/default
-  - ./controller_def_1.4.1.yaml
   - ./controller_def_1.4.2.yaml
   - ./controller_def_1.4.3.yaml
   - ./controller_def_1.4.4.yaml
+  - ./controller_def_1.5.0-rc1.yaml
   - ./prometheus-monitors.yaml
+  - https://raw.githubusercontent.com/kubernetes/autoscaler/vpa-release-1.0/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml
```
juliev0, line 11: CRD for VPA

tests/manifests/default/kustomization.yaml
```diff
@@ -3,11 +3,12 @@ kind: Kustomization
 
 resources:
   - ../../../config/default
-  - ./controller_def_1.4.1.yaml
   - ./controller_def_1.4.2.yaml
   - ./controller_def_1.4.3.yaml
   - ./controller_def_1.4.4.yaml
+  - ./controller_def_1.5.0-rc1.yaml
```
juliev0, line 9: not related to this PR but I incorporated the latest numaflow controller and removed an old one

afugazzotto (approved):

PR #746, opened 2025-05-06
feat: add some important progressive upgrade related fields to the logs

### LINKED ISSUE #740: Add to all progressive logs the rollout object and the promoted and upgrading objects
# Summary

Add to all progressive related logs (maybe in context) the rollout object and the promoted and upgrading objects. Do this using the `WithValues` func. To avoid increasing the messages length during local debugging, create a feature flag to add these values in the logs.

# Use Cases

These extra log values could be helpful in debugging preprod and prod issues via Splunk logs.
NOTE: in Splunk, the fields can be collapsed avoiding the issue of having to read messages that are too long.

---

afugazzotto:
Fixes #740 

### Modifications

Added the progressive upgrade `rolloutObject`, `promotedChild`, and `upgradingChild` fields to the logs to help with debugging via Splunk.
Created the `progressive.logObjects` config var to only enable these detailed logs only on clusters.

### Verification

Manually verified the logs locally.

None.

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -113,6 +113,9 @@ type ProgressiveConfig struct {
 
 	// timeout duration which AnalysisRuns cannot continue to run after
 	AnalysisRunTimeout string `json:"analysisRunTimeout" mapstructure:"analysisRunTimeout"`
+
+	// when set to true, adds the rollout, promoted, and upgrading objects to all the log messages during progressive upgrade
+	LogObjects bool `json:"logObjects" mapstructure:"logObjects"`
 }
 
 // DefaultAssessmentSchedule defines a default schedule for each Kind
@@ -381,5 +384,4 @@ func (config *ProgressiveConfig) GetAnalysisRunTimeout() (time.Duration, error) 
 	}
 
 	return time.Duration(analysisRunTimeout) * time.Second, nil
-
 }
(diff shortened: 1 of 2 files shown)


internal/controller/config/config.go
```diff
@@ -113,6 +113,9 @@ type ProgressiveConfig struct {
 
 	// timeout duration which AnalysisRuns cannot continue to run after
 	AnalysisRunTimeout string `json:"analysisRunTimeout" mapstructure:"analysisRunTimeout"`
+
+	// when set to true, adds the rollout, promoted, and upgrading objects to all the log messages during progressive upgrade
+	LogObjects bool `json:"logObjects" mapstructure:"logObjects"`
```
afugazzotto, line 118: @juliev0 should we enable this in the NMG?

afugazzotto: Also, don't really like the name, what do you think about `enableDetailedLogs` or `enableExtraLogFields`? Any other ideas?

juliev0 (approved):

internal/controller/config/config.go
juliev0: > @juliev0 should we enable this in the NMG?

sounds good!

juliev0: > Also, don't really like the name, what do you think about `enableDetailedLogs` or `enableExtraLogFields`? Any other ideas?

I'm okay with the name, thanks!

PR #768, opened 2025-05-16
chore: refactor to simplify some progressive logic

juliev0:
Fixes #TODO

### DIFF
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -142,18 +142,13 @@ func ProcessResource(
 	}
 
 	// is there currently an "upgrading" child?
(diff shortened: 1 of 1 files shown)


internal/controller/progressive/progressive.go
```diff
@@ -142,68 +142,56 @@ func ProcessResource(
...
 			} else {
 				return false, 0, err
 			}
-		}
-	}
 
-	// nothing to do (either there's nothing to upgrade, or we just created an "upgrading" child, and it's too early to start reconciling it)
-	if currentUpgradingChildDef == nil {
-		return true, 0, err
-	}
```
juliev0, line 170: this was the only place where we weren't previously calling the Live K8S API to get the upgrading child. 

This is not called very often actually, so we weren't saving much on Live K8S API calls anyway, and the code was more complicated for no good reason.

PR #769, opened 2025-05-16
chore: remove redundant calls

juliev0:
### Modifications

While I started working on the issue to do "premature cancellation of in progress rollout" for progressive, I realized there was a small refactor that could be done.

At some point, I was trying to save on Kubernetes API live calls by only getting the live version in certain circumstances if I really needed it. 

But now with the current code, it seems like the savings is very limited and is just making the code harder to read.

### DIFF
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -142,18 +142,14 @@ func ProcessResource(
 	}
 
 	// is there currently an "upgrading" child?
-	currentUpgradingChildDef, err := ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradeInProgress, nil, false, c)
+	currentUpgradingChildDef, err := ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradeInProgress, nil, true, c)
 	if err != nil {
 		return false, 0, err
 	}
 
 	// if there's a difference between the desired spec and the current "promoted" child, and there isn't already an "upgrading" definition, then create one and return
 	if promotedDifference && currentUpgradingChildDef == nil {
-		// Create it, first making sure one doesn't already exist by checking the live K8S API
-		currentUpgradingChildDef, err = ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradeInProgress, nil, true, c)
-		if err != nil {
-			return false, 0, fmt.Errorf("error getting %s: %v", currentUpgradingChildDef.GetKind(), err)
-		}
+		// Create it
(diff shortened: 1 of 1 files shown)


internal/controller/progressive/progressive.go
```diff
@@ -171,12 +167,6 @@ func ProcessResource(
 
 	// There's already an Upgrading child, now process it
 
-	// Get the live resource so we don't have issues with an outdated cache
-	currentUpgradingChildDef, err = kubernetes.GetLiveResource(ctx, currentUpgradingChildDef, rolloutObject.GetChildGVR().Resource)
-	if err != nil {
-		return false, 0, err
-	}
-
```
juliev0, line 179: up on line 169 I think is the only place where we were not calling the live K8S API, and really we don't hit that line very often

internal/controller/progressive/progressive.go
```diff
@@ -142,18 +142,14 @@ func ProcessResource(
 	}
 
 	// is there currently an "upgrading" child?
-	currentUpgradingChildDef, err := ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradeInProgress, nil, false, c)
```
juliev0, line 145: this gets the value from the cache if you pass it false, but from the live API if you pass it true

internal/controller/progressive/progressive.go
```diff
@@ -142,18 +142,14 @@ func ProcessResource(
...
 		return false, 0, err
 	}
 
 	// if there's a difference between the desired spec and the current "promoted" child, and there isn't already an "upgrading" definition, then create one and return
 	if promotedDifference && currentUpgradingChildDef == nil {
-		// Create it, first making sure one doesn't already exist by checking the live K8S API
-		currentUpgradingChildDef, err = ctlrcommon.FindMostCurrentChildOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradeInProgress, nil, true, c)
-		if err != nil {
-			return false, 0, fmt.Errorf("error getting %s: %v", currentUpgradingChildDef.GetKind(), err)
-		}
```
juliev0, line 156: we were getting the live one here

internal/controller/progressive/progressive.go
```diff
@@ -171,12 +167,6 @@ func ProcessResource(
 
 	// There's already an Upgrading child, now process it
 
-	// Get the live resource so we don't have issues with an outdated cache
-	currentUpgradingChildDef, err = kubernetes.GetLiveResource(ctx, currentUpgradingChildDef, rolloutObject.GetChildGVR().Resource)
-	if err != nil {
-		return false, 0, err
-	}
```
juliev0, line 178: we were getting the live one here

afugazzotto (approved):

PR #771, opened 2025-05-17
feat: Preempt progressive upgrade in progress if there's a new spec

### LINKED ISSUE #535: Enable Premature Cancellation feature for progressive
# Summary

If user rolls back to the prior version of their Rollout (especially PipelineRollout and MonoVertexRollout), it would be best if they don't have to wait for 5 minutes to start updating back to that version. (This is an ask from the Multiworkload team who will be doing multi-workload automatic rollback.)

Take the scenario of PipelineRollout in a progressive state:
- pipeline-0: "promoted" and pipeline-1: "upgrading"
- PipelineRollout gets put back to the spec identical to "pipeline-0" before assessment is complete
- If there is any new update, we can discontinue the in-progress update and start a new one.  I think we [can move the logic](https://github.com/numaproj/numaplane/blob/v0.22.0/internal/controller/progressive/progressive.go#L409-L446) that we currently do when there is a new PipelineRollout spec after the assessment has been deemed "Failed" such that instead it occurs somewhere near the beginning of the progressive-processing logic. In this case, the logic would occur whether a Failure had occurred or even if the assessment is still in progress (or even if we haven't gotten to "assessment start time" yet). 

Currently, we are setting [common.LabelValueProgressiveFailure](https://github.com/numaproj/numaplane/blob/v0.22.0/internal/controller/progressive/progressive.go#L428) as the "reason for recycling". Perhaps we could have these 2 instead?: `common.LabelValueProgressiveFailureReplacement` and `common.LabelValueProgressiveReplacement`. Part of me thinks that at the end of the day it's just `common.LabelValueProgressiveReplacement` regardless of whether it failed or not, but the behavior of [PipelineRollout.Recycle()](https://github.com/numaproj/numaplane/blob/v0.22.0/internal/controller/pipelinerollout/pipelinerollout_controller.go#L1160) does depend on whether it's failed or not, so we need to handle that one way or another...

---

juliev0:
Fixes #535 

#### Main modification:
No longer wait until Assessment is complete and failed to replace the upgrading child with a new one. 

Now while we're in Progressive, we can just check if there's a difference between the current "upgrading" one and the current Rollout spec to determine if we want to preempt the current upgrade with a new one. 

#### Secondary modification:
When deleting previous Pipeline, we have been pausing and requiring the Pipeline to be fully drained in order for it to be deleted. This means that if it doesn't drain successfully, the old Pipeline will remain for user to need to delete on their own. Until we make this configurable, I think we should default to not requiring drain for the deletion (this is described inline).

### Verification

Created PipelineRollout and ISBServiceRollout. Updated both at the same time to a new spec causing progressive rollout. Before progressive rollout had been assessed, updated the specs back to their originals. Both caused the previous pipeline and isbsvc to be replaced with new ones and succeed.

N/A

### DIFF
--- a/internal/controller/monovertexrollout/monovertexrollout_controller.go
+++ b/internal/controller/monovertexrollout/monovertexrollout_controller.go
@@ -248,9 +248,6 @@ func (r *MonoVertexRolloutReconciler) reconcile(ctx context.Context, monoVertexR
 			if err != nil {
 				return ctrl.Result{}, fmt.Errorf("error processing existing MonoVertex: %v", err)
 			}
-			if requeueDelay > 0 {
-				return ctrl.Result{RequeueAfter: requeueDelay}, nil
-			}
 			// process status
 			r.processMonoVertexStatus(ctx, existingMonoVertexDef, monoVertexRollout)
 		}
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -179,7 +179,8 @@ func ProcessResource(
 	}
 
 	// if the Upgrading child status exists but indicates that we aren't done with upgrade process, then do postupgrade process
-	if !childStatus.InitializationComplete && childStatus.AssessmentResult == apiv1.AssessmentResultUnknown {
+	initializationIncomplete := !childStatus.InitializationComplete && childStatus.AssessmentResult == apiv1.AssessmentResultUnknown
+	if initializationIncomplete {
 		needsRequeue, err := startPostUpgradeProcess(ctx, rolloutObject, existingPromotedChild, currentUpgradingChildDef, controller, c)
 		if needsRequeue {
 			return false, common.DefaultRequeueDelay, err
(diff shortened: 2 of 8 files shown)


internal/controller/monovertexrollout/monovertexrollout_controller.go
```diff
@@ -248,9 +248,6 @@ func (r *MonoVertexRolloutReconciler) reconcile(ctx context.Context, monoVertexR
 			if err != nil {
 				return ctrl.Result{}, fmt.Errorf("error processing existing MonoVertex: %v", err)
 			}
-			if requeueDelay > 0 {
-				return ctrl.Result{RequeueAfter: requeueDelay}, nil
-			}
```
juliev0, line 253: there was no need for this to be here. We are checking below on line 276 for `requeueDelay > 0` and requeueing there. Having it here prevented processing monovertex status below and garbage collecting children.

internal/controller/progressive/progressive.go
```diff
@@ -188,6 +189,12 @@ func ProcessResource(
 		}
 	}
 
+	// determine if we need to replace the Upgrading child with a newer one
+	needsRequeue, err := checkForUpgradeReplacement(ctx, rolloutObject, controller, existingPromotedChild, currentUpgradingChildDef, c)
+	if needsRequeue {
+		return false, common.DefaultRequeueDelay, err
+	}
```
juliev0, line 196: added new function `checkForUpgradeReplacement()` to consolidate the logic into a function

juliev0: and now calling it here instead of after we assess failure

afugazzotto (approved):

PR #776, opened 2025-05-20
fix: a better way to ensure pipeline is drainable for recycling

juliev0:
### Modifications

I found an issue with this call to [ensurePipelineIsDrainable()](https://github.com/numaproj/numaplane/blob/v0.22.1/internal/controller/pipelinerollout/pipelinerollout_controller.go#L1191) similar to the bug we found recently. This call is made when recycling a pipeline, and the purpose of that call is to make sure that if we scaled the Promoted Pipeline Vertex to 0 pods, we don't leave it that way - otherwise, the pipeline can't be drained. So, basically the logic was looking to see how many Pods were running, and bump it up to 1 if it's at 0. However, the issue is: what if the user actually wants max=0? Then we should probably not scale it above 0 in that case.

So, really we need to use the user's Pipeline spec (from the way it was originally defined for that iteration of the pipeline) to make this determination. 

Since we don't really save this information anywhere for "recyclable" Pipelines, the easiest thing to do was to make it so that directly prior to recycling it, we scale it back up to its original values (as defined in `upgradingChildStatus`). Then it will be >0 if user wants that, but otherwise it will be 0 if they set `max=0`.

### Verification

Tested both the case of the recycling of the promoted pipeline and the recycling of the previous upgrading pipeline to ensure it scaled correctly.


internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -1185,17 +1185,6 @@ func (r *PipelineRolloutReconciler) Recycle(ctx context.Context,
...
-	if requireDrain {
-		modified, err := r.ensurePipelineIsDrainable(ctx, pipeline)
-		if err != nil {
-			return false, err
-		}
-		if modified {
-			// since we just changed the pipeline, give numaflow a chance to reconcile it and then pause it, so requeue
-			return false, nil
-		}
-	}
```
juliev0, line 1197: This code was making sure that if there were only 0 pods of a vertex, we bump it to 1 so it can drain.

Note: this is actually logic we should have been executing in the case of `if pause` and not just in the case of `if requireDrain` (we would want to be able to drain it if we're trying to pause at all, regardless of whether we needed to drain it)

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -432,14 +443,25 @@ func (r *MonoVertexRolloutReconciler) ProcessPromotedChildPostFailure(
 		return true, errors.New("unable to perform post-upgrade operations because the rollout does not have promotedChildStatus set")
 	}
 
-	requeue, err := scalePromotedMonoVertexToOriginalValues(ctx, monoVertexRollout.Status.ProgressiveStatus.PromotedMonoVertexStatus, promotedMonoVertexDef, c)
+	err := scalePromotedMonoVertexToOriginalValues(ctx, monoVertexRollout.Status.ProgressiveStatus.PromotedMonoVertexStatus, promotedMonoVertexDef, c)
```
juliev0, line 446: removed the boolean from the function because we were no longer using it

tests/manifests/default/controller_def_1.5.0-rc4.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: numaflow-controller-definitions-1.5.0-rc1
```
juliev0, line 4: I removed all the older NC definitions and just left the latest one there

tests/manifests/default/kustomization.yaml
```diff
@@ -6,7 +6,6 @@ resources:
   - ./controller_def_1.4.2.yaml
   - ./controller_def_1.4.3.yaml
   - ./controller_def_1.4.4.yaml
-  - ./controller_def_1.5.0-rc1.yaml
```
juliev0, line 9: for now, with new RC versions coming, maybe just don't incorporate into kustomization.yaml until there's one we're ready to test with?

afugazzotto (approved):

PR #777, opened 2025-05-20
E2e coverage test

chandankumar4:
Fixes #TODO


.github/workflows/ci.yaml
```diff
@@ -168,13 +177,44 @@ jobs:
...
+          kubectl wait --for=condition=Ready --timeout=5m pod/$POD_NAME -n numaplane-system
 
       - name: Run e2e tests
         env:
           GOPATH: /home/runner/go
-        run: KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true STRATEGY=${{matrix.strategy}} ENABLE_POD_LOGS=true make test-${{matrix.case}}-e2e
+        run: |
+          KUBECONFIG=~/.kube/config VERSION=${{ github.sha }} DOCKER_PUSH=true STRATEGY=${{matrix.strategy}} ENABLE_POD_LOGS=true make test-${{matrix.case}}-e2e
+
+      ## This is a temporary solution to kill the process twice, until we have fix for a stuck process after initial kill.
```
juliev0, line 195: that's interesting that using SIGTERM twice works. I guess SIGTERM allows the program to try to shut down gracefully, right?

If it doesn't work the first time, I don't know why it would work the second time. ?

I would think it would need to be SIGTERM followed by SIGKILL (Not that I want to hold up this PR from this but it's weird)

tests/manifests/special-cases/no-strategy/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-  - ../../default
+  - ../../e2e
```
juliev0, line 5: If these manifest are based on `e2e` then how about we nest the `special-cases` directory under the `e2e` directory?

juliev0: and maybe we add a `default` directory under `e2e` which is where that manifest should go.

Come to think of it, but at this point we should probably make the `default` case "progressive" (and "pause and drain" can become a special case).

juliev0 (approved):

PR #781, opened 2025-05-21
chore: simplify calls which return *UnstructuredList to return UnstructuredList

juliev0:
Fixes #TODO

### DIFF
--- a/internal/controller/common/numaflowtypes/pipeline.go
+++ b/internal/controller/common/numaflowtypes/pipeline.go
@@ -91,7 +91,7 @@ func GetRolloutForPipeline(ctx context.Context, c client.Client, pipeline *unstr
 	return pipelineRollout, nil
 }
 
-func GetPipelinesForRollout(ctx context.Context, c client.Client, pipelineRollout *apiv1.PipelineRollout, live bool) (*unstructured.UnstructuredList, error) {
+func GetPipelinesForRollout(ctx context.Context, c client.Client, pipelineRollout *apiv1.PipelineRollout, live bool) (unstructured.UnstructuredList, error) {
 	if live {
 		return kubernetes.ListLiveResource(ctx, common.PipelineGVR.Group, common.PipelineGVR.Version, "pipelines", pipelineRollout.GetNamespace(), fmt.Sprintf("%s=%s", common.LabelKeyParentRollout, pipelineRollout.GetName()), "")
 	} else {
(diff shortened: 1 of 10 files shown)


internal/util/kubernetes/unstructured_util.go
```diff
@@ -72,13 +72,17 @@ func ListLiveResource(
...
 		Group:    apiGroup,
 		Version:  version,
 		Resource: pluralName,
 	}
-	return DynamicClient.Resource(gvr).Namespace(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector, FieldSelector: fieldSelector})
+	ul, err := DynamicClient.Resource(gvr).Namespace(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector, FieldSelector: fieldSelector})
+	if err != nil || ul == nil {
+		return unstructured.UnstructuredList{}, err
+	}
+	return *ul, nil
```
juliev0, line 85: Now if we get a nil back, we return `unstructured.UnstructuredList{}` instead

internal/controller/common/numaflowtypes/pipeline.go
```diff
@@ -91,7 +91,7 @@ func GetRolloutForPipeline(ctx context.Context, c client.Client, pipeline *unstr
 	return pipelineRollout, nil
 }
 
-func GetPipelinesForRollout(ctx context.Context, c client.Client, pipelineRollout *apiv1.PipelineRollout, live bool) (*unstructured.UnstructuredList, error) {
+func GetPipelinesForRollout(ctx context.Context, c client.Client, pipelineRollout *apiv1.PipelineRollout, live bool) (unstructured.UnstructuredList, error) {
```
chandankumar4, line 94: I agree that we can avoid `nil` check after returning struct instead of pointer but also It can affect performance in case of larger dataset as it has to copy the complete data to return.

chandankumar4 (approved):

internal/controller/common/numaflowtypes/pipeline.go
juliev0: Good question. This is actually the UnstructuredList struct:
```
type UnstructuredList struct {
	Object map[string]interface{}

	// Items is a list of unstructured objects.
	Items []Unstructured `json:"items"`
}
```

It's not doing a deep copy on returning, so really setting both the map and setting the slice just copies a reference to data rather than copying all of the data.

Issue #783, opened 2025-05-22
NumaflowController is not pruning manifests


juliev0:
**Describe the bug**
`numaflow-server` `Deployment` was removed between 1.4.2 and 1.4.3 - however, appears that it is not being pruned. This is probably a gitops-engine setting that we're missing.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

juliev0:
Unfortunately, I thought this was going to be trivial and it wasn't. I tried setting [WithPruneLast()](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L449) to true and also tried WithPrune(). The test I did was this: change our 1.4.3 manifest to remove `numaflow-server` `Deployment`. Apply a `NumaflowControllerRollout` with v1.4.2. Then edit it to use v1.4.3. This did not successfully remove the `numaflow-server` `Deployment`, however. 

As an aside: you may know that a [few months ago](https://github.com/numaproj/numaplane/commit/1a5fad32a6b7c93ef7b0f1f43dccc29a442d477b) we set this [Replace](https://github.com/numaproj/numaplane/blob/main/internal/controller/numaflowcontroller/numaflowcontroller_controller.go#L451) field true for the purpose of replacing the old manifest with the new manifest (the problem being solved was that sometimes in-place updates of manifests can fail).

juliev0:
To reproduce this, you need to remove the Numaflow Server from the 1.4.3 manifest. The issue is specifically related to resources which don’t exist in the new version. (They need to be pruned.)

PR #786, opened 2025-05-29
fix: Prune NumaflowController resource which is removed in latest update

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/783

### Modifications
- Handle case for pruning resource when it's deleted in new version
  - Add `LastAppliedVersion` in CRD status field
  - `LastAppliedVersion` is set each time after successful reconciliation.
- It won't fix existing issue as `LastAppliedVersion` is unknown.

### Verification
Verified in local k8s cluster with below scenatios
- Deployed `NumaflowController` v1.4.2` which has the `numaflow-server`, then deployed `NumaflowController` v1.4.3 which doesn't have `numaflow-server`.

Yes

### DIFF
--- a/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
+++ b/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
@@ -241,20 +241,37 @@ func (r *NumaflowControllerReconciler) reconcile(
 		return ctrl.Result{}, err
 	}
 
+	// Generate the manifest for the NumaflowController based on the version specified in the controller spec. Which is used for
+	// creating the Deployment and other resources.
 	newVersion := controller.Spec.Version
 	newVersionTargetObjs, err := determineTargetObjects(controller, newVersion, namespace)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("unable to determine the target objects for the new version %s: %w", newVersion, err)
 	}
-
 	numaLogger.Debugf("found %d target objects associated with NumaflowController version %s", len(newVersionTargetObjs), newVersion)
 
-	// apply controller - this handles syncing in the cases in which our Controller  isn't updating
+	// Generate the manifest for the old version of the NumaflowController, which is used for calculating the diff between the old and new versions.
+	oldVersion := controller.Status.LastAppliedVersion
+
+	// TODO: This is temporary logic to handle the case where the controller is being created for the first time.
+	// Will remove this code once we have LastAppliedVersion set in the controller.
+	if oldVersion == "" {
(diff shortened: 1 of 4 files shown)


internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -494,7 +512,7 @@ func (r *NumaflowControllerReconciler) compareState(
 		return gitopsSync.ReconciliationResult{}, nil, err
 	}
 	infoProvider = clusterCache
-	liveObjByKey, err := r.stateCache.GetManagedLiveObjs(controller.Name, namespace, targetObjs)
+	liveObjByKey, err := r.stateCache.GetManagedLiveObjs(controller.Name, namespace, oldObjs)
```
juliev0, line 515: nice find here that we were only comparing the currently defined objects

internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -241,20 +241,37 @@ func (r *NumaflowControllerReconciler) reconcile(
...
+	// TODO: This is temporary logic to handle the case where the controller is being created for the first time.
+	// Will remove this code once we have LastAppliedVersion set in the controller.
+	if oldVersion == "" {
+		numaLogger.Debug("No previous version found, assuming this is the first reconciliation")
+		oldVersion = newVersion // if no previous version, assume it's the same as the new version
+	}
+
+	oldVersionTargetObjs, err := determineTargetObjects(controller, oldVersion, namespace)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to determine the target objects for the old version %s: %w", oldVersion, err)
```
juliev0, line 265: What if for some reason the ConfigMap no longer exists? In that case, how about we log a Warning but just move forward anyway?

juliev0 (commented):
I think this approach is good at least for now.

One thing I'm thinking about is this (unusual but possible) case:
1. 1.4.2 is applied for this Controller - this includes "numaflow-server"
2. at some point somebody changes the definition for 1.4.2 to remove "numaflow-server"
3. This Controller gets updated to 1.4.3
4. Since we are looking at the 1.4.2 ConfigMap, it doesn't include "numaflow-server" anymore, so "numaflow-server" doesn't get removed

The ideal approach may be to list in the Status all existing children owned by the `NumaflowController` and use those. This would also take care of the problem that the ConfigMap for 1.4.2 no longer exists. (Unless you disagree, maybe I'll add a follow up issue for that?)

Issue #787, opened 2025-05-29
Handle pruning of previous Numaflow Controller resources in the case of deleted or modified ConfigMap


juliev0:
# Summary

See comment: https://github.com/numaproj/numaplane/pull/786#pullrequestreview-2878513738

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #789, opened 2025-05-30
fix: Prune NumaflowController resource which are removed in latest version

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/783 
Fixes https://github.com/numaproj/numaplane/issues/787

### Modifications
- Handle case for pruning resource when it's deleted in new version
  - List all the resources in cluster whose owner reference API Version is `numaplane`
  - Then compute the diff with upcoming resource and remove if any.

### Verification
Verified in local k8s cluster with below scenatios

Case: 1
- Deployed `NumaflowController` v1.4.2 which has the `numaflow-server`, then deployed `NumaflowController` v1.4.3 which doesn't have `numaflow-server`. As expected `numaflow-server` is removed.
Case: 2
- Cluster is running `numaflow-server` which is not in current manifest due to bug in `Numaplane-controller`, Then updated the `numaplane-controller` with the fix image, as expected `numaflow-server` is removed.

### DIFF
--- a/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
+++ b/internal/controller/numaflowcontroller/numaflowcontroller_controller.go
@@ -37,6 +37,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	yamlserializer "k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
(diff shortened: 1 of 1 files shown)


internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -246,15 +247,19 @@ func (r *NumaflowControllerReconciler) reconcile(
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("unable to determine the target objects for the new version %s: %w", newVersion, err)
 	}
-
 	numaLogger.Debugf("found %d target objects associated with NumaflowController version %s", len(newVersionTargetObjs), newVersion)
 
+	existingClusterResources, err := r.determineExistingManagedResourceInCluster(ctx, namespace)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to determine existing managed resources in cluster: %w", err)
+	}
```
juliev0, line 255: In one of these functions, it would be great if we can add debug level log(s) for both target objects and existing objects. I'm imagining after we apply to the tech-pi cluster, we'll check that log and make sure it looks good. And if anything were to ever go wrong with this (which I don't think it should), we'll have that log line to look back at.

chandankumar4: okay...but it not straight forward to log the resource as it's "array to point object" so it will print address if log directly, need to loop through all the resource then log it as value? 
BTW I have added the length log for now!!

juliev0: > need to loop through all the resource then log it as value?

I'm thinking yeah, let's do that

chandankumar4: Added

juliev0 (approved):
PR looks good - feel free to merge

PR #791, opened 2025-06-01
fix: force promote is broken

juliev0:
Fixes #TODO

### DIFF
--- a/tests/e2e/progressive-monovertex-e2e/monovertex_test.go
+++ b/tests/e2e/progressive-monovertex-e2e/monovertex_test.go
@@ -100,42 +100,9 @@ var _ = Describe("Progressive MonoVertex E2E", Serial, func() {
 		VerifyMonoVertexRolloutInProgressStrategy(monoVertexRolloutName, apiv1.UpgradeStrategyNoOp)
 		VerifyMonoVertexRolloutHealthy(monoVertexRolloutName)
 
-		By("Updating the MonoVertex Topology to cause a Progressive change - Failure case")
-		updatedMonoVertexSpec := initialMonoVertexSpec.DeepCopy()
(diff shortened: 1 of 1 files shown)


internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -201,6 +202,29 @@ func (r *PipelineRolloutReconciler) needPPND(ctx context.Context, pipelineRollou
...
+// Determine if the Pipeline has changed and needs updating
+// We need to ignore any field that could be set by Numaplane in the pause-and-drain process
+// TODO: do we really need this or can we just use usde functionality and just look for data loss fields?
+func (r *PipelineRolloutReconciler) pipelineNeedsUpdatingForPPND(ctx context.Context, from, to *unstructured.Unstructured) (bool, error) {
+	numaLogger := logger.FromContext(ctx)
+	fromCopy := from.DeepCopy()
+	toCopy := to.DeepCopy()
+	// remove lifecycle.desiredPhase field from comparison to test for equality
+	numaflowtypes.PipelineWithoutDesiredPhase(fromCopy)
+	numaflowtypes.PipelineWithoutDesiredPhase(toCopy)
```
juliev0, line 214: For this function, we ignore `lifecycle.desiredPhase`. For the other function we ignore fields changed when doing Progressive upgrade (e.g. `scale`)

internal/util/maps.go
```diff
@@ -35,3 +40,35 @@ func CompareMaps(existing, new map[string]string) bool {
 	}
 	return CompareStructNumTypeAgnostic(existing, new)
```
juliev0, line 41: @chandankumar4 do we need to call `CompareStructNumTypeAgnostic()` since we're passing it 2 `map[string]string`?

Issue #678, opened 2025-03-27
Rolling Window Assessment


juliev0:
# Summary

Instead of always starting our check at a particular time (e.g. 5 minutes) and then checking for 1 minute and requiring every measurement be successful, we could do the following:

Start the check earlier, such as at 2 minutes. Allow it to run up until say 7 minutes. If there's every a 1 minute period in there in which every check is Successful, we return Success at that point rather than waiting.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
My understanding is, this is the feature that AnalysisRun supports, but we need to build the mechanism in Numaplane to translate the AnalysisTemplate to the corresponding AnalysisRun, correct?

> Note that we will want our AnalysisTemplate(s) to do the same thing. Recall an AnalysisTemplate can have multiple [Metrics](https://github.com/argoproj/argo-rollouts/blob/v1.8.2/pkg/apis/rollouts/v1alpha1/analysis_types.go#L90) configured. We may be able to accomplish this with just 1 Metric or if needed, multiple.
> 
> I see one field in there called [ConsecutiveSuccessLimit](https://github.com/argoproj/argo-rollouts/blob/v1.8.2/pkg/apis/rollouts/v1alpha1/analysis_types.go#L127): "the number of consecutive times the measurement must succeed for the metric to be successful" - this is an example of a field that could enable the notion of "early success". If we get n number of measurements that have been successful, we stop. And then if we reach [Count](https://github.com/argoproj/argo-rollouts/blob/v1.8.2/pkg/apis/rollouts/v1alpha1/analysis_types.go#L101), and we didn't achieve the `ConsecutiveSuccessLimit`, we will be failed. cc [@whynowy](https://github.com/whynowy)

juliev0:
> My understanding is, this is the feature that AnalysisRun supports, but we need to build the mechanism in Numaplane to translate the AnalysisTemplate to the corresponding AnalysisRun, correct?
> 
> > Note that we will want our AnalysisTemplate(s) to do the same thing. Recall an AnalysisTemplate can have multiple [Metrics](https://github.com/argoproj/argo-rollouts/blob/v1.8.2/pkg/apis/rollouts/v1alpha1/analysis_types.go#L90) configured. We may be able to accomplish this with just 1 Metric or if needed, multiple.
[...]

PR #793, opened 2025-06-06
feat: change AnalysisRun launch time

### LINKED ISSUE #770: Launch AnalysisRun once resource health check succeeds
# Summary

Only once the resource health check succeeds should we start the AnalysisRun. This will save resources and reduce load on Argo Rollouts.

It will also fit better with the rolling window algorithm. 

---

juliev0:
Fixes #770 

### Modifications

Previously, the AnalysisRun was being launched at the beginning of the resource health check window. This PR makes it so that it only gets launched after the resource health check window has passed (i.e. if for 1 minute (configurable), it succeeded every time). 

The reason we are moving it to there is because of [this](https://github.com/numaproj/numaplane/issues/678) plan to turn our fixed window into a rolling window. That will mean that we will declare a successful resource health check after one consecutive minute of success, regardless of when that minute is. That rolling window means we need to rethink the question: when do we create the AnalysisRun? Derek recommended in this case starting it after that one minute of success is completed. 

So, this PR makes it so we start the AnalysisRun after 1 minute of success is complete.

### Verification

tested PipelineRollout, MonoVertexRollout and ISBServiceRollout to confirm they're working with the changes. Observed their Status change as progressive upgrades were done. For PipelineRollout and MonoVertexRollout confirmed that AnalysisRun got created once basic assessment window was completed.


internal/controller/isbservicerollout/isbservicerollout_controller.go
```diff
@@ -476,7 +476,7 @@ func (r *ISBServiceRolloutReconciler) processExistingISBService(ctx context.Cont
 				return 0, fmt.Errorf("error getting PipelineRollouts; can't enqueue pipelines: %s", err.Error())
 			}
 			for _, pipelineRollout := range pipelineRollouts {
-				numaLogger.WithValues("pipeline rollout", pipelineRollout.Name).Debugf("Created new upgrading isbsvc; now enqueueing pipeline rollout")
+				numaLogger.WithValues("pipeline rollout", pipelineRollout.Name).Debugf("Not done processing upgrading isbsvc; now enqueueing pipeline rollout")
```
juliev0, line 479: todo: check if we really need to be enqueuing here in the case that the new isbsvc did not just get created

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -10,6 +10,8 @@ spec:
     analysis:
       args:
       templates:
+      - templateName: mvtx-template
+        clusterScope: false
```
juliev0, line 14: todo: remove

config/crd/bases/numaplane.numaproj.io_isbservicerollouts.yaml
```diff
@@ -238,17 +238,17 @@ spec:
...
                           or succeeded, or to be determined
                         type: string
-                      assessmentStartTime:
-                        description: AssessmentStartTime indicates the time at/after
+                      basicAssessmentEndTime:
+                        description: BasicAssessmentEndTime indicates the time after
+                          which no more assessments will be performed
+                        format: date-time
+                        type: string
+                      basicAssessmentStartTime:
```
juliev0, line 250: I renamed these to clarify it's just the "basic resource health check" we're assessing for

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -45,55 +42,56 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
+		mvtxRollout.SetUpgradingChildStatus(childStatus)
+		return assessment, reasonFailure, nil
+	}
+	if assessment == apiv1.AssessmentResultSuccess {
+		// has AssessmentEndTime been set? if not, set it - now we can start our interval
+		if !childStatus.IsAssessmentEndTimeSet() {
+			assessmentEndTime := metav1.NewTime(time.Now().Add(assessmentSchedule.Period))
+			childStatus.BasicAssessmentEndTime = &assessmentEndTime
+			numaLogger.WithValues("childStatus", *childStatus).Debug("set upgrading child AssessmentEndTime")
+			mvtxRollout.SetUpgradingChildStatus(childStatus)
```
juliev0, line 73: This logic for determining the Assessment End Time used to be in `progressive.go`, but I needed to move it to here and to `pipelinerollout_progressive.go` because these `AssessUpgradingChild()` functions are now being called all the way from the beginning of the resource health check window to when the AnalysisRun is complete (if there is an AnalysisRun).

internal/controller/monovertexrollout/monovertexrollout_progressive.go
```diff
@@ -45,55 +42,65 @@ func (r *MonoVertexRolloutReconciler) CreateUpgradingChildDefinition(ctx context
...
+
+		// if end time has arrived (i.e. we continually determined "Success" for the entire prescribed period of time),
+		// if we need to launch an AnalysisRun, we can do it now;
+		// otherwise, we can declare success
+		if childStatus.BasicAssessmentEndTimeArrived() {
+			analysis := mvtxRollout.GetAnalysis()
+			// only check for and create AnalysisRun if templates are specified
+			if len(analysis.Templates) > 0 {
+				// this will create an AnalysisRun if it doesn't exist yet; or otherwise it will check if it's finished running
+				analysisStatus, err := progressive.PerformAnalysis(ctx, existingUpgradingChildDef, mvtxRollout, mvtxRollout.GetAnalysis(), mvtxRollout.GetAnalysisStatus(), r.client)
```
juliev0, line 91: moved to shared function

internal/controller/progressive/analysis.go
```diff
@@ -134,5 +142,95 @@ func CreateAnalysisRun(ctx context.Context, analysis apiv1.Analysis, existingUpg
...
+	analysisStatus *apiv1.AnalysisStatus,
+	c client.Client,
+) (*apiv1.AnalysisStatus, error) {
+	if analysisStatus == nil {
+		return analysisStatus, errors.New("analysisStatus not set")
+	}
+
+	analysisRun := &argorolloutsv1.AnalysisRun{}
+
+	analysisRunName := fmt.Sprintf("%s-%s", strings.ToLower(existingUpgradingChildDef.GetKind()), existingUpgradingChildDef.GetName())
```
juliev0, line 164: I realized the theoretical case that somebody has an Asset which contains a Pipeline named "my-event-consumer" and a MonoVertex named "my-event-consumer". We don't want these to have the same AnalysisRun names or there will be issues. Therefore, I am prefixing them with the kind.

dpadhiar (approved):
LGTM

PR #799, opened 2025-06-13
test: Add e2e tests for analysis template for monovertex and pipeline

chandankumar4:
Fixes #TODO


tests/e2e/analysistemplate.go
```diff
@@ -0,0 +1,76 @@
...
+				Fail("An unexpected error occurred when fetching the AnalysisTemplate: " + err.Error())
+			}
+			return false
+		}
+		if analysisRun.Status.Phase == expectedStatus {
+			return true
+		}
+		return false
+	}).Should(BeTrue())
+}
```
juliev0, line 76: thanks for breaking all of this out into a separate file!

juliev0 (commented):
It looks like these are only verifying the success cases for AnalysisTemplate and not the failure cases ? Could we verify the failure cases in a subsequent PR (if you prefer not to add to this one)?

chandankumar4:
> It looks like these are only verifying the success cases for AnalysisTemplate and not the failure cases ? Could we verify the failure cases in a subsequent PR (if you prefer not to add to this one)?

Yaa, I'll add it as a separate PR as this PR is getting bigger.

juliev0 (approved):

PR #804, opened 2025-06-23
feat: Use two AnalysisTemplates at same time

chandankumar4:
Fixes #TODO

### DIFF
--- a/tests/e2e/progressive-analysis-monovertex-e2e/analysis_monovertex_test.go
+++ b/tests/e2e/progressive-analysis-monovertex-e2e/analysis_monovertex_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package e2e
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


tests/e2e/progressive-analysis-monovertex-e2e/analysis_monovertex_test.go
```diff
@@ -135,10 +144,51 @@ var _ = Describe("Progressive MonoVertex E2E", Serial, func() {
...
+		By("Updating the MonoVertex Topology to cause a Progressive change Force promoted failure into success")
+		updatedMonoVertexSpec := updateMonoVertexRolloutForFailure()
+
+		VerifyMonoVertexProgressiveSuccess(monoVertexRolloutName, monoVertexScaleMinMaxJSONString, monoVertexScaleTo, updatedMonoVertexSpec,
+			0, 1, true, false)
+
+		// Verify the previously promoted monovertex was deleted
+		VerifyMonoVertexDeletion(GetInstanceName(monoVertexRolloutName, 0))
+
+		//VerifyAnalysisRunStatus(GetInstanceName(analysisRunName, 1), argov1alpha1.AnalysisPhaseFailed)
```
chandankumar4, line 170: I noticed that after running a failure case of `MonoVertexRollout`, I don't see any `analysisRunTemplate` gets created, @juliev0  Is that expected?

juliev0: If you do "ForcePromote = true", then yes, it will skip all of the usual analysis and declare success right away. Why do "ForcePromote" for this test?

chandankumar4: Oh! Okay, I was not aware about that it's setup in that way, Let me update it.

PR #808, opened 2025-07-04
fix: default timeout for Consistently() check should be much shorter 

juliev0:
### Modifications

I noticed a couple of places where I was calling `CheckConsistently()`, and unfortunately if you forget to override the default test timeout, it sets it to 6 minutes. I want to make the default much shorter.

This reduces the pipeline-progressive test from 32 minutes to 21 :) 

### Verification

e2e still passes, observed the Consistently check took only 15 seconds

n/a


chandankumar4 (approved):
lgtm

PR #817, opened 2025-07-08
chore: Feature flag to ignore AnalysisRun result

juliev0:
### Modifications

This adds a feature flag for ignoring the result of the AnalysisRun as part of the assessment of either Pipeline or MonoVertex.

I've also updated our sample MonoVertexRollout to use the AnalysisTemplate we're using on clusters.

### Verification

Did two tests: once with the feature flag turned on and once with it turned off. Verified that the result was ignored in the former case and not in the latter.

N/A

### DIFF
--- a/config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
+++ b/config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
@@ -6,11 +6,11 @@ metadata:
 spec:
   strategy:
     progressive:
-      assessmentSchedule: "60,60,10"
+      assessmentSchedule: "120,60,10"
     analysis:
       args:
       templates:
-      - templateName: mvtx-template
+      - templateName: mvtx-acknowledged-messages
         clusterScope: false
   riders:
     - definition:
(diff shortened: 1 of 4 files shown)


config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -41,8 +41,47 @@ spec:
             image: quay.io/numaio/numaflow-go/source-simple-source:stable
         transformer:
           container:
-            image: docker.intuit.com/quay-rmt/numaio/numaflow-rs/source-transformer-now:stable
+            image: quay.io/numaio/numaflow-rs/source-transformer-now:stable
```
juliev0, line 44: that shouldn't have been going to the Intuit proxy, even though it worked for us

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -6,11 +6,11 @@ metadata:
 spec:
   strategy:
     progressive:
-      assessmentSchedule: "60,60,10"
+      assessmentSchedule: "120,60,10"
```
juliev0, line 9: sometimes starting the window too early causes a failure

PR #818, opened 2025-07-09
fix: logging needs to reference the right name of the resource

juliev0:
### Modifications

I noticed this log line wasn't showing the correct name of the upgrading child. The issue is the location of the log line. 

### Verification

Verified log line while executing.

N/A

### DIFF
--- a/internal/controller/progressive/progressive.go
+++ b/internal/controller/progressive/progressive.go
@@ -486,7 +486,6 @@ func checkForUpgradeReplacement(
 		}
 
 		if differentFromPromoted {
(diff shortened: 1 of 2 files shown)


internal/controller/progressive/progressive.go
```diff
@@ -486,7 +486,6 @@ func checkForUpgradeReplacement(
 		}
 
 		if differentFromPromoted {
-			numaLogger.WithValues("old child", existingUpgradingChildDef.GetName(), "new child", newUpgradingChildDef.GetName()).Debug("replacing 'upgrading' child")
```
juliev0, line 489: The `newUpgradingChildDef` actually doesn't have the right name when logged up here (due to the way it's created)

PR #821, opened 2025-07-14
chore: incorporate numaflow version 1.5.2 into e2e 

juliev0:
### Modifications

Updating e2e test and sample NCR/NC to use latest numaflow.

@chandankumar4 figured out that by updating the Numaflow CRDs to use the latest 1.5.2, some of the progressive tests which were previously failing, started succeeding. (The only real difference between them being the existence of the ServingPipeline.)

### Verification

n/a

n/a


config/crd/kustomization.yaml
```diff
@@ -8,4 +8,4 @@ resources:
 - bases/numaplane.numaproj.io_monovertexrollouts.yaml
 - bases/numaplane.numaproj.io_numaflowcontrollers.yaml
 # install numaflow minimal CRDs as a dependency
-- https://github.com/numaproj/numaflow/config/advanced-install/minimal-crds?ref=stable
+- https://github.com/numaproj/numaflow/config/advanced-install/minimal-crds?ref=v1.5.2
```
juliev0, line 11: I guess we'll need to make sure we update this going forward as well.

On one hand, we could try "latest", but I wonder if that may cause its own failures.

PR #833, opened 2025-07-30
feat: When performing pause and drain, make sure Pipeline resumes with the number of replicas it had pre-pause

juliev0:
### Modifications

Numaflow is adding a new annotation called "numaflow.numaproj.io/resume-strategy" which can either be "fast" or "slow". If it's not set, it's defaulted to "fast".

We need to use "fast" during PPND, but afterward if the PipelineRollout says:
```
pipeline:
  metadata:
    annotations: 
      "numaflow.numaproj.io/resume-strategy": "slow"
```

then we must restore that. 

The thing I wanted to make sure of is that we only restore it once we're sure that Numaflow processed our request for "fast". For this, I make sure that only once the phase is no longer set to "Paused" or "Pausing", we take the Pipeline out of `upgradeInProgress: Pause-and-Drain`. Then after that, we can apply all of the usual changes to the Pipeline again, including setting  "numaflow.numaproj.io/resume-strategy": "slow" again. 

### Verification

I did a test in which I set PipelineRollout to this:
```
pipeline:
  metadata:
    annotations: 
      "numaflow.numaproj.io/resume-strategy": "slow"
```
I observed that the underlying Pipeline was updated with that annotation. 
Then I did a PPND update by updating topology. I observed that the annotation transitioned to "fast" during PPND and afterward went back to "slow". 

Before merging this, I will try it out with the new Numaflow to make sure it's all working. 

This will be okay for older numaflow versions because it will simply set an annotation that will be ignored.

I don't see any issues wrt backward incompatibility as far as numaplane resources or numaflow resources.

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_ppnd.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_ppnd.go
@@ -213,6 +213,7 @@ func (r *PipelineRolloutReconciler) needPPND(ctx context.Context, pipelineRollou
 // Determine if the Pipeline has changed and needs updating
 // We need to ignore any field that could be set by Numaplane in the pause-and-drain process as well as any labels or annotations that might be set directly on the Pipeline
 // by some Controller
+// TODO: can we replace this function with existing usde logic?
 func (r *PipelineRolloutReconciler) pipelineNeedsUpdatingForPPND(ctx context.Context, from, to *unstructured.Unstructured) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
 	fromCopy := from.DeepCopy()


internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -101,7 +110,7 @@ func (r *PipelineRolloutReconciler) processExistingPipelineWithPPND(ctx context.
 	}
 
 	// are we done with PPND?
-	doneWithPPND := !shouldBePaused
+	doneWithPPND := resumed
```
juliev0, line 113: if pipeline is resumed (no longer in Paused or Pausing), then we can return true from this function, which will enable us to no longer be in ppnd mode, and then if the "numaflow.numaproj.io/resume-strategy"=="slow", we will set it back to that then.

internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -215,13 +224,13 @@ func (r *PipelineRolloutReconciler) pipelineNeedsUpdatingForPPND(ctx context.Con
...
-	annotationsEqual := util.CompareMapsWithExceptions(from.GetAnnotations(), to.GetAnnotations(), common.KeyNumaplanePrefix)
-	numaLogger.Debugf("annotationsEqual (excluding Numaplane annotations): %t, from Annotations=%v, to Annotations=%v", annotationsEqual, from.GetAnnotations(), to.GetAnnotations())
+	// We need to restrict to just looking specifically at the labels and annotations we care about for data loss; otherwise, some platform (such as Numaflow) may set an annotation
+	// that we don't want to accidentally concern ourselves with
+	metadataRisk := usde.ResourceMetadataHasDataLossRisk(ctx, from, to)
+	numaLogger.Debugf("metadataRisk: %t, from=%v, to=%v\n",
+		metadataRisk, from.Object["metadata"], to.Object["metadata"])
 
-	return !specsEqual || !labelsEqual || !annotationsEqual, nil
+	return !specsEqual || metadataRisk, nil
```
juliev0, line 233: We need to make sure to ignore the new "numaflow.numaproj.io/resume-strategy" annotation is ignored when we're in PPND. We don't really need to be looking at all annotations and labels for determining if "pipeline needs to update" while we're in PPND mode. 

(Actually, it's on my TODO to see if we can just replace this function with something which just uses the existing USDE logic. I think the only reason it doesn't is because the idea was if we're already say pausing the pipeline due to isbsvc update, then if pipeline needs to update in any way (not just a data loss way), then we could simply update the pipeline now instead of waiting. But this may not be needed and could be made more simple).

afugazzotto (approved):
Discussed over call with Julie, approved.

PR #838, opened 2025-08-05
chore: Update Makefile to alternatively use podman and not just docker as container runtime

juliev0:
### Modifications

Since a couple of us are now trying out `podman` with `kind` as an alternative to `docker`, I wanted to include Makefile commands to accommodate that. 

For k3d option, that always uses docker. But for kind option, that could either use docker or podman. Users using podman, should set `CONTAINER_RUNTIME=podman`

### Verification

I am using podman with kind, and I ran `CONTAINER_RUNTIME=podman make image` and confirmed that it loaded the image.

@dpadhiar is using Rancher Desktop (docker with k3d) and he confirmed that when running `make image`, he saw it import the image.

N/A


juliev0:
Hey Chandan. Can you review and maybe try it out on your end? Thanks.

chandankumar4 (approved):
Verified in my local system with podman, it's working good!! Thanks for adding it.

PR #849, opened 2025-08-24
fix: consolidate redundant CONTAINER_RUNTIME and CONTAINER_TOOL variables in Makefile

juliev0:
### Modifications

After having added the CONTAINER_RUNTIME variable to the Makefile, I later realized there was already a variable for this purpose: CONTAINER_TOOL. I didn't want to have duplicate variables meaning the same thing so I decided to use the existing one.

### Verification

Ran `CONTAINER_TOOL=podman make start` and verified it worked

N/A


Makefile
```diff
@@ -194,7 +189,7 @@ clean:
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: image
 image: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMAGE_FULL_PATH} -f ${DOCKERFILE} .
+	$(CONTAINER_TOOL) build -t ${IMAGE_FULL_PATH} --load -f ${DOCKERFILE} .
```
juliev0, line 192: @chandankumar4 I saw you added this in your PR so I thought I'd do the same?

Makefile
```diff
@@ -78,20 +78,15 @@ else
...
 # CONTAINER_TOOL defines the container tool to be used for building images.
 # Be aware that the target commands are only tested with Docker which is
 # scaffolded by default. However, you might want to replace it to use other
 # tools. (i.e. podman)
 CONTAINER_TOOL ?= docker
-CONTAINER_TOOL:=$(shell command -v docker 2> /dev/null)
-ifndef CONTAINER_TOOL
-CONTAINER_TOOL:=$(shell command -v podman 2> /dev/null)
-endif
+
```
juliev0, line 89: Since some of us are aliasing "docker" to "podman", I didn't want to leave this in, because it will cause `CONTAINER_TOOL=docker`, and that will perform the wrong logic in `make image-import`.

PR #853, opened 2025-08-29
feat: Enable users to resume slow or fast for Pipeline and MonoVertex

### LINKED ISSUE #858: Support "slow resume" feature for Pipeline and MonoVertex
# Summary

Need to enable users to be able to resume their Pipelines and MonoVertices (i.e. when setting `desiredPhase=Running` after it was set to "Paused") either gradually or quickly depending on configuration.

Need to also have configuration for the case of PPND (Pipeline) (when pipeline is resumed after pause).

---

juliev0:
Fixes #858 

### Modifications

This feature is for Numaplane to make it so that Pipelines and MonoVertices can be unpaused either gradually or quickly, depending on user preference. "Gradually" means the Vertices/MonoVertices start from their "min" and autoscale back up. "Quickly" means they don't.

The MonoVertex and the Vertex both have a field "replicas" which indicates the number of Pods they should have. Upon pausing, this value will remain the same. To unpause "quickly", we just need to set `desiredPhase: Running` and the `replicas` will already be what it needs to be. To unpause "gradually", we need to reset the `replicas` value to null, which will be interpreted by numaflow as "min". 

A previous [PR](https://github.com/numaproj/numaplane/pull/833/files#diff-00bcf30e52375230575763d7076594d12990dd5c3e8b5654d528d558ac664d35) made it so that we handle pipeline "unpause gradual" through an annotation called "resume-strategy". What this was doing was causing Numaflow Pipeline Controller to receive the "resume-strategy" annotation, and if it was set to "slow", then it sets `replicas=nil` for each Vertex. However, we can do that ourselves, which enables us to have parity with how we do MonoVertex.

This PR does 3 things:
1. Intercept `desiredPhase=Running` for MonoVertexRollout to set `replicas=nil` if strategy.pauseResume.fastResume==slow
2. Intercept `desiredPhase=Running` for PipelineRollout to set `replicas=nil` for each Vertex  if strategy.pauseResume.fastResume==slow
3. PPND for Pipeline: When resuming set `replicas=nil` for each Vertex  if strategy.ppnd.fastResume==slow 

### Verification

e2e test has been augmented to include the following tests:
- MonoVertex: 
-- leave strategy unset and unpause (default is gradual)
-- set strategy.pauseResume.fastResume==fast and unpause
- Pipeline:
-- leave strategy unset and unpause (default is gradual)
-- set strategy.pauseResume.fastResume==fast and unpause
-- ppnd: set fastResume = slow and unpause, set fastResume = fast and unpause

N/A


internal/controller/common/numaflowtypes/pipeline.go
```diff
@@ -171,6 +172,13 @@ func CheckIfPipelineWontPause(ctx context.Context, pipeline *unstructured.Unstru
 
 func GetPipelineDesiredPhase(pipeline *unstructured.Unstructured) (string, error) {
 	desiredPhase, _, err := unstructured.NestedString(pipeline.Object, "spec", "lifecycle", "desiredPhase")
+	if err != nil {
+		return desiredPhase, err
+	}
+
+	if desiredPhase == "" {
+		desiredPhase = string(numaflowv1.PipelinePhaseRunning)
+	}
```
juliev0, line 181: this should've been in here before - since "" is equivalent to Running

internal/controller/pipelinerollout/pipelinerollout_ppnd.go
```diff
@@ -64,23 +64,15 @@ func (r *PipelineRolloutReconciler) processExistingPipelineWithPPND(ctx context.
...
 	if shouldBePaused {
 		if err := r.setPipelineLifecyclePaused(ctx, existingPipelineDef); err != nil {
 			return false, err
 		}
 	} else {
-		// make sure this is set to running
 		if err := r.setPipelineLifecycleRunning(ctx, pipelineRollout, existingPipelineDef, forceResume); err != nil {
 			return false, err
 		}
-
```
juliev0, line 78: this was code I had to put in in [this](https://github.com/numaproj/numaplane/pull/833/files#diff-00bcf30e52375230575763d7076594d12990dd5c3e8b5654d528d558ac664d35) PR, but it was only necessary because we were making a change to the pipeline "resume-strategy" then as the way to control the Pipeline and we needed to make sure to reset it afterward. We no longer need to do any of that.

PR #869, opened 2025-09-09
chore: various minor modifications to rolling window PR

juliev0:
### Modifications

A few small modifications, described inline. 

### Verification

N/A - CI is sufficient

N/A


tests/manifests/e2e/special-cases/no-strategy/controller-config.yaml
```diff
@@ -10,13 +10,4 @@ data:
...
     defaultUpgradeStrategy: "no-strategy"
-    progressive:
-      defaultAssessmentSchedule:
-        - kind: Pipeline
-          schedule: "120,360,60,10"
-        - kind: MonoVertex
-          schedule: "120,360,60,10"
-        - kind: InterstepBufferService
-          schedule: "0,0,0,10"
-      analysisRunTimeout: "1200"
```
juliev0, line 21: I noticed that for the ppnd ConfigMap we're not setting this at all, so for parity I'm making this one the same.

config/install.yaml
```diff
@@ -2625,9 +2625,9 @@ data:
     progressive:
       defaultAssessmentSchedule:
         - kind: Pipeline
-          schedule: "120,360,60,10"
+          schedule: "10,360,60,10"
```
juliev0, line 2628: if we do testing manually, it will be nice if the window starts earlier

chandankumar4 (approved):

PR #872, opened 2025-09-14
fix: make test less rigid as far as order of operations

juliev0:
partial fix to #873 

### Modifications

The Progressive Pipeline test updates isbsvc and pipeline at the same time. We need to make the test less rigid as far as what order updates occur in. 

This is a partial fix to the issue for this test overall, since I noticed a failure for this particular test case.

### Verification

ran CI a few times


chandankumar4 (approved):
LGTM

Need to fix the code coverage issue, failing for golang version
`golang.org/x/tools@v0.37.0 requires go >= 1.24.0 (running go 1.23.12)`

Issue #879, opened 2025-09-19
Determine how to handle pipeline phase=Failed in the case of a Force Drain


juliev0:
# Summary

A Pipeline that is being force drained - if it's in "Failed" phase, currently we delete it right away. Need to determine if there's a possibility that this failure could be intermittent. If so, do we want to wait until pauseGracePeriodSeconds elapses instead? Or is it unnecessary?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #900, opened 2025-10-27
fix: Add wait time before deleting pipeline in transient failure during force drain

chandankumar4:
Fixes https://github.com/numaproj/numaplane/issues/879

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_recycle.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_recycle.go
@@ -168,19 +168,11 @@ func (r *PipelineRolloutReconciler) registerFinalDrainStatus(namespace, pipeline
 // apply a spec that's considered valid (from a promoted pipeline) over top a spec that's not working.
 // The new spec will enable it to drain.
 // Then pause it.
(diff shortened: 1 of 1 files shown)


internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -229,14 +231,45 @@ func (r *PipelineRolloutReconciler) forceDrain(ctx context.Context,
...
+		return r.checkForPipelineTransientFailures(ctx, c, pipelineRollout, pipeline)
+	}
+
+	return false, nil
+}
+
+// checkForPipelineTransientFailures checks if the Pipeline has been in Failed state for long enough to consider it a permanent failure
+// if so, delete it; otherwise, return false to indicate we haven't deleted it yet.
+// decision: just use original failure start time in the case of Pipeline switching Failed->Running->Failed
+func (r *PipelineRolloutReconciler) checkForPipelineTransientFailures(ctx context.Context, c client.Client,
```
juliev0: nit: do you mind if we call it "checkForFailedPipeline()" or something like that? (since it's doing more than just looking for the transient aspect of the failure, it's really looking for failures in general)

PR #904, opened 2025-10-31
feat: if we aren't doing full progressive then we don't need to do force draining

juliev0:
### Modifications

If the user's strategy is set to skip the assessment of Progressive rollout, then we really have no idea whether a "promoted" pipeline is good for force draining. The fact that they have this strategy also indicates that they don't care about data loss. 

### Verification

Verified that with ForcePromote strategy set to true, we skip draining and just delete the resource.

N/A


dpadhiar (approved):
LGTM

PR #905, opened 2025-11-01
fix: comparison with both promoted child definition and with upgrading child definition

juliev0:
#### During Progressive Rollout:

This fixes the comparisons which are done during Progressive rollout to:
1. compare the Rollout spec to the current "in-process" child to determine if it needs to be replaced
2. compare the Rollout spec to the current "promoted" child - if they're equal, then the "in-process" one can simply be deleted.

The problem was specifically related to templating of:
- the child spec itself
- the metadata (labels and annotations)

**Whenever we compare two things, we must make sure that they are templated with the same child name.** 

#### Outside of Progressive Rollout, when we compare the rollout definition to the "promoted" child:

This also fixes the standard comparison which is done outside of during progressive rollout which is in usde.go which compares the Rollout definition with the "promoted" child to determine if they are different. The issue was that in the comparison of metadata, we were doing a comparison to determine if the annotations were equal, and if the labels were equal. We cannot do that, or we will start to take into account the labels/annotations that are applied outside of what's defined in the rollout (e.g. "numaflow.numaproj.io/pause-annotation"). Rather, we can only check that the labels and annotations required by the rollout are present. 

....

#### Manual Verification

I performed the following test for both pipeline and monovertex:
I created a templated annotation (i.e. including `{{.monovertex-name}}` or `{{.pipeline-name}}`) in the Rollout definition, then did a progressive upgrade to cause a second child to be created, verified the annotation was correct, then restored the original Rollout definition, which effectively deleted the new child. 

##### Testing templating of metadata and child spec
I updated the functional tests for monovertex and pipeline to use templating for both labels and annotations and templating of the spec itself. The tests verify the content. 

##### Testing templating of Riders
I updated the Rider tests for monovertex and pipeline to check the templated content.

N/A

### DIFF
--- a/internal/usde/usde.go
+++ b/internal/usde/usde.go
@@ -385,18 +385,6 @@ func getNamesAndKinds(ulist unstructured.UnstructuredList) string {
 	return namesAndKinds
 }
 
-func checkMapsEqual(map1 map[string]string, map2 map[string]string) bool {
-	tempMap1 := map1
-	if tempMap1 == nil {
-		tempMap1 = map[string]string{}
-	}
-	tempMap2 := map2
-	if tempMap2 == nil {
-		tempMap2 = map[string]string{}
-	}
-	return util.CompareStructNumTypeAgnostic(tempMap1, tempMap2)
-}
-
 // return the upgrade strategy that represents what the user prefers to do when there's a concern for data loss
 func getDataLossUpgradeStrategy(ctx context.Context, namespace, resourceKind string) (apiv1.UpgradeStrategy, error) {
 	userUpgradeStrategy, err := GetUserStrategy(ctx, namespace, resourceKind)


internal/controller/progressive/progressive.go
```diff
@@ -471,7 +478,16 @@ func checkForUpgradeReplacement(
...
+
+	existingUpgradingChildName := existingUpgradingChildDef.GetName()
+	// we need to create an Upgrading child definition which is template-evaluated using the existing Upgrading child's name so that we can effectively compare them below
+	newUpgradingChildDef, err := makeUpgradingObjectDefinition(ctx, rolloutObject, controller, c, &existingUpgradingChildName)
+	if err != nil {
+		return false, false, err
+	}
+	// we need to create an Upgrading child definition which is template-evaluated using the existing Promoted child's name so that we can effectively compare them below
+	existingPromotedChildName := existingPromotedChildDef.GetName()
+	newUpgradingChildDefUsingPromotedName, err := makeUpgradingObjectDefinition(ctx, rolloutObject, controller, c, &existingPromotedChildName)
```
juliev0: This is one of the core changes of the PR. When we compare the "promoted" child to the PipelineRollout, we must make sure that the PipelineRollout-defined spec and metadata are all evaluated using the "promoted" child's name so we can compare

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -8,7 +8,6 @@ spec:
     progressive:
       assessmentSchedule: "10,200,60,10"
     analysis:
-      args:
```
juliev0: my IDE didn't seem to like this so I just deleted it

dpadhiar (approved):

PR #906, opened 2025-11-03
fix: usde check for whether Direct Apply is required for metadata differences must ignore labels and annotations applied outside of the rollout definition

juliev0:
Fixes #TODO


juliev0:
duplicate

PR #908, opened 2025-11-04
fix: add volume mount as a data loss/progressive field in USDE 

juliev0:
### Modifications

There are 2 motivations for this:

1. previously, we've listed not only "image" but also things like "env" and "args" as potential data loss fields for pipeline and progressive fields for monovertex. That is because those things can affect the behavior of the image running. Likewise, a volumeMount of either a ConfigMap, a Secret, or a PVC could also affect the behavior. Therefore, I am adding that to our previous lists.
2. Within Intuit, this will help with our end to end testing from UI to Numaplane to test the progressive rollout path every time. Whenever a PR is generated, it makes this volumeMount change. Therefore, whenever we do end-to-end testing, it will always cause a Progressive rollout. (see [here](https://github.intuit.com/ODL/odl-metrics-es-streamer-deployment/commit/4b06a1549753d44c7429d72424550aa3ba83784a) for an example of a PR)

### Verification

Did a change to just the volumeMount and observed that it did progressive rollout. 

N/A

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -2715,6 +2715,8 @@ data:
       - path: spec.source.udsource.container.command
       - path: spec.source.udsource.container.args
       - path: spec.source.udsource.container.env
+      - path: spec.volumes
+        includeSubfields: true
   numaflowcontroller: |
     dataLoss:
      - path: spec.version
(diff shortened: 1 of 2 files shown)


tests/e2e/monovertex-rider-e2e/monovertex_rider_test.go
```diff
@@ -185,13 +185,17 @@ var _ = Describe("Rider E2E", Serial, func() {
 			return rollout, nil
 		})
 
-		// verify ConfigMap is created
+		monoVertexIndex++
```
juliev0: Now the volumeMounts change to the spec causes there to be a progressive rollout, so we need to check the new index

afugazzotto (approved):

PR #920, opened 2025-11-24
fix: if an environment variable changes, this should cause progressive upgrade

juliev0:
### Modifications

I noticed certain Environment variable-related fields were missing from the USDE file (as a "dataLoss" field for Pipeline and a "progressive" field for MonoVertex). And certain others were present, but they were missing `includeSubFields: true` (which is necessary in order for the fields that are subfields (i.e. `env[x].name`, `env[x].value`) to be considered to be "dataLoss"/"progressive" fields themselves).

**This is why we want them there:**

If an environment variable on a container changes, it could conceivably change the way that data is processed on that image, which means:
- we should ideally perform progressive upgrade with it
- even in the case of PPND, we can consider it a possible "data loss" change, since it alters the way the container processes data

**Reformatting of the install.yaml file:**

I noticed that the install.yaml file was being created by kustomize in an unreadable way - some of the lines which were previously nicely formatted yaml on multiple lines were being written instead as lines containing "\n" characters. Cursor generated a script to improve the formatting so that the file is readable.

### Verification

N/A

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,7 @@ help: ## Display this help.
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) crd paths="./..." output:crd:artifacts:config=config/crd/bases
 	$(KUBECTL) kustomize config/default > config/install.yaml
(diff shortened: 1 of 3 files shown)


chandankumar4 (approved):
lgtm

PR #925, opened 2025-12-02
feat: Retry force drain multiple times

### LINKED ISSUE #901: Attempt force drain multiple times
# Summary

We may not succeed after the first force drain. We should try again if that happens, up until some timeout in which we stop. 

---

juliev0:
Fixes #901 

### Modifications

Previously, when draining we did the following:
1. Drain with original spec first (or if Pipeline failed, we skip this step). Allow it to go to "Paused", which either means it did drain or it timed out
2. If it did drain, delete it
3. Else wait for a new "promoted" Pipeline to show up and force drain with that

This PR changes step 3 to this:
3. Else wait for a new "promoted" Pipeline to show up and force drain with that
4. If that reaches "Paused" without draining, then wait for another "promoted" Pipeline to show up: in the meantime, scale to 0 Pods
5. numaplane-controller-config has a field called `maxRecyclableDurationMinutes` - if we reach that, we delete the Pipeline

We introduce 2 new annotations on the Pipeline which keep track of both which specs we've applied and started draining with, as well as those that we've already completed:
- `numaplane.numaproj.io/force-drain-specs-started`
- `numaplane.numaproj.io/force-drain-specs-completed`

### Verification

I first ran this manually and then updated the force-drain-e2e test to do it:

- Create PipelineRollout - "promoted" Pipeline-0
- Update PipelineRollout to failed spec - "upgrading" Pipeline-1
- Update PipelineRollout to second failed spec - "upgrading" Pipeline-2
- Force promote Pipeline-2. This causes Pipeline-0 and Pipeline-1 to drain. Pipeline-1 attempts to drain with the spec from Pipeline-2, which doesn't succeed. 
- Update PipelineRollout to another failed spec - "upgrading" Pipeline-3. Force promote it. This causes Pipeline-1 and Pipeline-2 to attempt to drain with Pipeline-3's spec, which doesn't succeed. Observe both Pipelines scale down waiting. 
- Fix PipelineRollout - "upgrading" Pipeline-4. Other Pipelines now try to force drain with this spec, which succeeds. Verify that by using the metric.

One thing this PR does is remove the annotation `numaplane.numaproj.io/overridden-spec` which indicated if we had updated the spec with the new one. Instead, there are two new annotations:

- `numaplane.numaproj.io/force-drain-specs-started`
- `numaplane.numaproj.io/force-drain-specs-completed`

Let's say that there's a Pipeline that's in the middle of a force drain with `numaplane.numaproj.io/overridden-spec: true`. What will happen if we update Numaplane at that time?: 

- We will see that `numaplane.numaproj.io/force-drain-specs-started` does not include the new "promoted" Pipeline, so we will try to apply its spec even though we already have. We will toggle the Pipeline back to "desiredPhase: Running" briefly (with 0 Source Pods so no incoming data), and then set it to back to "desiredPhase: Paused". 

This seems okay.

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_recycle.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_recycle.go
@@ -290,7 +290,7 @@ func (r *PipelineRolloutReconciler) forceDrain(ctx context.Context,
 			numaLogger.WithValues("promotedPipeline", promotedPipeline.GetName()).Infof("Pipeline has the promoted pipeline's spec but was not able to drain")
 		}
 	}
-	// If the Pipeline failed during force drain,, we need to wait some time before deleting it, as there may be transient failures.
+	// If the Pipeline failed during force drain, we need to wait some time before deleting it, as there may be transient failures.
 	if failed {
 		nonTransientFailure, err := r.checkForFailedPipeline(ctx, pipeline)
 		if err != nil {
@@ -720,12 +720,16 @@ func checkForValueInCommaDelimitedAnnotation(pipeline *unstructured.Unstructured
 		return false
 	}
 
-	targetPattern := value + ","
-	return strings.Contains(forceDrainSpecs, targetPattern)
+	for _, p := range strings.Split(forceDrainSpecs, ",") {
+		if strings.TrimSpace(p) == value {
+			return true
+		}
+	}
+	return false
 }
 
 // we use an annotation to indicate when we've started force draining with a promoted pipeline's spec
-// if the annotation is unset, then the pipeline spec is still theoriginal
+// if the annotation is unset, then the pipeline spec is still the original
 func isPipelineSpecOriginal(pipeline *unstructured.Unstructured) bool {
 	forceDrainSpecs, found := pipeline.GetAnnotations()[common.AnnotationKeyForceDrainSpecsStarted]
 	return !found || forceDrainSpecs == ""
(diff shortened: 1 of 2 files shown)


internal/controller/common/rollout_controller.go
```diff
@@ -169,26 +168,35 @@ func FindMostCurrentChildOfUpgradeState(ctx context.Context, rolloutObject Rollo
...
-		patchJson = `{"metadata":{"labels":{"` + common.LabelKeyUpgradeState + `":"` + string(upgradeState) + `","` + common.LabelKeyUpgradeStateReason + `":"` + string(*upgradeStateReason) + `"}}}`
-	} else {
-		patchJson = `{"metadata":{"labels":{"` + common.LabelKeyUpgradeState + `":"` + string(upgradeState) + `"}}}`
+		labelsToSet[common.LabelKeyUpgradeStateReason] = string(*upgradeStateReason)
 	}
-	childObject.SetLabels(labels)
-	return kubernetes.PatchResource(ctx, c, childObject, patchJson, k8stypes.MergePatchType)
+
+	// Patch the labels in Kubernetes
+	return kubernetes.SetAndPatchLabels(ctx, c, childObject, labelsToSet)
```
juliev0: none of this logic changed, just consolidated into a function

internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -508,6 +584,9 @@ func calculateScaleForRecycle(
 				originalPodsRunning, found := historicalPodCount[vertexName]
 				if found {
 					newScaleValue = int64(math.Ceil(float64(originalPodsRunning) * float64(percent) / 100.0))
+					if newScaleValue < 1 {
+						newScaleValue = 1
+					}
```
juliev0: make sure that we don't allow the value to be < 1, which would prevent draining from working

PR #941, opened 2025-12-16
fix: if user updates their scale.max=0 and we're in the middle of Progressive upgrade, we need to notice it

### LINKED ISSUE #940: Need to recognize when source max is zero when in the middle of progressive upgrade
**Describe the bug**
If we're in the middle of a Progressive Upgrade (such as in a failure state), currently the CheckForDifferences() function for both PipelineRollout and MonoVertexRollout will ignore if the user sets the scale to 0. 

---

juliev0:
Fixes #940 

### Modifications

There are 2 changes:
1. We need to make sure that we **notice** it if somebody updates their Rollout from non-zero scale to zero. Therefore, our existing `CheckForDifferences()` functions which ignore scale altogether when comparing the existing Upgrading child to the Rollout definition won't work. Therefore, what I am doing is introducing a new field to our Upgrading Child Status called `OriginalScaleDefinition` for MonoVertex, and `OriginalScaleDefinitions` for Pipeline. This is the original field that the upgrading child had before it was scaled down, so we can effectively compare that to the current Rollout definition.
2. When creating the new Upgrading child, if the user wants 0, we scale it to 0. 

### Verification

**Manual tests:**
I ran a test in which I had a promoted pipeline and a trial pipeline. I then updated the PipelineRollout to set the source's Vertex min=max=0. This caused a new trial pipeline with the source vertex min=max=0 to replace the original and immediately be force promoted. 

I did the same test for MonoVertexRollout as well. 

I also ran the backward compatibility test mentioned below.

**New unit tests:**
Any non-trivial functions I added have unit tests.

**e2e test:**
I updated the skip-progressive e2e test such that what it does now is create a situation of a promoted and trial pipeline, and then simply update the scale. 

This logic depends on the value for `OriginalScaleDefinition` for MonoVertex and `OriginalScaleDefinitions` for Pipeline. Older Trial Pipelines and MonoVertices won't have them. Therefore, what I've done is first check to see if the field is set ([here](https://github.com/numaproj/numaplane/blob/check-differences-0/internal/controller/pipelinerollout/pipelinerollout_progressive.go#L236-L244) for Pipeline, [here](https://github.com/numaproj/numaplane/blob/check-differences-0/internal/controller/monovertexrollout/monovertexrollout_progressive.go#L181-L193) for MonoVertex). If it's not, I basically set it to whatever the Rollout says at that time. This should generally be sufficient. It is possible that there will be a Trial Pipeline that hasn't yet caught up to the latest scale definition in the Rollout, but in that case, the only danger is that we will say the two are the same when they're different, which is exactly equivalent to what the code was doing previously anyway. (That only matters when we're trying to go to 0.)

I actually tested the backward compatibility as well. For both PipelineRollout and MonoVertexRollout, I created a situation of a "promoted" and a failed "trial" child while in the main branch. Then I updated to this branch and observed that the new `originalScaleDefinition` got populated.

### DIFF
--- a/tests/e2e/skip-progressive-e2e/skip_progressive_test.go
+++ b/tests/e2e/skip-progressive-e2e/skip_progressive_test.go
@@ -61,11 +61,7 @@ var (
 		},
 	}
 
-<<<<<<< HEAD
 	invalidUDTransformerImage = "quay.io/numaio/numaflow-rs/source-transformer-now:invalid"
-=======
-	validUDTransformerImage = "quay.io/numaio/numaflow-rs/source-transformer-now:stable"
->>>>>>> origin/main
 
 	pipelineSpecSourceRPU      = int64(5)
 	pipelineSpecSourceDuration = metav1.Duration{


config/crd/bases/numaplane.numaproj.io_pipelinerollouts.yaml
```diff
@@ -458,6 +458,10 @@ spec:
                       name:
                         description: Name of the upgrading child
                         type: string
+                      originalScaleDefinitions:
+                        items:
+                          type: string
+                        type: array
```
juliev0: these are the "full scale definitions", defined as an array, since it's per vertex

internal/controller/common/numaflowtypes/pipeline.go
```diff
@@ -138,49 +139,46 @@ func CheckPipelineDrained(ctx context.Context, pipeline *unstructured.Unstructur
...
-					}
+				maxValue, maxValid := util.ToInt64(maxInterface)
+				// If max is found and is not 0, it's not scaled to zero
+				if maxValid && maxValue != 0 {
+					return false, nil
 				}
 			}
 		}
 	}
+	return true, nil
```
juliev0: this is effectively unit tested in `Test_CanPipelineIngestData()` function

internal/controller/pipelinerollout/pipelinerollout_progressive.go
```diff
@@ -472,48 +550,61 @@ func createScaledDownUpgradingPipelineDef(
 		return err
 	}
 
-	// map each vertex name to new min/max, which is based on the number of Pods that were removed from the corresponding
```
juliev0: none of this code below has changed - it's just indented because of a new if/else in front of it

PR #942, opened 2025-12-16
fix: move recycle expiration logic to apply in all cases

juliev0:
### Modifications

I realized there's a case in which we're not honoring the Recycle Expiration time [here](https://github.com/numaproj/numaplane/blob/main/internal/controller/pipelinerollout/pipelinerollout_recycle.go#L110). In this case, if the user has paused their PipelineRollout, the Pipelines will accumulate for a long time needing to be deleted. If we're going to have a Recycle Expiration time, I think it should apply in all cases. 

Therefore, I've moved the logic for checking that time so it happens as the first thing.

### Verification

I tested the case of a Recyclable Pipeline in which the PipelineRollout is paused. Previously, the Recyclable Pipeline would've stayed there despite having reached the recyclable expiration time, but this time it did not. 

Also added a unit test for the new function `shouldDeleteRecyclablePipeline()`.

n/a - this will delete any Pipelines that have been marked "recyclable" for awhile that haven't been deleted, which is what I'm looking for.

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_recycle.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_recycle.go
@@ -727,7 +727,7 @@ func (r *PipelineRolloutReconciler) shouldDeleteRecyclablePipeline(
 		return true, nil
 	}
 
-	expired, err := r.checkForRecycleExpirationTime(ctx, pipeline)
+	expired, err := r.isRecycledPipelineExpired(ctx, pipeline)
 	if err != nil {
 		return false, err
 	}
@@ -740,9 +740,9 @@ func (r *PipelineRolloutReconciler) shouldDeleteRecyclablePipeline(
 	return false, nil
 }
 
-// checkForRecycleExpirationTime checks if the max recyclable duration has been exceeded.
+// isRecycledPipelineExpired checks if the max recyclable duration has been exceeded.
 // Returns true if the pipeline should be deleted due to expiration.
-func (r *PipelineRolloutReconciler) checkForRecycleExpirationTime(
+func (r *PipelineRolloutReconciler) isRecycledPipelineExpired(
 	ctx context.Context,
 	pipeline *unstructured.Unstructured) (bool, error) {
 	numaLogger := logger.FromContext(ctx)


internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -77,23 +77,18 @@ func (r *PipelineRolloutReconciler) Recycle(
...
-		numaLogger.Info("Pipeline will be deleted now")
+		r.registerFinalDrainStatus(pipelineRollout.Namespace, pipelineRollout.Name, pipeline, false, metrics.LabelValueDrainResult_DrainNotRequired)
 		err = kubernetes.DeleteResource(ctx, r.client, pipeline)
 		return true, err
 	}
 
-	// if pipeline doesn't require drain then just delete it
-	if pipeline.GetAnnotations() == nil || pipeline.GetAnnotations()[common.AnnotationKeyRequiresDrain] != "true" {
-		numaLogger.Info("Pipeline does not require drain and will be deleted now")
+	shouldDelete, err := r.shouldDeleteRecyclablePipeline(ctx, pipelineRollout, pipeline)
```
juliev0: consolidated the various cases of needing to delete into one function

internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -734,3 +704,70 @@ func isPipelineSpecOriginal(pipeline *unstructured.Unstructured) bool {
...
+		r.registerFinalDrainStatus(pipelineRollout.Namespace, pipelineRollout.Name, pipeline, false, metrics.LabelValueDrainResult_NeverDrained)
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// checkForRecycleExpirationTime checks if the max recyclable duration has been exceeded.
+// Returns true if the pipeline should be deleted due to expiration.
+func (r *PipelineRolloutReconciler) checkForRecycleExpirationTime(
```
chandankumar4: `isRecycleExpired()` fn name can be more meaningful?

juliev0: cool, I like the conciseness of that, will change

PR #947, opened 2025-12-18
fix: update USDE paths to include some that were missing and remove duplicates

juliev0:
### Modifications

fix: update USDE paths to include some that were missing and remove duplicates

### Verification

e2e

n/a


config/install.yaml
```diff
@@ -2755,9 +2761,6 @@ data:
         includeSubfields: true
       - path: spec.templates     # note: this one could be broken down further into just the fields that are at risk
         includeSubfields: true
-      
-      - path: spec.sideInputs.container.envfrom
-        includeSubfields: true
```
juliev0: duplicate

PR #988, opened 2026-05-05
fix: update USDE to include all of the built-in sources and sinks

juliev0:
### Modifications

I noticed we weren't triggering Progressive Rollout when changing SQS fields. This caused me to look at the numaflow pipeline/monovertex spec and see that more Sources and Sinks have been added since we wrote the USDE config. For simplicity, I ended up setting all of them to trigger Progressive when any subfield of them has changed. 

I also ended up fixing a Snyk issue (go.mod change), as well as a bug related to the script which reformats our ConfigMaps.

### DIFF
--- a/config/manager/usde-config.yaml
+++ b/config/manager/usde-config.yaml
@@ -52,9 +52,10 @@ data:
         includeSubfields: true
       - path: spec.vertices.source.transformer.builtin
         includeSubfields: true
-      - path: spec.vertices.sink.log
       - path: spec.vertices.sink.kafka
-      - path: spec.vertices.sink.blackhole
+      - path: spec.vertices.sink.serve
+      - path: spec.vertices.sink.sqs
+      - path: spec.vertices.sink.pulsar
       - path: spec.vertices.sink.udsink.container.image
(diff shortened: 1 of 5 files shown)


config/manager/usde-config.yaml
```diff
@@ -52,9 +52,10 @@ data:
         includeSubfields: true
       - path: spec.vertices.source.transformer.builtin
         includeSubfields: true
-      - path: spec.vertices.sink.log
       - path: spec.vertices.sink.kafka
-      - path: spec.vertices.sink.blackhole
```
juliev0, line 57: blackhole and log shouldn't matter for data loss

hack/fix-configmap-formatting.py
```diff
@@ -88,7 +88,7 @@ def main():
         
         # Scan backward to find the start of this ConfigMap
         configmap_start = None
-        for i in range(label_idx - 1, max(0, label_idx - 100), -1):
```
juliev0, line 91: fix bug involving ConfigMap parsing which was limiting the amount of the file that we scanned

PR #998, opened 2026-06-10
fix: use more current gitops-engine from argocd repo

juliev0:
### Modifications

Updates to using the latest gitops-engine from the argocd repo instead of from the outdated gitops-engine repo.

Given that various imports got updated, there were some other changes that had to be made to go along with them.


internal/controller/numaflowcontroller/numaflowcontroller_controller.go
```diff
@@ -512,11 +512,7 @@ func (r *NumaflowControllerReconciler) sync(
 	opts := []gitopsSync.SyncOpt{
 		gitopsSync.WithLogr(*numaLogger.LogrLogger),
 		gitopsSync.WithOperationSettings(false, true, true, false),
-		// Temporary: don't perform per-resource validation by hitting OpenAPI due to authentication issue.
-		// OpenAPI schema validation uses a temp kubeconfig in gitops-engine that can fail to
-		// authenticate to /openapi/v2 even when the controller's rest.Config works (e.g. after
-		// cluster auth changes). Equivalent to kubectl apply --validate=false.
-		gitopsSync.WithManifestValidation(false),
```
juliev0, line 519: restore back

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -313,16 +313,16 @@ func (r *PipelineRolloutReconciler) processQueueKey(ctx context.Context, key str
 	if err != nil {
 		numaLogger.Errorf(err, "PipelineRollout %v reconcile returned error: %v", namespacedName, err)
 		r.Queue.AddRateLimited(key)
-	} else {
-		if result.Requeue {
```
juliev0, line 317: apparently `Requeue` field is deprecated in the newer version of controller-runtime. Recommendation is to instead look for `!result.IsZero()`

go.mod
```diff
@@ -130,84 +136,87 @@ require (
...
-	k8s.io/kubectl => k8s.io/kubectl v0.32.2
-	k8s.io/kubelet => k8s.io/kubelet v0.32.2
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.32.2
-	k8s.io/metrics => k8s.io/metrics v0.32.2
-	k8s.io/mount-utils => k8s.io/mount-utils v0.32.2
-	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.32.2
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.32.2
-	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.32.2
-	k8s.io/sample-controller => k8s.io/sample-controller v0.32.2
+	github.com/argoproj/argo-cd/gitops-engine => github.com/argoproj/argo-cd/gitops-engine v0.0.0-20260528113041-1801122b4391
```
juliev0, line 188: `v0.0.0-20260528113041-1801122b4391` is the commit [here](https://github.com/argoproj/argo-cd/blob/1801122b4391cad4961301f787006dc9a88c2dd3/gitops-engine/go.mod). You can see that the version of kubectl which that uses is [0.34.0](https://github.com/argoproj/argo-cd/blob/1801122b4391cad4961301f787006dc9a88c2dd3/gitops-engine/go.mod#L131)

go.mod
```diff
@@ -1,126 +1,132 @@
 module github.com/numaproj/numaplane
 
-go 1.25.10
+go 1.26.0
 
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
-	github.com/argoproj/argo-cd/v3 v3.0.23
+	github.com/argoproj/argo-cd/gitops-engine v0.7.1-0.20250908182407-97ad5b59a627
```
juliev0, line 7: this line is replaced with a different (more current) version below (see [comment](https://github.com/numaproj/numaplane/pull/998/changes#r3391563810)).

I asked Cursor if we could just use that version up here instead of needing to do the replace, but apparently that doesn't work due to the fact that there's some [erroneous line](https://github.com/argoproj/argo-cd/blob/1801122b4391cad4961301f787006dc9a88c2dd3/go.mod#L15) in ArgoCD's go.mod which is requiring this version.

whynowy (approved):

PR #1000, opened 2026-06-15
fix: force gitops-engine to use BearerTokenFile rather than BearerToken (cached file) for authenticating to Kubernetes API

juliev0:
### Modifications

Fixes error failing to authenticate to Kubernetes API.

Issue appears to be seen when there is ServiceAccount token rotation happening on the cluster. 

**Background:** 
So, in Kubernetes, there is a token (associated with the ServiceAccount) provided by kubelet to every Pod for it to authenticate to Kubernetes API. This token rotates once an hour. It is provided by kubelet in the form of a file named `/var/run/secrets/kubernetes.io/serviceaccount/token`.
(When a pod starts, the kubelet automatically add a volume and mount so that it can update this file.)

In the gitops-engine code (from ArgoCD) which Numaplane uses, it caches the value of this file in memory when the Pod starts. So, when the token rotates after an hour, it's now using the old cached value.

The fix is to clear that cached value in the Kube restConfig which is passed to gitops-engine each time it does a sync so that it will be forced instead to look at the latest and greatest file value instead.

### Verification

e2e tests

### DIFF
--- a/internal/util/kubernetes/kubectl.go
+++ b/internal/util/kubernetes/kubectl.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/argoproj/argo-cd/gitops-engine/pkg/utils/kube"
 	"github.com/argoproj/argo-cd/gitops-engine/pkg/utils/tracing"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubectl/pkg/util/openapi"
 
 	"github.com/numaproj/numaplane/internal/util/logger"
 )
@@ -17,6 +19,16 @@ func init() {
 	}
 }
 
+type kubectlCmd struct {
+	*kube.KubectlCmd
+}
+
+func (k *kubectlCmd) ManageResources(config *rest.Config, openAPISchema openapi.Resources) (kube.ResourceOperations, func(), error) {
+	return k.KubectlCmd.ManageResources(RestConfigForTempKubeconfig(config), openAPISchema)
+}
+
 func NewKubectl() kube.Kubectl {
-	return &kube.KubectlCmd{Tracer: tracer, Log: *logger.New().LogrLogger}
+	return &kubectlCmd{
+		KubectlCmd: &kube.KubectlCmd{Tracer: tracer, Log: *logger.New().LogrLogger},
+	}
 }
--- a/internal/util/kubernetes/restconfig.go
+++ b/internal/util/kubernetes/restconfig.go
@@ -1,0 +1,26 @@
+package kubernetes
+
+import (
+	"k8s.io/client-go/rest"
+)
+
+const inClusterServiceAccountTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+
+// RestConfigForTempKubeconfig returns a copy of config for gitops-engine's temp kubeconfig.
+//
+// gitops-engine WriteKubeConfig copies rest.Config.BearerToken into the kubeconfig as a static
+// token. client-go refreshes credentials from BearerTokenFile on each request, but kubectl
+// subprocesses use only what is written to the kubeconfig. After a projected service account
+// token rotates, the static token goes stale while the main controller client keeps working.
+//
+// When BearerTokenFile is set, drop the startup BearerToken snapshot so WriteKubeConfig falls
+// back to tokenFile in the generated kubeconfig (the standard in-cluster path for pods).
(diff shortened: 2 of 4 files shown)


internal/util/kubernetes/kubectl.go
```diff
@@ -17,6 +19,16 @@ func init() {
 	}
 }
 
+type kubectlCmd struct {
+	*kube.KubectlCmd
+}
+
+func (k *kubectlCmd) ManageResources(config *rest.Config, openAPISchema openapi.Resources) (kube.ResourceOperations, func(), error) {
+	return k.KubectlCmd.ManageResources(RestConfigForTempKubeconfig(config), openAPISchema)
+}
```
juliev0, line 28: ManageResources is called every time we try to do a sync (`NumaflowController` resources)

Now we call `RestConfigForTempKubeconfig` function defined below.

internal/util/kubernetes/kubectl.go
```diff
@@ -17,6 +19,16 @@ func init() {
...
+
 func NewKubectl() kube.Kubectl {
-	return &kube.KubectlCmd{Tracer: tracer, Log: *logger.New().LogrLogger}
+	return &kubectlCmd{
+		KubectlCmd: &kube.KubectlCmd{Tracer: tracer, Log: *logger.New().LogrLogger},
+	}
```
juliev0, line 33: need to use our own `kubectlCmd` struct so that we call our custom `ManageResources` function

internal/util/kubernetes/restconfig.go
```diff
@@ -0,0 +1,26 @@
...
+// back to tokenFile in the generated kubeconfig (the standard in-cluster path for pods).
+func RestConfigForTempKubeconfig(config *rest.Config) *rest.Config {
+	if config == nil || config.BearerTokenFile == "" {
+		return config
+	}
+
+	copied := rest.CopyConfig(config)
+	copied.BearerToken = ""
+	return copied
+}
```
juliev0, line 26: clear out the cached `BearerToken` value and force gitops-engine to read from `BearerTokenFile` (populated by kubelet) instead

whynowy: Since `controller-runtime` is used in the project, why don't we use `GetConfigOrDie`? 

https://github.com/kubernetes-sigs/controller-runtime/blob/main/alias.go#L114

whynowy: Or `k8s.io/client-go/rest.InClusterConfig()` for in-cluster config.

https://github.com/numaproj/numaflow/blob/main/pkg/shared/util/kubeconfig.go#L38-L44

juliev0: Cursor says GetConfigOrDie won't work:

**What `GetConfigOrDie()` actually gives you:**
It builds a rest.Config for client-go/controller-runtime:

In-cluster: reads the service account token once into BearerToken and sets BearerTokenFile
Wraps HTTP transport so requests can re-read the token file on each call
That works fine for normal reconcilers (Get, List, Patch, etc.) — which is why those keep working after token rotation.

**Why that doesn’t help gitops-engine sync**
NumaflowController sync does not use rest.Config directly for apply. gitops-engine does:

NewSyncContext(...) → kubectl.ManageResources(rawConfig, ...)
ManageResources → WriteKubeConfig(config, ...) → temp kubeconfig file
kubectl subprocess reads that file, not the live rest.Config transport
WriteKubeConfig copies rest.Config.BearerToken into the kubeconfig as a static token: field. It does not preserve the “refresh from file on every request” behavior that client-go gets from GetConfigOrDie().

So the failure isn’t “we didn’t call GetConfigOrDie()”; it’s “we re-serialize that config into a temp kubeconfig for kubectl, and that serialization drops token refresh.”

Calling GetConfigOrDie() again on each sync would still produce a rest.Config with the same shape (BearerToken snapshot + BearerTokenFile), and gitops-engine would still write the stale snapshot into the temp kubeconfig unless you fix that path (e.g. RestConfigForTempKubeconfig).

juliev0: Why calling `InClusterConfig()` again doesn’t help sync
The sync failure isn’t “we never loaded in-cluster config.” It’s:

- gitops-engine gets that same *rest.Config
- ManageResources → WriteKubeConfig writes a temp kubeconfig
- kubectl uses the static token: from that file, not the live refresh behavior

whynowy (approved):

Issue #929, opened 2025-12-05
Handle transient pipeline failures during original drain, not just force drain


juliev0:
**Describe the bug**

Currently, if we have a Pipeline Failure during the original drain, we stop and immediately switch to force drain. Should we handle transient failures, however, and give it an opportunity to recover?

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

PR #1005, opened 2026-06-29
fix: a force draining pipeline shouldn't switch its isbsvc

### LINKED ISSUE #892: Force Draining for a Pipeline in the case that the isbsvc has been updated
**Describe the bug**

Currently as part of Force Draining, we apply over top the "promoted" Pipeline. This works in general. However, if for some reason the "promoted" Pipeline is using a different isbsvc, this won't work. We apply the entire new pipeline spec, including the new isbsvc. 

However, you can never update a Pipeline's isbsvc. So, this update causes the Pipeline to fail.

Therefore, as part of Force Draining, we should always make sure to keep the isbsvc the same.

---

juliev0:
Fixes #892 #929 

### Modifications

1. Make it so a Pipeline force draining won't ever switch to using a newer isbsvc (which will never work, since a Pipeline can't switch isbsvc) - it always remains on the original isbsvc for this reason
2. If a Pipeline is briefly in a "Failed" state while it's trying to drain (prior to force drain), give it some time before giving up. 
3. Make sure "recyclable-start-time" annotation is only set once, when resources enters recyclable state.

### Backward (and forward) incompatibilities

The annotation "force-drain-failure-start-time" changed its name - this would only affect Pipelines in the middle of transient failure, so should be unusual and not terribly consequential.

### DIFF
--- a/internal/controller/pipelinerollout/pipelinerollout_recycle.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_recycle.go
@@ -702,6 +702,8 @@ func markPipelineForceDrainStarted(ctx context.Context, c client.Client, pipelin
 	}); err != nil {
 		return err
 	}
+	// just in case a previous drain had a state in which the Pipeline was "failed", we clear that time so we can reuse the annotation from scratch
+	// if we need to
 	return clearDrainFailureStartTimeAnnotation(ctx, c, pipeline)
 }
 
--- a/internal/controller/pipelinerollout/pipelinerollout_recycle.go
+++ b/internal/controller/pipelinerollout/pipelinerollout_recycle.go
@@ -311,7 +311,7 @@ func (r *PipelineRolloutReconciler) checkForFailedPipeline(ctx context.Context, 
 	// check if we've waited long enough
 	startTime, err := time.Parse(time.RFC3339, pipeline.GetAnnotations()[common.AnnotationKeyDrainFailureStartTime])
 	if err != nil {
-		return false, fmt.Errorf("failed to parse drain failure start time annotation %q on pipeline %s/%s: %w", startTime, pipeline.GetNamespace(), pipeline.GetName(), err)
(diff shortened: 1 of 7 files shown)


internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -274,8 +282,6 @@ func (r *PipelineRolloutReconciler) forceDrain(ctx context.Context,
 			currentVal, _ := kubernetes.GetAnnotation(pipeline, common.AnnotationKeyForceDrainSpecsCompleted)
 			err := kubernetes.SetAndPatchAnnotations(ctx, r.client, pipeline, map[string]string{
 				common.AnnotationKeyForceDrainSpecsCompleted: currentVal + promotedPipeline.GetName() + ",",
-				// reset this so we won't try to look at it on the next force drain attempt
-				common.AnnotationKeyForceDrainFailureStartTime: "",
```
juliev0, line 278: moved this clear to occur when about to perform a new force drain instead

internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -340,22 +346,41 @@ func (r *PipelineRolloutReconciler) checkForPromotedPipelineForForceDrain(ctx co
...
+	}
+
+	// Preserve the current pipeline's ISBService name so we don't change it during force drain (a Pipeline cannot have its ISBService changed or it won't work)
+	currentISBSvcName, err := numaflowtypes.GetPipelineISBSVCName(currentPipeline)
+	if err != nil {
+		return nil, err
+	}
+	if err := numaflowtypes.PipelineWithISBServiceName(forceAppliedPipelineSpec, currentISBSvcName); err != nil {
+		return nil, err
+	}
```
juliev0, line 366: this code is new. The rest of the code in this function is simply extracted from `forceApplySpecOnUndrainablePipeline`

PR #1008, opened 2026-07-04
feat: track force drains in the PipelineRollout status

juliev0:
#### Recording Drain attempts per recyclable Pipeline in the PipelineRollout Status

This adds a new section to the PipelineRollout Status which is related to Recyclable Pipelines. When there's a Pipeline in upgrade-state "recyclable" or "recyclable-expired", we want to know the periods of time that it attempted to drain, was it the original spec or was it another pipelines's spec force applied to it, was it stopped from draining because its phase was "Failed" for a period of time? or did it reach pauseGracePeriodSeconds? 

This also removes some of the previously used annotations on the Pipeline which were being used for tracking this information within the code:
- "force-drain-specs-started"
- "force-drain-specs-completed"
- "drain-failure-start-time"

For backward compatibility, in order to handle the case that certain pipelines are already in the middle of the force draining process, we migrate the previous annotations into the new `RecyclablePipelineStatus.DrainAttempts`.

#### Allow user to override whether automatic deletion occurs on failure to drain

Up until now, there was just one general setting for `KeepUndrainedPipelines` in the Controller config - this indicates if Pipelines should be deleted vs marked "recycable-expired" after they've failed to drain/force drain. Now the user can override this setting as well.

### Verification

Made sure that the force drain e2e test is still working and that the Status looks correct in the PipelineRollout.

Added a temporary e2e test to verify that the migration of annotations to DrainAttempts is working (backward compatibility).

### Backward (and forward) incompatibilities

see above

## Summary by CodeRabbit

* **New Features**
  * Added `keepUndrainedPipelines` to retain undrained recyclable pipelines past `maxRecyclableDurationMinutes`.
  * Extended `PipelineRollout` status with `recyclablePipelinesStatus` to record drain-attempt history, completion state/reason, timestamps, and per-vertex replica counts.
* **Bug Fixes**
  * Improved recyclable force-drain tracking to use drain-attempt state as the source of truth, including migration of legacy signals and pruning of stale status entries.
* **Tests**
  * Added a new e2e suite for force-drain backward compatibility and enhanced existing force-drain e2e assertions.
* **Chores**
  * Updated the e2e CI matrix and consolidated e2e test target generation in build tooling.


internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -1325,7 +1325,37 @@ func (r *PipelineRolloutReconciler) garbageCollectChildren(
 
 	}
 
-	return ctlrcommon.GarbageCollectChildren(ctx, pipelineRollout, r, r.client)
+	allDeleted, recyclableObjects, err := ctlrcommon.GarbageCollectChildren(ctx, pipelineRollout, r, r.client)
```
juliev0: We update the recyclable Pipeline status inside of `GarbageCollectChildren()` for all existing recyclable children. Then below where we call ` pruneRecyclablePipelinesStatus`, we clear out any status related to the recyclable children which have been deleted.

internal/controller/common/rollout_controller.go
```diff
@@ -42,15 +42,15 @@ type RolloutController interface {
...
+// Garbage Collect all recyclable children; return true if we've deleted all that are recyclable.
 func GarbageCollectChildren(
 	ctx context.Context,
 	rolloutObject RolloutObject,
 	controller RolloutController,
 	c client.Client,
 ) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
-	recyclableObjects, err := getRecyclableObjects(ctx, rolloutObject, c)
+	recyclableObjects, err := FindChildrenOfUpgradeState(ctx, rolloutObject, common.LabelValueUpgradeRecyclable, nil, false, c)
```
juliev0, line 53: this was equivalent - we didn't need that other function

internal/controller/pipelinerollout/pipelinerollout_controller.go
```diff
@@ -1325,7 +1325,50 @@ func (r *PipelineRolloutReconciler) garbageCollectChildren(
...
+// that still exist
+func (r *PipelineRolloutReconciler) pruneRecyclablePipelinesStatus(
+	ctx context.Context,
+	pipelineRollout *apiv1.PipelineRollout,
+) error {
+	recyclableObjects, err := ctlrcommon.FindChildrenOfUpgradeState(ctx, pipelineRollout, common.LabelValueUpgradeRecyclable, nil, false, r.client)
+	if err != nil {
+		return fmt.Errorf("error listing recyclable pipelines: %w", err)
+	}
+	recyclableExpiredObjects, err := ctlrcommon.FindChildrenOfUpgradeState(ctx, pipelineRollout, common.LabelValueUpgradeRecyclableExpired, nil, false, r.client)
```
juliev0: we look for the "recyclable" and "recyclable-expired" Pipelines - anything else would've been deleted

internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -110,7 +113,7 @@ func (r *PipelineRolloutReconciler) Recycle(
 	}
 
 	// Is the pipeline still defined with its original spec or have we overridden it with that of the "promoted" pipeline?
-	originalSpec := isPipelineSpecOriginal(pipeline)
+	originalSpec := !recyclablePipelineStatus.HasForceDrainStarted(pipeline.GetName())
```
juliev0, line 118: Now instead of looking at the pipeline annotation "force-drain-specs-started", we look at the PipelineRollout.Status.RecyclablePipelineStatus for this particular Pipeline to see if there's evidence of a force drain.

internal/controller/pipelinerollout/pipelinerollout_recycle.go
```diff
@@ -184,7 +201,7 @@ func (r *PipelineRolloutReconciler) checkAndPerformForceDrain(ctx context.Contex
 
 	if promotedPipeline != nil {
 		// did we already try and fail to force drain with this promoted pipeline spec?
-		if checkForValueInCommaDelimitedAnnotation(pipeline, promotedPipeline.GetName(), common.AnnotationKeyForceDrainSpecsCompleted) {
+		if recyclablePipelineStatus.IsDrainAttemptComplete(promotedPipeline.GetName()) {
```
juliev0, line 210: no longer need to look in the annotation - now we look at the DrainAttempt in the PipelineRollout Status

Makefile
```diff
@@ -156,6 +156,7 @@ test-pipeline-rider-e2e:
 test-monovertex-rider-e2e:
 test-rollback-e2e:
 test-force-drain-e2e:
+test-force-drain-backward-compatibility-e2e:
```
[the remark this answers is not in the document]
juliev0: fixed

juliev0:
> **Actionable comments posted: 1**
> 
> 🧹 Nitpick comments (2)
> > tests/e2e/force-drain-e2e/force_drain_test.go (1)> `324-370`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Remove vestigial, unused annotations lookup.**
> > > `annotations` is fetched and guarded on but never actually used — the drain-attempt/spec-status checks below don't reference it. This looks like leftover logic from the previous annotation-based tracking this PR replaces with `DrainAttempts`. Beyond being dead code, it can spuriously return `false` (delaying convergence) if the pipeline momentarily has no annotations map even though the real success conditions are already met.

Thanks. I updated this test and this is one of the changes that was made. 

> > 
> > 
> > internal/controller/pipelinerollout/pipelinerollout_recycle.go (1)> `703-785`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Dead merge-logic in `migrateForceDrainAnnotationsToDrainAttempts`.**
> > > The early return at Line 730-732 (`if len(recyclablePipelineStatus.DrainAttempts) > 0 { return nil }`) guarantees `recyclablePipelineStatus.DrainAttempts` is always empty by the time `existingBySource` is built (Line 740-743). Consequently, every `existingBySource[...]` lookup (Lines 759-764, 775-778) always misses, and the "preserve existing attempt" branches are unreachable. This is confirmed by the test suite: "skips migration when DrainAttempts already set" only exercises the early-return path, never the merge path.
> > > This is currently harmless (behavior matches the always-not-exists branch), but it's confusing dead code that could mislead a future change to the guard into assuming merge-with-existing-state is already implemented and tested.

I decided not to change this since 1. it's temporary code for backward compatibility and will be removed soon, 2. it's harmless and in case the first part were removed for some reason, the second part would take over.

juliev0:
> 🧹 Nitpick comments (1)
> > tests/e2e/force-drain-e2e/force_drain_test.go (1)> `314-331`: _🩺 Stability & Availability_ | _🔵 Trivial_ | _💤 Low value_
> > > **Simplify the error handling in this poll** The `errors.IsNotFound(err)` branch is redundant here because the generic `err != nil` path also returns `false`.
> > > 


fixed, thanks

afugazzotto (approved):

PR #1015, opened 2026-09-14
feat: Redesign MonoVertex AnalysisTemplate around critical-error rate

rahulsgh21:
### What this PR does / why we need it

Redesigns the MonoVertex progressive `AnalysisTemplate` to use `monovtx_critical_error_total` (Numaflow ≥ 1.8.0, [numaflow#3154](https://github.com/numaproj/numaflow/pull/3154)) instead of read/ack-only checks.

The old read/ack template was brittle on low-traffic assets and did not reflect real processing failures (EOT, UDF crashes, etc.). The new primary metric passes when no critical errors occurred in the lookback window, or when the counter is absent (lazy-initialized counter / no errors yet).

Per reviewer feedback, the template now runs **two metrics in parallel** (Option 2); both must pass:

| Metric | Purpose |
|--------|---------|
| `no-critical-errors` | `increase(monovtx_critical_error_total[interval]) == 0 OR absent(...)` |
| `acknowledged-messages` | Existing read/ack check (`read == 0 OR ack > 0`) for Numaflow &lt; 1.8 fallback |

On Numaflow &lt; 1.8, critical-error passes via `absent()` and read/ack still validates. On ≥ 1.8, both checks run.

**Timing / limits (sample template):**
- **`interval`**: `60s` — PromQL `increase()` lookback (~2 Prometheus scrapes at default 30s interval); matches UBC production config
- **`initialDelay`**: `60s` — wait one evaluation interval before the first `increase()` query (full lookback window). Not tied to Numaplane’s consecutive-health gate — AnalysisRun starts after that gate completes
- **`no-critical-errors-failure-limit`**: `1` — fail fast on critical errors
- **`acknowledged-messages-failure-limit`**: `3` — more tolerance for read/ack on low-traffic assets
- **`consecutiveSuccessLimit`**: `3` — promotion still requires 3 consecutive passes

**Scope:** sample `AnalysisTemplate` + e2e only — no controller/CRD changes.

`monovtx_critical_error_total` requires Numaflow ≥ 1.8.0; on older versions the template still works via `absent()` + read/ack fallback.

### Testing

- [x] `progressive-analysis-monovertex` e2e passes in CI (Numaflow v1.8.3)
- [x] Local e2e on Rancher Desktop:
  - **v1.7.5** — 4/4 passed (&lt;1.8 fallback: `absent()` + read/ack)
  - **main-efeb82b** — 4/4 passed (latest main)

```bash
# <1.8 fallback
NUMAFLOW_CONTROLLER_VERSION=1.7.5 make test-progressive-analysis-monovertex-e2e

# latest main
NUMAFLOW_CONTROLLER_VERSION=main-efeb82b make test-progressive-analysis-monovertex-e2e

### CODE REVIEW

File: config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml [hunk 1]
```diff
@@ -39,32 +39,42 @@ spec:
...
+      value: "3"
   metrics:
-  - name: acknowledged-messages
+  - name: no-critical-errors
     consecutiveSuccessLimit: '{{args.success-limit}}'
     failureLimit: '{{args.failure-limit}}'
     interval: '{{args.interval}}'
+    # Wait one interval after AnalysisRun creation before the first Prometheus query so the
+    # increase() lookback does not overlap the upgrading MonoVertex health ramp-up window.
+    initialDelay: '{{args.interval}}'
```
juliev0, line 66: waiting 1 minute seems good for making sure we get past the initial period of time in which the MonoVertexRollout containers could have been restarting.

juliev0: @vaibhavtiwari33 what's the frequency of the polling? Is it 30 seconds or 1 minute?

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -39,32 +39,42 @@ spec:
...
+  name: mvtx-no-critical-errors
   namespace: example-namespace
   labels:
     used_by: numaplane
 spec:
   args:
     - name: upgrading-monovertex-name
     - name: monovertex-namespace
+    # Shared by metric interval, initialDelay, and the PromQL increase() lookback.
+    # Align with the MonoVertex defaultAssessmentSchedule consecutive-health period (60s).
```
juliev0: As far as the "consecutive health period of 60 seconds", keep in mind that the AnalysisRun isn't launched until that period is over.

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
vaibhavtiwari33: By default it is 30s

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -39,32 +39,42 @@ spec:
...
+        # Pass when no critical errors (EOT, UDF crash, etc.) occurred in the lookback window.
+        # absent() treats assets with no traffic / no counter yet as healthy.
+        query: |
+          (
+            sum(increase(monovtx_critical_error_total{namespace="{{args.monovertex-namespace}}", mvtx_name="{{args.upgrading-monovertex-name}}"}[{{args.interval}}])) == 0
+          )
+          or
+          (
+            absent(monovtx_critical_error_total{namespace="{{args.monovertex-namespace}}", mvtx_name="{{args.upgrading-monovertex-name}}"})
+          )
```
vaibhavtiwari33, line 80: Once the error metric is identified to be missing should we rather fallback to the previously used metric?

juliev0, line 80: This query looks good to me. Basically, if the metric did not increase in the last 1 minute or if the metric doesn't exist: then pass.

juliev0: (of course, I'm not a prometheus expert, so testing it for the case of "metric doesn't exist" as well as the case of "metric exists and is 0" (assuming that both are possible to occur) would be ideal). I'm not sure how difficult that is? 

Do you know if the metric is emitted regardless, even if it's 0, or is it only emitted if there actually is an error?

vaibhavtiwari33: The thing I'm concerned about is regarding running this check on numaflow v1.7.x and less, it may always pass in those cases

vaibhavtiwari33: > Do you know if the metric is emitted regardless, even if it's 0, or is it only emitted if there actually is an error?

Depends on the type of counter metric being used. Let me check which one the error metric is.

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
juliev0: In that case, I think that that doesn't need to weigh in as far as this value, right?

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -39,32 +39,42 @@ spec:
...
     used_by: numaplane
 spec:
   args:
     - name: upgrading-monovertex-name
     - name: monovertex-namespace
+    # Shared by metric interval, initialDelay, and the PromQL increase() lookback.
+    # Align with the MonoVertex defaultAssessmentSchedule consecutive-health period (60s).
     - name: interval
-      value: 10s
+      value: 60s
```
juliev0, line 53: Just a note to anyone who is looking at this that even though it shows the previous value as 10s, in UBC we actually set it to 60s

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
vaibhavtiwari33: Since, we're using [Family](https://docs.rs/prometheus-client/0.24.0/prometheus_client/metrics/family/struct.Family.html) crate, no metric seems to be emitted until value is incremented (lazy initialization) ([ref](https://github.com/prometheus/client_rust/pull/279))

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -39,32 +39,42 @@ spec:
...
-          mvtx_name="{{args.upgrading-monovertex-name}}"}) or vector(0)) == 0)) or
-          (vector(1) and on() ((sum(numaflow_monovtx_ack_total{namespace="{{args.monovertex-namespace}}",
-          mvtx_name="{{args.upgrading-monovertex-name}}"}) or vector(0)) > 0)) or
-          vector(0)
+        # Requires Numaflow >= 1.8.0 (monovtx_critical_error_total, numaflow#3154).
+        # Pass when no critical errors (EOT, UDF crash, etc.) occurred in the lookback window.
+        # absent() treats assets with no traffic / no counter yet as healthy.
+        query: |
+          (
+            sum(increase(monovtx_critical_error_total{namespace="{{args.monovertex-namespace}}", mvtx_name="{{args.upgrading-monovertex-name}}"}[{{args.interval}}])) == 0
```
whynowy, line 75: What happens if the 1st detected error metric counter is > 0 (no previous data point), but no increase afterwards?

juliev0: This is a good point. Given that the failure-limit is 3 and the success-limit is 3, this means that a single failure is not enough to cause this to fail. But I think it should be. I think we should change failure-limit to 1, right? Then just one failure immediately causes this to fail.

juliev0: (unless somebody who knows the underlying numaflow code better than I do thinks that 1 failure is okay?)
(The idea is that we only run this after one consecutive minute of resource health of course, so presumably we should be past the ephemeral restarts.)

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
juliev0: Yes, ultimately, we need to decide how we want to do this wrt Numaflow version. 
Option 1: have logic in the deployment of the AnalysisTemplate through github which determines which AnalysisTemplate to include, based on user's numaflow controller version. 
Option 2: have a second metric in the same AnalysisTemplate and deploy that regardless of version. Metrics are exercised in parallel, and failures are basically ANDed. The metric shown above works for older versions, so it would essentially fall back to the original metric. 
Maybe Option 2 is good, right?

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
rahulsgh21: Yeah! The AnalysisRun only starts after the consecutive health window, so that shouldn't justify interval/initialDelay. Updated the comment. I kept 60s for the PromQL lookback (2 scrapes at 30s) and to match UBC.

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
adarsh0728: Yes, i also feel that we should go with Option 2 where critical_error metric and previous logic is ANDed. This would be much more stricter and will also work for all the versions

adarsh0728:
note that critical_error metric will be functional from 1.9 where we are avoiding numa restarts https://github.com/numaproj/numaflow/issues/3367

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
rahulsgh21: Added ack-messages alongside no-critical-errors in the same template; both must pass. On < 1.8, critical-error passes via absent() and read/ack still validates.

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
rahulsgh21: failure-limit is now 1 (sample + e2e).

config/samples/numaplane.numaproj.io_v1alpha1_monovertexrollout.yaml
```diff
@@ -39,29 +39,53 @@ spec:
...
-      value: 10s
+      value: 60s
     - name: success-limit
-      value: "3" # if we do this check too early and the prometheus metric doesn't know about the read count yet, it could
-                        # succeed when it shouldn't
-    - name: failure-limit 
-      value: "3" # if we do this check too early and the read has occurred but not the ack, it could fail when it shouldn't
+      value: "3"
+    - name: failure-limit
+      value: "1"
```
juliev0: Can we have 2 args for failure-limit, one which applies to the new metric and one which applies to the old? We purposely chose "3" for the original metric.

rahulsgh21: addressed!

juliev0:
This all looks good to me. Only thing is, I don't think you've tested it on the latest numaflow yet, right? If it's easy to test, let's do it. If it's not, I am okay to merge this and revisit once Numaflow has been released to include the _"numa" container not restarting_ feature. 

@adarsh0728 @whynowy @vaibhavtiwari33 - is there anything holding up the release for that? 

If you do want to test it now, I believe this should work:

- don't use a NumaflowControllerRollout. Just run the Numaflow Controller directly. But you need to make sure to use the [numaflow minimal CRDs](https://github.com/numaproj/numaflow/tree/main/config/base/crds/minimal) and not the [full CRDs](https://github.com/numaproj/numaflow/tree/main/config/base/crds/full). Numaplane doesn't work with full CRDs. I'm pretty sure that if you run `make start` from Numaflow, it uses the full CRDs by default.
- create your MonoVertexRollout (and AnalysisTemplate) and then update it to cause a Progressive Rollout

adarsh0728:
> * NumaflowControllerRollout

_numa container not restarting feature_ is already in latest main , @rahulsgh21 , take the image from latest numaflow main and plan to test the above changes as Julie suggested.

vaibhavtiwari33:
> is there anything holding up the release for that?

@juliev0 No blockers right now, it is targeted to be released by end of quarter. 

> If you do want to test it now, I believe this should work

I believe @rahulsgh21 has tested this in his local cluster using both old and new versions. Do we generally test numaplane changes in dev clusters at intuit as well?

juliev0 (approved):

PR #1025, opened 2026-09-20
fix: trigger progressive rollout on MonoVertex UDF image change

### LINKED ISSUE #1016: Monovertex UDF container image change does not trigger progressive rollout
**Describe the bug**

When a MonoVertex's UDF container image (`spec.udf.container.image`) is changed — e.g. from a stable image to a failing/EOT one — the change is applied directly instead of going through a progressive rollout with analysis. As a result, no AnalysisRun is created, `monovtx_critical_error_total` is never checked, and there is no automatic rollback if the new UDF image is broken.

**To Reproduce**
1. Run a monovertex with source -> udf(image A) -> sink
2. Update udf image to image B
3. Upgrade happens in place without progressive rollout

**Expected behavior**

Progressive rollout must happen.

rahulsgh21:
## What this PR does / why we need it

When a MonoVertex's UDF container image (`spec.udf.container.image`) is changed, the change was applied directly instead of going through a progressive rollout with analysis. As a result, no `AnalysisRun` was created, `monovtx_critical_error_total` was never checked, and there was no automatic rollback if the new UDF image was broken.

**To Reproduce**
1. Run a MonoVertex with `source -> udf(image A) -> sink`
2. Update UDF image to image B
3. Upgrade happens in place without progressive rollout

Fixes #1016

## Testing

Added an e2e case in `analysis_monovertex_test.go` that changes only the UDF image to a known-failing image (`map-badcat`) and verifies:
- Progressive rollout is triggered (candidate MonoVertex created)
- The `no-critical-errors` AnalysisRun metric fails via `monovtx_critical_error_total`
- Rollback occurs: the upgrading MonoVertex scales to 0, the promoted MonoVertex is left untouched

## How to reproduce

**Before the fix (on `main`):**
1. `make start` to deploy Numaplane locally (kind/k3d, with Prometheus + Argo Rollouts).
2. Create a MonoVertexRollout with `source -> udf(image A) -> sink`.
3. Patch only `spec.udf.container.image` to a different image.
4. Observe: the promoted MonoVertex is edited in place — no `-1` upgrading instance is created, no `AnalysisRun` appears (`kubectl get analysisrun -n <ns>`), and `MonoVertexRollout.status.upgradeInProgress` stays `NoOp`.

**After the fix (on this branch):**
1. Check out this branch and `make start` (or re-apply `config/manager/usde-config.yaml` / `config/install.yaml` to an existing cluster).
2. Repeat the same UDF image patch.
3. Observe: a `-1` upgrading instance is created, an `AnalysisRun` runs against `no-critical-errors`/`acknowledged-messages`, and `status.upgradeInProgress` becomes `Progressive`.

**Automated:**
```bash
make test-progressive-analysis-monovertex-e2e
# Requires a cluster context with Prometheus and Argo Rollouts installed (make start sets this up).
```


juliev0 (approved):

juliev0:
Thanks @rahulsgh21 ! I approved the change.

Sorry if this isn't due to your change, but since you recently changed the progressive-analysis-monovertex tests, do you mind looking into why it failed? (you can download the log files and the resource change files from CI) - maybe there's some flakiness that can be addressed? 

Also, were you able to verify that there were no other recent additions to numaflow that failed to get incorporated into the USDE? 

Thank you!

rahulsgh21:
> Sorry if this isn't due to your change, but since you recently changed the progressive-analysis-monovertex tests, do you mind looking into why it failed? (you can download the log files and the resource change files from CI) - maybe there's some flakiness that can be addressed?

Sure! This was in the new e2e test I added, not flakiness. My original UDF test image (`map-badcat`) was not present in quay, that's why the ci failed. I simplified the test to match the PR's actual scope: it now adds a UDF (previously absent) using an ordinary working image (`map-cat:stable`, already used elsewhere in this suite) and asserts Progressive triggers successfully.

tests/e2e/progressive-analysis-monovertex-e2e/analysis_monovertex_test.go
```diff
@@ -60,6 +62,10 @@ var (
 	monoVertexScaleTo               = int64(2)
 	monoVertexScaleMinMaxJSONString = fmt.Sprintf("{\"disabled\":null,\"max\":%d,\"min\":%d}", monoVertexScaleMax, monoVertexScaleMin)
 	monovertexSinkBadImage          = "quay.io/numaio/numaflow-go/sink-log-failure:stable"
+	// map-badcat is the known-bad counterpart of the standard map-cat UDF; it errors on every
+	// invocation, which drives the UDF's gRPC map redrive-error path and increments
+	// monovtx_critical_error_total (see rust/numaflow-core/src/mapper/map.rs in numaflow).
+	monovertexUDFBadImage = "quay.io/numaio/numaflow-go/map-badcat:stable"
```
adarsh0728: the image is not there in quay

rahulsgh21:
> Also, were you able to verify that there were no other recent additions to numaflow that failed to get incorporated into the USDE?
> 
> Thank you!

Yes, found a few more fields that fall through to direct-apply instead of triggering progressive rollout, same root cause as this one: spec.bypass on MonoVertex, spec.sink.onSuccess.* on both kinds, fallback/initContainers/sidecars on Pipeline vertices, and spec.ordered/spec.interStepBuffer on Pipeline.

juliev0 (approved):

juliev0:
Thanks. Do you mind adding the missing fields @rahulsgh21 ? Don't worry about adding any new tests for it. Just add the fields to USDE, and we should be good. Thanks!

juliev0:
I apologize that I didn't explicitly state that there was no need to add additional testing, and then you bothered to create a new monovertex in the test, which I know took time. I guess the question is: do you want to leave your test changes in or no? One thing to keep in mind is that we already have so much e2e testing which takes a long time, so in general I'd like to be conservative about only adding new e2e tests in when they provide some critical value. For a new field in the USDE, it's not really needed. Sorry again that this wasn't stated up front!

rahulsgh21:
> Thanks. Do you mind adding the missing fields @rahulsgh21 ? Don't worry about adding any new tests for it. Just add the fields to USDE, and we should be good. Thanks!

Added!

rahulsgh21:
> I apologize that I didn't explicitly state that there was no need to add additional testing, and then you bothered to create a new monovertex in the test, which I know took time. I guess the question is: do you want to leave your test changes in or no? One thing to keep in mind is that we already have so much e2e testing which takes a long time, so in general I'd like to be conservative about only adding new e2e tests in when they provide some critical value. For a new field in the USDE, it's not really needed. Sorry again that this wasn't stated up front!

I do realize that there are so many e2e tests. I have already confirmed that it's working, so I have dropped it in the new commit. Thank you.

adarsh0728 (approved):

juliev0 (approved):

juliev0:
> > I apologize that I didn't explicitly state that there was no need to add additional testing, and then you bothered to create a new monovertex in the test, which I know took time. I guess the question is: do you want to leave your test changes in or no? One thing to keep in mind is that we already have so much e2e testing which takes a long time, so in general I'd like to be conservative about only adding new e2e tests in when they provide some critical value. For a new field in the USDE, it's not really needed. Sorry again that this wasn't stated up front!
> 
> I do realize that there are so many e2e tests. I have already confirmed that it's working, so I have dropped it in the new commit. Thank you.

thanks so much for the bother!

